PR #300, opened 2018-08-09
Turn off maintenance mode

izgeri:
Since nothing is being published to prod, we can turn off maintenance mode for now.


sgnn7:
LGTM!

PR #311, opened 2018-08-10
WIP - Cleaned up channel closing in main roxy loop

sgnn7:
This should allow us to close the channel gracefully as well as make
sure that we don't get any panics.


sgnn7:
@doodlesbykumbi With your fix, this branch builds fine 👍 . Merging :)

PR #312, opened 2018-08-11
feat: add kubernetes secrets provider

### LINKED ISSUE #175: a kubernetes secret provider exists for secretless
as opposed to using explicit volume mounts and the file provider

`GIVEN` credentials stored within k8s secrets
`WHEN` secretless config uses the k8s provider
`THEN` the provider should use the service account of the pod to authenticate against the k8s API
`AND` retrieve a secret from the given path

doodlesbykumbi:
Resolves #175

## What does it do ?

+ Adds a Kubernetes Secrets provider that uses the Kubernetes API (via credentials local to the pod) to fetch secrets

This is particularly slick in how it allows instant rotation when using Kubernetes Secrets :rocket:

**TODO:** update k8s-demo to use this

## Confirm it works

```
# build secretless-dev:latest inside minikube
$(minikube docker-env)
./bin/build

# create namespace
kubectl create namespace meow

# create initial secret
kubectl create -n meow secret generic test-secret --from-literal=username=prodxuser --from-literal=password=Y4nys7f11

# run dev container
kubectl run -n meow -i --rm --tty secretless-dev-example --overrides='
{
  "apiVersion": "v1",
  "kind": "Pod",
  "metadata": {
    "name": "secretless-dev-example"
  },
  "spec": {
    "containers": [
      {
        "name": "secretless",
        "image": "secretless-dev:latest",
        "imagePullPolicy": "Never",
        "args": [
          "bash"
        ],
        "stdin": true,
        "stdinOnce": true,
        "tty": true,
        "volumeMounts": [{
          "mountPath": "/secretless",
          "name": "store"
        }]
      }
    ],
    "volumes": [{
      "name":"store",
      "hostPath": {
        "path": "/Users/supremeoverlord/go/src/github.com/cyberark/secretless-broker"
      }
    }]
  }
}
' --image secretless-dev:latest --restart=Never -- bash

# run secretless
cat << EOL | go run cmd/secretless-broker/main.go -f /dev/stdin
listeners:
  - name: http_good_basic_auth
    debug: true
    protocol: http
    address: 0.0.0.0:8080

handlers:
  - name: http_good_basic_auth_handler
    type: basic_auth
    listener: http_good_basic_auth
    debug: true
    match:
      - ^http.*
    credentials:
      - name: username
        provider: kubernetes
        id: test-secret#username
      - name: password
        provider: kubernetes
        id: test-secret#password
EOL

# use secretless

## exec into dev container
kubectl exec -n meow -it secretless-dev-example bash

## poll http endpoint
export http_proxy=localhost:8080
make_request() {
    response=$(curl --request GET --url http://scooterlabs.com/echo.json)
    pretty_resp=$(echo "$response" | jq -r .headers.Authorization)
    echo '"'"$(echo "${pretty_resp##* }" | base64 --decode)"'"' | jq .
}

while true; do
   sleep 2
   make_request
done

# rotate password as first argument
rotate() {
    new_password=$(echo "$1" | base64)
    kubectl -n meow patch secret test-secret -p='{"data":{"password": "'${new_password}'"}}' -v=1
}
rotate some_new_password
```


internal/app/secretless/providers/kubernetes-secrets/provider.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"log"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	plugin_v1 "github.com/cyberark/secretless-broker/pkg/secretless/plugin/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/api/core/v1"
```
sgnn7: k8s imports should be above internal imports

internal/app/secretless/providers/providers.go
```diff
@@ -7,6 +7,7 @@ import (
 	keychainProvider "github.com/cyberark/secretless-broker/internal/app/secretless/providers/keychain"
 	literalProvider "github.com/cyberark/secretless-broker/internal/app/secretless/providers/literal"
 	vaultProvider "github.com/cyberark/secretless-broker/internal/app/secretless/providers/vault"
+	kubernetesProvider "github.com/cyberark/secretless-broker/internal/app/secretless/providers/kubernetes-secrets"
```
sgnn7: This should be right above keychainProvider import (alphabetized)

internal/app/secretless/providers/providers.go
```diff
@@ -19,4 +20,5 @@ var ProviderFactories = map[string]func(plugin_v1.ProviderOptions) plugin_v1.Pro
 	"keychain": keychainProvider.ProviderFactory,
 	"literal":  literalProvider.ProviderFactory,
 	"vault":    vaultProvider.ProviderFactory,
+	"kubernetes-secrets":  kubernetesProvider.ProviderFactory,
```
sgnn7: Nees alphabetic sort too

internal/app/secretless/providers/kubernetes-secrets/provider.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return value, nil
+}
+
+// KubeClient represents Kubernetes client and calculated namespace
+type KubeClient struct {
+	clientset *kubernetes.Clientset
+	clientConfig clientcmd.ClientConfig
+}
```
sgnn7: This type def probably needs to move up to the top of this module

internal/app/secretless/providers/kubernetes-secrets/provider.go
```diff
@@ -0,0 +1,115 @@
...
+
+// GetName returns the name of the provider
+func (p Provider) GetName() string {
+	return p.Name
+}
+
+// parseKubernetesSecretID returns the secret id and field name.
+func parseKubernetesSecretID(id string) (string, string) {
+	tokens := strings.SplitN(id, "#", 2)
```
sgnn7: We should check if the token count returned is 2 and that both have something before proceeeding

internal/app/secretless/providers/kubernetes-secrets/provider.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	secret, err := p.Client.GetSecret(currentNamespace, id);
+	if err != nil {
+		return nil, err
+	}
+
+	value, ok := secret.Data[fieldName];
+	if !ok {
+		return nil, fmt.Errorf("Kubernetes Secrets provider expects the secret '%s' to contain field '%s'", id, fieldName)
```
sgnn7: Might need a rewording of the error output - the current one almost makes it sound like it's exclusively a server issue when it could be also to client requesting a field that doesn't exist. I'm thinking something along the lines of  `ERROR: Could not find field '%s'  in Kubernetes secret '%s'` but feel free to make your own.

internal/app/secretless/providers/providers.go
sgnn7: Maybe the name could be a bit shoter? `k8s/secret` or something similar?

internal/app/secretless/providers/kubernetes-secrets/provider.go
doodlesbykumbi: addressed

internal/app/secretless/providers/providers.go
doodlesbykumbi: addressed

internal/app/secretless/providers/providers.go
doodlesbykumbi: addressed the first point. i'm going for kubernetes for the sake of being abundantly clear

internal/app/secretless/providers/kubernetes-secrets/provider.go
doodlesbykumbi: addressed

internal/app/secretless/providers/kubernetes-secrets/provider.go
doodlesbykumbi: addressed

internal/app/secretless/providers/kubernetes-secrets/provider.go
doodlesbykumbi: addressed

doodlesbykumbi:
@sgnn7 comments addressed

sgnn7:
Loog good!

PR #313, opened 2018-08-11
fix: pg_test.go TCP requires sslmode=disable

doodlesbykumbi:
I have no idea how tests have been passing prior to this.


sgnn7:
Great catch! LGTM!

PR #319, opened 2018-08-13
Fixed soft-reload 100% CPU bug

sgnn7:
Old code would keep old listener gofuncs running after listeners were
closed which would keep erroring out on `Accept`s and keep the loops
spinning forever, driving the CPU usage up to 100%. We now flag the
listerner as being wound down and check that flag in the accept loops,
allowing us to exit listener once the shutdown has begun.

[Jenkins Build](https://jenkins.conjur.net/view/cyberark/job/cyberark--secretless-broker/job/fix-broken-shutdown/)

### DIFF
--- a/internal/app/secretless/handlers/mysql/handler.go
+++ b/internal/app/secretless/handlers/mysql/handler.go
@@ -1,10 +1,10 @@
 package mysql
 
 import (
-	"log"
-	"net"
 	"fmt"
 	"io"
+	"log"
+	"net"
 
 	"github.com/cyberark/secretless-broker/internal/app/secretless/handlers/mysql/protocol"
 	plugin_v1 "github.com/cyberark/secretless-broker/pkg/secretless/plugin/v1"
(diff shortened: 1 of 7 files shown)


pkg/secretless/plugin/v1/listener.go
```diff
@@ -47,9 +47,11 @@ type BaseListener struct {
 	Config         config.Listener
 	EventNotifier  EventNotifier
 	HandlerConfigs []config.Handler
+	IsShuttingDown bool
```
doodlesbykumbi: Perhaps there's a more accurate name for this field e.g. `ShutdownStarted`, which would be true while the listener is shutting down and has already shut down.

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -27,16 +27,18 @@ type BackendConfig struct {
...
-	BackendConfig    *BackendConfig
+	BackendConfig *BackendConfig
 }
 
 func (h *Handler) abort(err error) {
-	mysqlError := protocol.Error{
-		Code:     protocol.CRUnknownError,
-		SQLSTATE: protocol.ErrorCodeInternalError,
-		Message:  err.Error(),
+	if h.GetClientConnection() != nil {
```
doodlesbykumbi: Let's add this check to the pg handler as well.

doodlesbykumbi:
LGTM!

PR #327, opened 2018-08-14
Minor styling and content fixes to the website

### LINKED ISSUE #318: Cosmetic text / style feedback on the website
Sorry for the multiple reports in this one issue, but they're all small:

-  [x] `Github` should always be spelled `GitHub`. It's on the top menu like that, for one.
- [x] The explanation below the "What is the Secretless Broker?" could benefit from mentioning the word "password" as a way to explain what a "secret" is in this context. We're assuming our visitors have that in their daily vocabulary but they might not.
- [x] In the same explanation as above, we should also add `microservices`, because `web APIs` might get construed as only external APIs for 3rd party services.
- [x] `Handle Secrets Rotation Seamlessly` should be `Handle Secret Rotations Seamlessly`
- [x] `It's hard for applications to respond to changing secrets.` doesn't read clearly to me. Something like "It's hard for applications to deal with values of secrets being changed" would be better.
- [x] `without becoming part of the secrets management threat surface` needs a period at the end.
- [x] `Credentials carry privilege. Don't allow your applications...` Those 2 sentences are written very differently: first it's a statement of fact and then it's followed by an order, and then later it changes again. 
- [x] `because Secretless integrates with your vault of choice` - shouldn't that say `the Secretless Broker`?
- [x] There's no CTA or next obvious place for the developer to go after they have finished quick start. The site needs something below that to take them to the next level of investigation, otherwise it will look like there's not much more meat on the bone. Maybe something like "Ready to use Secretless Broker in your Kubernetes environment?" and link forward to the next natural step for that.

sgnn7:
There were a small number of nits that needed fixing (e.g. Github ->
GitHub) and similar that needed fixing.

Resolves #318


docs/_layouts/landing.html
```diff
@@ -19,17 +19,17 @@ <h3>Work on Features, not Security</h3>
...
 			<div class="col-md-4 about-subitem">
 				<div class="inner">
 					<div class ="in-line-frame">
 						<i class="fa fa-angle-double-right btn-arrow-right" aria-hidden="true"></i>
-						<h3>Handle Secrets Rotation Seamlessly</h3>
+						<h3>Handle Secret Rotations Seamlessly</h3>
 					</div>
 					<p class="inner-text-description">
-					It's hard for applications to respond to changing secrets. Until now, applications required additional logic to handle secrets rotation. With the Secretless Broker, no code changes or application logic is necessary - the Secretless Broker handles it all for you.</p>
+					Until now, applications usually required additional logic and disruptive restarts to handle credential rotation. With the Secretless Broker, no code changes are necessary and no application restarts are needed for secrets rotations - the Secretless Broker handles it all for you!</p>
```
izgeri: `secret rotations` instead of `secrets rotations`?

docs/docs/get_started/quick_start.md
```diff
@@ -121,6 +121,10 @@ docker container run \
   </div>
 </div>
 
+<div>
+  Ready to use Secretless Broker in your Kubernetes environment? Continue with <a href="/docs/get_started/deploy_to_kubernetes.html">Deploying to Kubernetes</a>!
```
izgeri: can we link to `docs/reference/how_to_deploy.html` instead?

izgeri: Also, this needs to get added to index.md as well

docs/_layouts/landing.html
sgnn7: Yup - that was explicitly mentioned in the issue

PR #341, opened 2018-08-16
push sidecar-injector image to dockerhub as part of CI/CD

doodlesbykumbi:
resolves https://github.com/cyberark/secretless-broker/issues/336

### DIFF
--- a/sidecar-injector/bin/publish
+++ b/sidecar-injector/bin/publish
@@ -1,0 +1,10 @@
+#!/bin/bash -e
+#
+# Build and push secretless sidecar injector mutating webhook service
+# usage: ./bin/publish
+set -ex
+
+cd $(dirname $0)/..
+
+docker build -t cyberark/sidecar-injector:latest .
+docker push cyberark/sidecar-injector:latest
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,6 +52,7 @@ pipeline {
 
       steps {
         sh './bin/publish'
+        sh './sidecar-injector/bin/publish'
       }
     }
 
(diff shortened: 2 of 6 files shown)


sidecar-injector/bin/publish
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+#
+# Build and push secretless sidecar injector mutating webhook service
+# usage: ./bin/publish
+set -ex
+
+cd $(dirname $0)/..
```
sgnn7: use the new helper `abspath` - you can see how this is done in other scripts

sidecar-injector/bin/publish
```diff
@@ -0,0 +1,10 @@
...
+
+docker build -t cyberark/sidecar-injector:latest .
```
sgnn7: Use path from `abspath` for the context instead of `.`

sidecar-injector/bin/publish
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
```
sgnn7: Don't we need to build the image first?

sidecar-injector/bin/publish
doodlesbykumbi: I did see that but since this is not in `$ROOT_DIR/bin` then using `abspath` seemed like more work than what I've done. Also this is bound to be split out.

sidecar-injector/bin/publish
doodlesbykumbi: yup, it's happening below

doodlesbykumbi: i didn't see any value to having an independent `build` step

PR #349, opened 2018-08-20
Conjur handler consumes access token from Conjur provider

### LINKED ISSUE #340: Conjur handler can use Conjur k8s authenticator
The Conjur handler is updated so that it can leverage the Conjur K8s authenticator if it is configured to do so

When it uses the Conjur k8s authenticator, the handler should have a config that points to the token file:
```
credentials:
      - name: accessToken
        provider: literal
        id: /run/conjur/access-token
```

when the handler receives a connection request, if a file is specified and there is a `CONJUR_AUTHN_URL` env var set that includes `authn-k8s`, it should kick off the Conjur k8s authentication process (if it has not already) to retrieve the time-limited access token.

dsbyrne:
Resolves #340

This change cleans up handler implementations (mostly changing receiver types) and instantiates a Conjur provider from the Conjur handler.

### DIFF
--- a/internal/app/secretless/handlers/http/conjur.go
+++ b/internal/app/secretless/handlers/http/conjur.go
@@ -38,7 +38,15 @@ func (h ConjurHandler) Authenticate(values map[string][]byte, r *http.Request) e
 
 // ConjurHandlerFactory instantiates a handler given HandlerOptions
 func ConjurHandlerFactory(options plugin_v1.HandlerOptions) plugin_v1.Handler {
-	return &ConjurHandler{
+	handler := &ConjurHandler{
 		BaseHandler: plugin_v1.NewBaseHandler(options),
 	}
+
+	// Force instantiate the Conjur provider so we can use an access token.
+	// This will fail unless a means of authentication to Conjur is available.
+	if handler.Resolver != nil {
+		handler.Resolver.GetProvider("conjur");
+	}
+
+	return handler
 }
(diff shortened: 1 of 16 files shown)


internal/app/secretless/providers/vault/provider.go
```diff
@@ -27,16 +27,16 @@ func ProviderFactory(options plugin_v1.ProviderOptions) plugin_v1.Provider {
 		log.Panicf("ERROR: Could not create Vault provider: %s", err)
```
izgeri: this should throw an error too

internal/app/secretless/handlers/http/conjur.go
```diff
@@ -38,7 +38,14 @@ func (h ConjurHandler) Authenticate(values map[string][]byte, r *http.Request) e
 
 // ConjurHandlerFactory instantiates a handler given HandlerOptions
 func ConjurHandlerFactory(options plugin_v1.HandlerOptions) plugin_v1.Handler {
-	return &ConjurHandler{
+	handler := &ConjurHandler{
 		BaseHandler: plugin_v1.NewBaseHandler(options),
 	}
+
+	// Force instantiate the Conjur provider so we can use the access token
+	if handler.Resolver != nil {
```
izgeri: Can you add a comment here for clarity that nothing will happen unless Secretless is configured to authenticate to Conjur?

PR #373, opened 2018-08-29
Made configuration file reading use a configuration manager plugin

### LINKED ISSUE #74: Manager plugins can manipulate the secretless configuration
GIVEN the secretless manager plugin receives a configuration from the server
THEN the plugin can pass the configuration to secretless
AND the remote configuration is used
AND the `-f` argument need not be supplied

### LINKED ISSUE #102: Secretless reloads config from disk if it never got external configs
If we don't use external plugins to provide us with configurations, we should reread the config file on soft-reload request.

AC:
- [ ] If secretless never receives configs outside of its file, soft-reload should reread secretless.yml from disk

### LINKED ISSUE #215: default secretless.yml bubbles up from a list of directories
the first directory to contain a `secretless.yml` is used:

`pwd` -> `home` -> `/etc/secretless`

### LINKED ISSUE #366: Configuration file processing should be done in a configuration module
Since we have new configuration plugins, it only makes sense to get our regular config file loading to be done through one of those.

sgnn7:
Resolves #366 
Resolves #102 
Resolves #74 
Resolves #215 

[Jenkins Build](https://jenkins.conjur.net/view/cyberark/job/cyberark--secretless-broker/job/366-configfile-config-manage/)

Since we now can use configuration manager plugins, we now have our
regular config file reading moved into this abstraction. The CLI
interactions have not been changed so it is a fully backward-compatible
feature.

How to test:
- Run the CLI with any of the default switches `-f <config>`, `-watch`, and then try without and to see the traversal logic.

### DIFF
--- a/internal/app/secretless/configurationmanagers/configfile/manager.go
+++ b/internal/app/secretless/configurationmanagers/configfile/manager.go
@@ -1,0 +1,160 @@
+package configfile
+
+import (
+	"fmt"
+	"log"
+	"net/url"
+	"os"
+	"os/signal"
+	"os/user"
+	"path"
+	"strconv"
+	"syscall"
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/config"
+	plugin_v1 "github.com/cyberark/secretless-broker/pkg/secretless/plugin/v1"
+)
+
+const (
+	// ConfigFileName is the default name expected from configuration files
+	ConfigFileName = "secretless.yml"
+
+	// PluginName is the external name that this plugin will be identified by
+	PluginName = "configfile"
+)
+
+type configurationManager struct {
+	ConfigChangedFunc func(string, config.Config) error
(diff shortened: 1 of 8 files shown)


internal/app/secretless/configurationmanagers/configfile/manager.go
```diff
@@ -0,0 +1,160 @@
...
+	watchFile := false
+	if watchFileParam, ok := configSpecObject.Query()["watch"]; ok == true {
+		watchFileParamValue, err := strconv.ParseBool(watchFileParam[0])
+		if err == nil {
+			watchFile = watchFileParamValue
+		}
+	}
+
+	if len(configFilePath) > 0 {
+		log.Println("Trying to load configuration file: %s", configFilePath)
```
dsbyrne: `log.Printf` or remove `%s`

PR #441, opened 2018-09-21
Blog post & Quick Start content updates

typaulhus:
**What does this PR do?**
- Updates the Announcing Secretless blog post with a few small content updates
- Changes the Quickstart language a bit so it's clear that users will go through a demo, not get Secretless up & running for their environments.

### DIFF
--- a/docs/docs/get_started.md
+++ b/docs/docs/get_started.md
@@ -7,7 +7,7 @@ permalink: docs/get_started.html
 ---
 
 To get started trying out the Secretless Broker, you can run through our [quick start](/docs/get_started/quick_start.html)
-that allows you to try it out with a simple example using PostgreSQL, SSH, or an HTTP web service.
+that allows you to try some simple examples using PostgreSQL, SSH, or an HTTP web service.
 
 If you are looking for a more detailed introduction to how the Secretless Broker works in
 practice, please look at our [Kubernetes tutorial](/docs/get_started/kubernetes_tutorial.html).
--- a/docs/docs/get_started.md
+++ b/docs/docs/get_started.md
@@ -6,8 +6,8 @@ description: Secretless Broker Documentation
 permalink: docs/get_started.html
 ---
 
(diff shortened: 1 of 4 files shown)


docs/docs/get_started.md
```diff
@@ -6,8 +6,8 @@ description: Secretless Broker Documentation
 permalink: docs/get_started.html
 ---
 
-To get started trying out the Secretless Broker, you can run through our [quick start demos](/docs/get_started/quick_start.html)
-that allow you to try it out with PostgreSQL, SSH, or an HTTP web service.
+To get started trying out the Secretless Broker, you can run through our [quick start](/docs/get_started/quick_start.html)
+that allows you to try it out with a simple example using PostgreSQL, SSH, or an HTTP web service.
```
izgeri: this sounds awkward - maybe just "that allows you to try some simple examples using"

docs/docs/get_started/quick_start.md
```diff
@@ -6,8 +6,7 @@ description: Secretless Broker Documentation
 permalink: docs/get_started/quick_start.html
 ---
 
-Try out the Secretless Broker brokering a connection to a PostgreSQL database, an SSH connection,
-or a connection to an HTTP service authenticating with basic auth.
+Try a simple example using the Secretless Broker to connect a PostgreSQL database, an SSH connection, or a connection to an HTTP service authenticating with basic auth.
```
dsbyrne: The wording is a little strange here. What about,

"Try a simple example which uses the Secretless Broker to connect to a PostgreSQL database,  open an SSH connection, or authenticate to an HTTP service protected with basic auth."

docs/docs/get_started.md
typaulhus: i like it. Will update

docs/docs/get_started/quick_start.md
typaulhus: sounds good!

Issue #365, opened 2018-08-24
CRD integration is documented


sgnn7:
CRD code should have some information about it in the READMEs and website

izgeri:
Things I'd like to see in the README:
- [ ] Definition of what CRD is (ie what the acronym means, link to k8s docs, etc)
- [ ] Reference to resources in repository that are related to CRDs
- [ ] Info on requirements to use CRD convenience scripts (mercurial / go.sum stuff)

We should also have the website updated to include plugin documentation for the config manager plugin.

izgeri:
From the closed PR #448, we replaced the text
> By default, the CRD we use for the Secretless Broker is under `configurations.secretless.io`.
with:
----
### Operation
 When this configuration manager plugin starts, a new CRD is registered (if not already available)
as `configurations.secretless.io` after which the configuration ID specified by the name after the `#`
symbol on the CLI will be watched in that CRD namespace. As soon as the configuration with that ID is
defined or updated in Kubernetes, Secretless will notify the main daemon to load the new configuration
and soft-reload itself. Any number of Secretless Broker instances can watch the same configuration ID.
----

izgeri:
I also noted in the closed PR:
```
I'd like it to include a link to the k8s docs on CRDs, and I'd like to have an "## Installation" section that "Operation" and "Kubernetes API Permissions" falls under.

Eventually we should do all of this with a helm chart, probably, but for now someone coming to this page needs to clearly see

- How do I install this into my cluster, including the serviceaccount
- How do I use it in a given Secretless deployment, including the command to start the broker and the method to use to link it to the service account
```

izgeri:
This issue was moved to cyberark/secretless-docs#90

PR #448, opened 2018-09-24
WIP - Added more information about the CRD watcher operation

sgnn7:
We needed to include a bit more info about auto-creation of the CRD and
a bit more detail about its operation.


izgeri:
Closing this PR, since this is work that will be completed working with a technical writer as part of #365 in a future release.

PR #450, opened 2018-09-25
Update link to GH issues for handlers in dev

izgeri:
Instead of pointing just to the GH issues page, point to the page that shows epics for Secretless Handlers.
This will more accurately show the handlers currently in dev.


sgnn7:
CR Pass

PR #487, opened 2018-10-16
add tests for kubernetes secrets provider

doodlesbykumbi:


test/kubernetes_provider/kubernetes_provider_test.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+
+	"github.com/cyberark/secretless-broker/internal/app/secretless/providers"
+	"github.com/cyberark/secretless-broker/internal/app/secretless/providers/kubernetessecrets"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	plugin_v1 "github.com/cyberark/secretless-broker/pkg/secretless/plugin/v1"
+	testclient "k8s.io/client-go/kubernetes/fake"
```
izgeri: <3

internal/app/secretless/providers/kubernetessecrets/client.go
```diff
@@ -4,47 +4,33 @@ import (
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/rest"
 )
 
-// KubeClient represents Kubernetes client and calculated namespace
+// KubeClient represents a Kubernetes client
 type KubeClient struct {
```
izgeri: does this still need to be a struct? why not `type KubeClient kubernetes.Interface`

doodlesbykumbi: @izgeri the struct is a wrapper with the convenience method `#getSecret`. e.g. it'd make it easier to add any logic to check if the environment is Openshift, or whatever else, without exposing all that in the provider. I suppose it could also just be a free method called `#getSecret` that takes `kc kubernetes.Interface` and `secretName string`

doodlesbykumbi: Agreed that would be cleaner. Alas, interfaces are a different type of animal and can't have methods at all^ :(

PR #488, opened 2018-10-17
Add metrics to benchmarks

izgeri:
- Auto-generates line of code counts and docker image size
- Formats as HTML to facilitate loading into website

### DIFF
--- a/bin/test_benchmarks
+++ b/bin/test_benchmarks
@@ -3,9 +3,15 @@
 set -eo pipefail
 
 project_dir=$PWD
+auto_bench_file="docs/_includes/auto_benchmarks.html"
+
 rm -f $project_dir/test/bench.output
 touch $project_dir/test/bench.output
 
+rm -f $project_dir/$auto_bench_file
+touch $project_dir/$auto_bench_file
+
+# test benchmarks in each directory
 pushd test
   exit_status="0"
   for dir in */; do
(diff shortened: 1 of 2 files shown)


bin/test_benchmarks
```diff
@@ -53,4 +57,43 @@ pushd test
     "
 popd || true
 
+# gather auto benchmark data for website
+
+# container size
+container_size=$(docker images \
+  | grep "cyberark\/secretless-broker[^-]" \
+  | awk '{ print $7 }')
```
sgnn7:
```
docker images -q --filter=reference='cyberark/secretless-broker' --format="{{ .Size }}"
```

Ideally we should be filtering by LABEL but we don't have one set in the final Dockerfile

bin/test_benchmarks
```diff
@@ -53,4 +57,43 @@ pushd test
...
+  | tee -a $project_dir/test/bench.output
+
+# lines of code count
+echo '<h2>Code Metrics</h2>
+<h3>Lines of Code</h3>
+<p>Code line counts were generated with
+<a href="https:github.com/hhatto/gocloc">gocloc</a> and are auto-updated
+with every build of this project.
+<pre>' \
+  | tee -a $project_dir/docs/_includes/auto_benchmarks.html
```
sgnn7: this `tee` command should be a variable to reduce duplication. Also might be good to consider turning it into a function.

bin/test_benchmarks
```diff
@@ -53,4 +57,43 @@ pushd test
...
+echo "<p>Docker image size: ${container_size}</p>" \
+  | tee -a $project_dir/docs/_includes/auto_benchmarks.html
+
+echo "Container size: $container_size\n" \
+  | tee -a $project_dir/test/bench.output
+
+# lines of code count
+echo '<h2>Code Metrics</h2>
+<h3>Lines of Code</h3>
+<p>Code line counts were generated with
```
sgnn7: This `<p>` tag is opened but I don't see it closed anywhere. Intended?

bin/test_benchmarks
```diff
@@ -6,6 +6,10 @@ project_dir=$PWD
 rm -f $project_dir/test/bench.output
 touch $project_dir/test/bench.output
 
+rm -f $project_dir/docs/_includes/auto_benchmarks.html
+touch $project_dir/docs/_includes/auto_benchmarks.html
```
sgnn7: Just a heads up that the rm/touch pattern is _really_ close to needing to be a loop but it's fine as-is

bin/test_benchmarks
sgnn7: Addendum: I feel like this is maybe ripe for having a template file and using that as the input for the html (or maybe just using markdown instead of pure html).

docs/.gitignore
```diff
@@ -5,4 +5,6 @@ _site/
 generated/*.html
 
 # Ignore the sass cache
-.sass-cache
\ No newline at end of file
+.sass-cache
+
+_includes/auto_benchmarks.html
```
sgnn7: Not required but would be good to add a comment why this file is ignored (even if it's rather obvious).

bin/test_benchmarks
izgeri: would you believe me if I said yes?

sgnn7:
We could improve this a touch with template files but no reason to keep this from getting merged right now.

Issue #463, opened 2018-09-28
PostgreSQL handler is updated to connect to backend via SSL


izgeri:
AC: 
- [ ] Based on the work done in #460, update the PostgreSQL handler to connect to the backend service via SSL
- [ ] PostgreSQL handler has integration test to verify SSL connection to backend
- [ ] Reference page for PostgreSQL Handler does not have alpha warning

izgeri:
Duplicate of #533 and duplicate of #535

Issue #460, opened 2018-09-28
A plan exists for updating the database handlers to use SSL for the backend connections


izgeri:
This story should be used to plan out how the PostgreSQL and MySQL handlers will be updated to enable communication with the backends via SSL. It will likely involve:

- [x] Digging into the handler code to understand how each of them currently works
- [x] Researching how MySQL clients typically implement SSL in the client-to-backend connection (the [MySQL protocol docs](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake_v10.html) may be helpful)
- [x] Researching how PostgreSQL clients typically implement SSL in the client-to-backend connection (the [PostgreSQL SSL docs](https://www.postgresql.org/docs/9.3/static/ssl-tcp.html) may be helpful)
- [x] Looking for OSS libraries (esp. in golang) that we can leverage in building this functionality (for example, the [Google CloudSQL Proxy](https://github.com/GoogleCloudPlatform/cloudsql-proxy) probably has some useful info)

The end result should be a solid plan for the path forward.

Plan: [Downstream-TLS-Support-for-Database-Handlers](https://github.com/cyberark/secretless-broker/wiki/Downstream-TLS-Support-for-Database-Handlers)

doodlesbykumbi:
## TLS Support for Database Handlers

### General approach
+ Extract sslmode params from credentials into backend connection options
+ Dial backend to get net.Conn
+ Default to insecure connection, regardless of support
+ Ensure client TLS requirements match server requirements and support, FAIL if not.
  + Default to insecure connection, regardless of support
  + Use TLS if server demands it
+ Upgrade net.Conn to tls.Conn and use appropriate strategy for sslmode

### Test cases

Given DB container services with TLS support and without:
+ each sslmode works or errors as intended
+ happy path to secretless via unix socket, and tcp

### TODO

+ update documentation for each db handler with descriptions of sslmode params in credentials

## MySQL
https://dev.mysql.com/doc/refman/8.0/en/using-encrypted-connections.html#using-encrypted-connections-client-side-configuration

### Handshake

https://dev.mysql.com/doc/internals/en/connection-phase.html

Note that throughout the handshake a sequenceID is maintained which is incremented at every step. This means a handshake with TLS-encryption will have a final sequenceID higher than a plain handshake.

1. Upon initial Dial the server responds to client with capabilities
2. Client creates HandshakeResponse with capabilities in common .e.g TLS
[...]

PR #491, opened 2018-10-18
DO NOT MERGE: reference implementation configurable TLS postgres

doodlesbykumbi:
Resolves https://github.com/cyberark/secretless-broker/issues/463

the idea is
+ parse backend connection options
+ dial backend to get net.Conn
+ check if the connections options for the backend service specify tls and how
+ if tls; check if the server supports ssl, fail if not
+ if tls; upgrade net.Conn to tls.Conn with relevant tls config, given backend connection options
+ business as usual

default tls behavior is `required`

References:

https://github.com/lib/pq/blob/master/ssl.go
https://github.com/CrunchyData/crunchy-proxy/tree/master/connect

for mysql https://github.com/go-sql-driver/mysql/blob/master/packets.go#L322-L335


izgeri:
@doodlesbykumbi which issue is this meant to help resolve? Is it #460?

test/pg_handler/etc/cert.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+BASEDIR=$(dirname "$0")
+LOGGING_PREFIX="gen_cert.sh >> "
+
```
jonahx: I wonder if we should be re-using something standard here.  eg, there are:

- https://github.com/conjurdemos/conjur-intro/tree/master/tools/simple-certificates
- https://github.com/conjurinc/appliance/tree/master/features/support/ca
- https://github.com/cyberark/conjur/blob/master/cucumber/api/features/support/ca_helpers.rb

we seem to be duplicating a lot of work around certs as a team...

test/pg_handler/etc/cert.sh
```diff
@@ -0,0 +1,62 @@
...
+echo "${LOGGING_PREFIX} Creating server certificate"
+openssl req -new -key ${BASEDIR}/server.key -out ${BASEDIR}/server.csr -subj "/C=UK/ST=/L=/O=IBM/OU=AIOS/CN=pg" -reqexts SAN -config <(cat /etc/ssl/openssl.cnf <(printf "[SAN]\nsubjectAltName=DNS:pg,DNS:localhost"))
+echo "${LOGGING_PREFIX} Server certificate signing request (${BASEDIR}/server.csr) is:"
+openssl req -verify -in ${BASEDIR}/server.csr -text -noout
+echo
+
+# use our CA certificate and key to create a signed version of the server certificate
+echo
+echo "${LOGGING_PREFIX} Signing server certificate using our root CA certificate and key"
+openssl x509 -req -sha256 -days 365 -in ${BASEDIR}/server.csr -CA ${BASEDIR}/rootCA.crt -CAkey ${BASEDIR}/rootCA.key -CAcreateserial -out ${BASEDIR}/server.crt -extensions SAN -extfile <(cat /etc/ssl/openssl.cnf <(printf "[SAN]\nsubjectAltName=DNS:pg,DNS:localhost"))
```
jonahx: the `<()` around `printf` shouldn't be needed.  can just do...

```
<(cat /etc/ssl/openssl.cnf; printf "[SAN]\nsubjectAltName=DNS:pg,DNS:localhost")
```

doodlesbykumbi:
@izgeri this PR came out of #460 but isn't the agreed upon deliverable (a plan). This was meant to be a proof of concept more than anything but it's ended up morphing into an actual implementation.

test/pg_handler/etc/cert.sh
doodlesbykumbi: agreed! in fact this whole script is not necessary. we only need the certs. those could be generated once and checked into source control.

izgeri:
Closing this PR, as content is being ported into PRs for #533 and #535

izgeri:
leaving the branch open for now in case content still needs to be grabbed from here. @doodlesbykumbi @jonahx please delete the branch once you're done with it

Issue #494, opened 2018-11-06
Remove dependency on Bash v4


izgeri:
Test script should not require Bash v4, since it is not installed by default on Macs.

jonahx:
I am going to split this up, since I'm not sure what's involved in fixing the mercurial dep atm.

izgeri:
@jonahx did this end up getting fixed? Wondering if we can close this story.

jonahx:
@izgeri yeah see above, it was merge here:

https://github.com/cyberark/secretless-broker/pull/497

izgeri:
Closed, since this was resolved in #497

PR #497, opened 2018-11-06
Remove bash4 dep 494

jonahx:
This remove the bash 4 dependency and partially resolves this issue:

https://github.com/cyberark/secretless-broker/issues/494


bin/test
```diff
@@ -3,15 +3,13 @@
...
-  [integration]="integration tests"
-  [quickstart]="Quick Start demo tests"
-)
+test_dirs=(unit integration quickstart)
+test_names=("unit tests" "integration tests" "quickstart demo tests")
 
-for test_key in "${!TESTS[@]}"; do
-  echo "Running ${TESTS[$test_key]}..."
-  $CURRENT_DIR/test_$test_key
+for i in "${!test_dirs[@]}"; do
```
sgnn7: Single-letter vars are a bit dicouraged. Maybe `test_dir_index`?

izgeri: aw, you beat me :P

jonahx: For an index I think `i` is preferred.  Example from the BashGuide:

https://mywiki.wooledge.org/BashGuide/Arrays

ctrl-F for "if we want to" for the relevant section.

sgnn7: While historically `i` is a well understood term (and Bash docs are generally of the older variety), I have found it to be of great value for variables to actually indicate what they hold and not just the type. This serves both as a readability aid as well as a bug prevention system.

I know this goes against a lot of the older languages docs but it's hard to explain without examples so I'll add what might jump to my mind when I see `i` vs something else in the block:
- `i` - It's an iterator, most likely an integer, probably starts at 0. I have no other context to make further assumptions.
- `test_dirs_index` - It's an iterator, very likely an integer, probably starts at zero, iterates over indexes of `test_dirs`. Anything else I know about `test_dirs` adds to the understanding of what this variable does.

jonahx: But... we're looping over integers.  `i` for looping over integers is about the most conventional, universal programming convention there is.

I take you're point if we're looping over something with meaning, like the _items_ of an array, but I disagree for looping over indexes.  By your reasoning every loop in C or JS or whatever would be:

```
for (index_of_thing = 0; index_of_thing < thing.length; index_of_thing++)
```

sgnn7:
LGTM!

PR #499, opened 2018-11-08
A bit of refactoring and cleanup 

jonahx:
1. Made one of our shell scripts compliant with shellcheck
2. Created a utility function for conditional debugging, and refactored a file with a lot of repeated `if` boilerplate to use it instead.

Jenkins:
https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--secretless-broker/detail/tiny-cleanup/7/pipeline

### DIFF
--- a/internal/pkg/util/util.go
+++ b/internal/pkg/util/util.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"log"
 	"net"
 
 	plugin_v1 "github.com/cyberark/secretless-broker/pkg/secretless/plugin/v1"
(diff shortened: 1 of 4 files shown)


internal/pkg/util/util.go
```diff
@@ -15,3 +16,19 @@ func Accept(l plugin_v1.Listener) (net.Conn, error) {
 	}
 	return conn, err
 }
+
+// Returns a function that will noop when debugEnabled=false, and will log
+// the given msg using `log.Print` or `log.Printf` (depending on the number
+// arguments given) debugEnabled=true
+func OptionalDebug(debugEnabled bool) func(string, ...interface{}) {
+	return func(msg string, args ...interface{}) {
+		if !debugEnabled {
```
doodlesbykumbi: Can we move this conditional into the parent function?  And conditionally return a noop function. Otherwise LGTM!

jonahx: Excellent point!

PR #515, opened 2018-11-22
MySQL Handler supports TLS defaulting to sslmode=require 

doodlesbykumbi:
Connected to #534 

The idea is:
+ Extract TLS configuration from backend connection options
+ Dial backend to get net.Conn
+ Kick off handshake
+ Send SSL Request and Upgrade

### TODOs
- [ ] Comprehensive support for TLS configurations
- [ ] FAIL when TLS requested but server does not have support
- [ ] Define sensible default TLS behaviour. Perhaps attempt TLS, if not supported do without
- [ ] Provide comprehensive configuration mechanism. Perhaps use [DSN](https://github.com/go-sql-driver/mysql#dsn-data-source-name) or however other mysql clients do it


internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -82,31 +89,80 @@ func (h *Handler) ConnectToBackend() (err error) {
...
+		tmp := packedHandshakeRespPacket[:(4+4+1+23)+4]
+		sslPacket := make([]byte, len(tmp))
+		copy(sslPacket, tmp)
+
+		// Update packet length for truncated packet
+		pktLen := len(sslPacket) - 4
+		sslPacket[0] = byte(pktLen)
+		sslPacket[1] = byte(pktLen >> 8)
+		sslPacket[2] = byte(pktLen >> 16)
+
```
jonahx: This stuff took my a long time to read and I had to go the links, and I still don't understand it all.  I think it should include more extensive comments.    In particular I'm confused  how the capability flags aren't being overwritten here.

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -1,13 +1,20 @@
...
 )
 
+// Various errors the handler might return
+var (
+	ErrNoTLS = errors.New("SSL connection error: SSL is required but the server doesn't support it")
+)
+
 // ConfigureBackend resolves the backend connection settings and credentials and sets the
 // BackendConfig field.
 func (h *Handler) ConfigureBackend() (err error) {
```
jonahx: The `values` map defined above this func and used within it should be renamed to `connection_details` or `options`, since it's a map, and not the values of a map.

test/mysql_handler/mysql_test.go
```diff
@@ -90,104 +110,81 @@ func TestMySQLHandler(t *testing.T) {
 
 	Convey("Connect over TCP", t, func() {
 
-		Convey("Without SSL", func() {
+		// Geri suggests: No TLS Upstream, TLS Downstream and sslmode default
+		//
+		Convey("No TLS Upstream, TLS Downstream and sslmode default", func() {
```
jonahx: This confused me, because "upstream" generally refers to the server, which in this case would be mysql.  It's still not clear to me what's going on.

test/mysql_handler/mysql_test.go
```diff
@@ -90,104 +110,81 @@ func TestMySQLHandler(t *testing.T) {
...
-			options := make(map[string]string)
-			_, err := net.LookupIP("secretless")
-			if err == nil {
-				host = "secretless"
-				port = 3306
-			} else {
-				host = "localhost"
-				port = 13306
-			}
+		Convey("Upstream SSL", func() {
```
jonahx: This is inside the "No TLS Upstream" block, adding to my confusion...

test/mysql_handler/mysql_test.go
```diff
@@ -90,104 +110,81 @@ func TestMySQLHandler(t *testing.T) {
...
 			options["--password"] = ""
 			flags := []string{"--ssl"}
 
-			_, err = mysql(host, port, "", []string{}, options, flags)
+			_, err := runTestQuery(host, port, "", []string{}, options, flags)
 
 			So(err, ShouldBeError)
 		})
+
+		Convey("sslmode default", func() {
```
jonahx: Also inside a block that says implies all of its cases are using "sslmode default", but this makes it seem that's not so, and only this one is...

PR #516, opened 2018-11-23
fix(k8s provider): use non-empty current namespace to fetch secrets

doodlesbykumbi:
Connected to #521

### DIFF
--- a/internal/app/secretless/providers/kubernetessecrets/client.go
+++ b/internal/app/secretless/providers/kubernetessecrets/client.go
@@ -1,29 +1,38 @@
 package kubernetessecrets
 
 import (
-	"k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
+	typedv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
-// NewKubeClient creates a Kubernetes API client using in-cluster config
-func NewKubeClient() (kubernetes.Interface, error) {
+// NewSecretsClient creates a SecretsClient for working with Secrets
+// in the pod namespace.
+// A SecretsClient uses an underlying Kubernetes API client
+// initialised with in-cluster config.
+func NewSecretsClient() (typedv1.SecretInterface, error) {
 	// creates the in-cluster config
-	config, err := rest.InClusterConfig()
+	rules := clientcmd.NewDefaultClientConfigLoadingRules()
+	overrides := &clientcmd.ConfigOverrides{}
+
+	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
+
+	config, err := clientConfig.ClientConfig()
 	if err != nil {
 		return nil, err
 	}
+
 	// creates the clientset
 	clientset, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		return nil, err
 	}
 
-	return clientset, nil
-}
+	// determine pod namespace
+	namespace, _, err := clientConfig.Namespace()
+	if err != nil {
+		return nil, err
+	}
 
-// GetSecret fetches a Secret with a given name
-func GetSecret(kc kubernetes.Interface, name string) (*v1.Secret, error) {
-	return kc.CoreV1().Secrets("").Get(name, metav1.GetOptions{})
+	return clientset.CoreV1().Secrets(namespace), nil
 }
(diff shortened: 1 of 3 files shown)


jonahx:
I was reading the entire test file and noticed this while I was there (not part of your PR):

```
So(err.Error(), ShouldEqual, "Kubernetes secret id must contain secret name and field name in the format secretName#fieldName, received 'foobar'")
```

We should not be asserting on copy/pasted strings.  We should either be using a custom error type (not familiar with how go does this or if it's idiomatic in go) or on a string exposed as a package level constant. 

Worth fixing this since we're in here anyway imo...

test/kubernetes_provider/kubernetes_provider_test.go
```diff
@@ -44,7 +45,7 @@ func TestKubernetes_Provider(t *testing.T) {
 		kubernetesProvider, ok = provider.(*kubernetessecrets.Provider)
 		So(ok, ShouldBeTrue)
 
-		kubernetesProvider.Client = testClientSet
+		kubernetesProvider.SecretsClient = testSecretsClient
```
jonahx: We shouldn't be putting "setup" code for the remaining _sibling_ tests inside of the first test.  Each test case should run independently.

It looks like the pattern GoConvey uses for shared setup is to place it into a container convey block, one level up from the tests being run: 

https://github.com/smartystreets/goconvey/blob/master/examples/bowling_game_test.go#L20

I still think we should remove Convey as a dep, but that's probably out of scope for this PR.

izgeri:
> I was reading the entire test file and noticed this while I was there (not part of your PR):
> 
> ```
> So(err.Error(), ShouldEqual, "Kubernetes secret id must contain secret name and field name in the format secretName#fieldName, received 'foobar'")
> ```
> We should not be asserting on copy/pasted strings. We should either be using a custom error type (not familiar with how go does this or if it's idiomatic in go) or on a string exposed as a package level constant.
> 
> Worth fixing this since we're in here anyway imo...
> 
> EDIT: so i looked into this a bit and typed errors are very much a thing in go and encouraged: https://blog.golang.org/error-handling-and-go (there is a simple square root of negative example about half way down)

@jonahx regarding your review comment - is that worth changing now? I don't disagree that named error types would help with a lot of stuff in the project, but I'd personally prefer defining what that should look like in general and implementing it consistently throughout the project, rather than ad-hoc adding something to this provider as part of the PR that we may later have to change if we decide on a different design / pattern. WDYT?

internal/app/secretless/providers/kubernetessecrets/client.go
```diff
@@ -1,29 +1,38 @@
...
+	typedv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
-// NewKubeClient creates a Kubernetes API client using in-cluster config
-func NewKubeClient() (kubernetes.Interface, error) {
+// NewSecretsClient creates a SecretsClient for working with Secrets
+// in the pod namespace.
+// A SecretsClient uses an underlying Kubernetes API client
+// initialised with in-cluster config.
```
jonahx: Despite this being a simple file, I'd like to see these comments expanded.  Specifically, 

1. What is an "in-cluster config"
2. An ~2-3 sentence overview of what the code does, explained assuming no familiarity with the underlying concepts.

I'd really like to start writing all new code with the POV of "this should be completely understandable to a new dev with no domain knowledge."  And familiarity with kubernetes secrets shouldn't be assumed for a dev on secretless.

jonahx:
> I'd personally prefer defining what that should look like in general and implementing it consistently throughout the project, rather than ad-hoc adding something to this provider as part of the PR that we may later have to change if we decide on a different design / pattern. WDYT?

@izgeri I think you're right :)

It's a codebase-wide concern and we can address it as such.  I just brought it up because I noticed it here.

jonahx:
> testclient "k8s.io/client-go/kubernetes/fake"

So per the conversation here:

https://conjurhq.slack.com/archives/CBTETAE20/p1543264914053200

The behavior of this confused me.  It looks like what's it's actually doing is hooking into the other packages so that, eg, this line:

https://github.com/cyberark/secretless-broker/blob/c97bfadc2248daf5d44a44a57fc7f484afa1df60/internal/app/secretless/providers/kubernetessecrets/provider.go#L22

is not triggering the actual execution of that real code.  It also looks like the fact that it does this isn't even documented:

https://godoc.org/k8s.io/client-go/kubernetes/fake

So I think we should document this in the top comments of the test.  Also, even though the behavior of the fake seems to save us this time, a better design in general (and more revealing of intent) would be to make the client injectable, and package private.  Then the test would inject the dep rather than mutate the already constructed field, as it does here:

https://github.com/cyberark/secretless-broker/blob/c97bfadc2248daf5d44a44a57fc7f484afa1df60/test/kubernetes_provider/kubernetes_provider_test.go#L48

The way to achieve this is with 2 constructors: 1 allowing an injected dep, and one taking no args which does nothing but call the first ctor with the default dep.

PR #517, opened 2018-11-23
fix(pipeline): add missing and fix broken tags

doodlesbykumbi:
* TODO: Pipeline stages really shouldn't rely on `secretless-*:latest`. We'll end up with race conditions on concurrent builds. A better alternative  is `secretless-*:$TAG`

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -21,19 +21,25 @@ else
 fi
 
 echo "Building secretless-broker:$TAG Docker image"
+# NOTE: the latest tag is required by downstream pipeline stages
 docker build -t "secretless-broker:${TAG}" \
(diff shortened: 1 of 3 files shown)


bin/build
```diff
@@ -22,18 +22,21 @@ fi
 
 echo "Building secretless-broker:$TAG Docker image"
 docker build -t "secretless-broker:${TAG}" \
+             -t "secretless-broker:latest" \
```
izgeri: Can we add comments here to say which downstream dependencies depend on the local latest images? It's not clear that those dependencies exist from this file, which led to me deleting things and breaking them once.

doodlesbykumbi: Done

Issue #518, opened 2018-11-24
Pipeline stages beyond build rely on the commit $TAG instead of latest


doodlesbykumbi:
Pipeline stages really shouldn't rely on `secretless-*:latest`. We'll end up with race conditions on concurrent builds. A better alternative is `secretless-*:$TAG`

- [ ] Update processes that run after the images are built to use the tagged image instead of the local `latest` image
- [ ] Update the gitlab build to use the tagged image instead of the local `latest` image
- [ ] Remove the `latest` tags from `bin/build`

szh:
This is now handled by the auto release process

PR #520, opened 2018-11-24
WIP - refactor: BASH_SOURCE for script location + prebaked pipeline container + .gitlab-ci.yml to independent script

doodlesbykumbi:
1. Improve portability of pipeline shell scripts. Use BASH_SOURCE over $0 to determine location of bash script. See https://stackoverflow.com/questions/35006457/choosing-between-0-and-bash-source

2. Improve developer experience by moving lengthy shell scripts from `.gitlab-ci.yml` to independent shell script

3. Improve gitlab-ci pipeline execution time by using docker image baked with all dependencies i.e. `bash`, `docker-compose`. This is `codebykumbi/docker-compose-bash:stable-git` for now until an official image is made. Dockerfile ->
    ```
    FROM docker:stable-git
    RUN apk add --no-cache bash
    RUN apk add --no-cache py-pip
    RUN pip install docker-compose
    ```

4.  ~~Fix gitlab pipeline by replace the cached `VERSION` tagged image in `.gitlab-ci.yml` with the new `$(version_tag)` tagged image~~. This tag isn't necessary for subsequent pipeline stages so it ought to be removed altogether as is done in the master branch :)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,10 +1,4 @@
-image: docker:stable-git
-
-variables:
-  IMAGE_TAGS: |
-    secretless-broker
-    secretless-dev
-    secretless-broker-quickstart
+image: codebykumbi/docker-compose-bash:stable-git
 
 stages:
   - build
@@ -99,68 +93,5 @@ website:
   only:
     - branches
 
-.utils: &utils |
-  # variables and functions
-  [[ "$TRACE" ]] && set -x
-  export CI_APPLICATION_REPOSITORY=$CI_REGISTRY_IMAGE/$CI_COMMIT_REF_SLUG
-  export CI_APPLICATION_TAG=$CI_COMMIT_SHA
-  export CI_CONTAINER_NAME=ci_job_build_${CI_JOB_ID}
-
-  export ALL_TAGS="";
-  VERSION=$(cat VERSION);
-
-  ALL_TAGS=$(echo "$IMAGE_TAGS" | sed '/^$/d' | while read -r line; do
-    echo "$line:latest"
-  done
-  )
-
-  function login_ci_registry() {
-    if [[ -n "$CI_REGISTRY_USER" ]]; then
-      echo "Logging to GitLab Container Registry with CI credentials..."
-      docker login -u "$CI_REGISTRY_USER" -p "$CI_REGISTRY_PASSWORD" "$CI_REGISTRY"
-      echo ""
-    fi
-  }
-
-  function setup_docker() {
-    login_ci_registry
-  }
-  
-  function save_image() {
-    tag=$(echo $1 | sed -e 's/\:/-/' -e 's/\//-/')
-    docker tag $1 "$CI_APPLICATION_REPOSITORY:$CI_APPLICATION_TAG-$tag"
-    docker push "$CI_APPLICATION_REPOSITORY:$CI_APPLICATION_TAG-$tag"
-  }
-  
-  function get_image() {
-    tag=$(echo $1 | sed -e 's/\:/-/' -e 's/\//-/')
-
-    docker pull "$CI_APPLICATION_REPOSITORY:$CI_APPLICATION_TAG-$tag"
-    docker tag "$CI_APPLICATION_REPOSITORY:$CI_APPLICATION_TAG-$tag" $1
-  }
-  
-  function save_images() {
-    echo "$ALL_TAGS" | while read -r line; do
-      save_image "$line"; 
-    done
-  }
-  
-  function optional_get_images() {
-    echo "$ALL_TAGS" | while read -r line; do
-      get_image "$line" 2> /dev/null || true; 
-    done
-  }
-  
-  function get_images() {
-    echo "$ALL_TAGS" | while read -r line; do
-      get_image "$line"; 
-    done
-  }
-
-  echo "loaded utils"
-
 before_script:
-  - apk add --no-cache bash
-  - apk add --no-cache py-pip
-  - pip install docker-compose
-  - *utils
+  - . ./bin/build_utils_gitlab
--- a/bin/build
+++ b/bin/build
@@ -4,11 +4,14 @@
 # usage: ./bin/build
 set -ex
 
-. bin/build_utils
-
-CURRENT_DIR=$($(dirname $0)/abspath)
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
 TOPLEVEL_DIR="$CURRENT_DIR/.."
 KEEP_ALIVE=${KEEP_ALIVE:-}
+
+# source build utilities
+pushd ${CURRENT_DIR}
+  . ./build_utils
+popd || true
 
 TAG="$(version_tag)"
 QUICK_START_DIR="$TOPLEVEL_DIR/demos/quick-start/docker"
--- a/bin/build
+++ b/bin/build
@@ -4,14 +4,14 @@
 # usage: ./bin/build
 set -ex
 
-CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
+CURRENT_DIR=$(cd $(dirname "${BASH_SOURCE[0]}"); pwd)
 TOPLEVEL_DIR="$CURRENT_DIR/.."
 KEEP_ALIVE=${KEEP_ALIVE:-}
 
 # source build utilities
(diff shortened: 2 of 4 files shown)


bin/build
```diff
@@ -4,12 +4,15 @@
...
-
-CURRENT_DIR=$($(dirname $0)/abspath)
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
 TOPLEVEL_DIR="$CURRENT_DIR/.."
 KEEP_ALIVE=${KEEP_ALIVE:-}
 
+# source build utilities
+pushd ${CURRENT_DIR}
+  . ./build_utils
+popd || true
```
sgnn7: `popd` should never fail if you got into the directory properly. If you're trying to hide the output, use `&>/dev/null`

bin/build_utils_gitlab
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash -e
+
+[[ "$TRACE" ]] && set -x
+
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
+
+# source build utilities
+pushd ${CURRENT_DIR}
+  . ./build_utils
+popd || true
```
sgnn7: Ditto on `popd` should never fail

bin/build
```diff
@@ -4,12 +4,15 @@
...
-. bin/build_utils
-
-CURRENT_DIR=$($(dirname $0)/abspath)
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
 TOPLEVEL_DIR="$CURRENT_DIR/.."
 KEEP_ALIVE=${KEEP_ALIVE:-}
 
+# source build utilities
+pushd ${CURRENT_DIR}
+  . ./build_utils
```
sgnn7: Why use the pushd instead of sourcing directly?

bin/build_utils_gitlab
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash -e
+
+[[ "$TRACE" ]] && set -x
+
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
+
+# source build utilities
+pushd ${CURRENT_DIR}
+  . ./build_utils
```
sgnn7: Ditto: Why use the pushd instead of sourcing directly?

bin/build_utils_gitlab
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash -e
+
+[[ "$TRACE" ]] && set -x
```
sgnn7: `[[ ]] && <something>`
should probably be
```
if [[ ]]; then
  <something>
fi
```
as the latter is much more readable

bin/publish
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash -e
 
-. bin/build_utils
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
+
+# source build utilities
+pushd ${CURRENT_DIR}
+  . ./build_utils
+popd || true
```
sgnn7: Ditto on sourcing directly and no need for `|| true`

bin/publish
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash -e
 
-. bin/build_utils
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
```
sgnn7: BASH_SOURCE is a lot more finicky and somewhat of a hack (though granted that it works in most cases):
```
BASH_SOURCE
    An array variable whose members are the source filenames where the
corresponding shell function names in the FUNCNAME array variable are
defined. The shell function ${FUNCNAME[$i]} is defined in the file 
${BASH_SOURCE[$i]} and called from ${BASH_SOURCE[$i+1]}
```
a


I don't think I have a problem with it but you really need to quote this line properly to avoid space escapes and make sure it passes something like shellcheck.

bin/build_utils_gitlab
```diff
@@ -0,0 +1,73 @@
...
+# source build utilities
+pushd ${CURRENT_DIR}
+  . ./build_utils
+popd || true
+
+CONTAINER_IMAGES="
+  secretless-broker
+  secretless-dev
+  secretless-broker-quickstart
+"
```
sgnn7: You should probably use an array here instead of a long string

bin/build_utils_gitlab
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash -e
+
+[[ "$TRACE" ]] && set -x
+
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
```
sgnn7: Ditto on comments on this in the other file

bin/build
```diff
@@ -4,12 +4,15 @@
 # usage: ./bin/build
 set -ex
 
-. bin/build_utils
-
-CURRENT_DIR=$($(dirname $0)/abspath)
+CURRENT_DIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd)
```
sgnn7: Ditto on comments about this in the other file

izgeri:
@doodlesbykumbi I'm wondering if we can close this PR. I was able to resolve the failing gitlab build by deleting the line that tried to save the `VERSION` tagged image to the gitlab registry (in #522)

I've updated your #518 to include a list of AC that would involve updating the gitlab pipeline to use the tagged image instead of `latest`, and I expect that changes similar to those in this PR can be used to resolve that issue when we get to it. Among the things that we have on our schedule for the next two weeks, however, that is lower priority imo.

.gitlab-ci.yml
```diff
@@ -1,10 +1,4 @@
-image: docker:stable-git
-
-variables:
-  IMAGE_TAGS: |
-    secretless-broker
-    secretless-dev
-    secretless-broker-quickstart
+image: codebykumbi/docker-compose-bash:stable-git
```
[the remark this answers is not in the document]
izgeri: Agreed, I wouldn't merge as-is.

bin/build
doodlesbykumbi: addressed

bin/publish
doodlesbykumbi: addressed. thanks, the quote is definitely necessary

bin/publish
doodlesbykumbi: addressed

bin/build_utils_gitlab
doodlesbykumbi: addressed

bin/build_utils_gitlab
doodlesbykumbi: true

bin/build_utils_gitlab
doodlesbykumbi: this is useful in IDEs for being able to find dependencies in `./build_utils`. without this relative path most IDEs won't be able to resolve the deps

bin/build_utils_gitlab
doodlesbykumbi: agreed

bin/build_utils_gitlab
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash -e
+
+if [[ "${TRACE}" ]]; then
```
sgnn7: This needs to be compared to something I think otherwise it's going to trigger if TRACE has any value:
```
$ TRACE=false
$  if [[ "${TRACE}" ]]; then echo "set"; else echo "not set"; fi
set

$ TRACE=0
$  if [[ "${TRACE}" ]]; then echo "set"; else echo "not set"; fi
set
```

bin/build_utils_gitlab
```diff
@@ -0,0 +1,63 @@
...
+fi
+
+CONTAINER_IMAGES=(
+  secretless-broker
+  secretless-dev
+  secretless-broker-quickstart
+)
+export REQ_CONTAINER_IMAGE_TAGS=();
+for CONTAINER_IMAGE in ${CONTAINER_IMAGES[@]}; do
+    REQ_CONTAINER_IMAGE_TAGS+=("${CONTAINER_IMAGE}:latest");
```
sgnn7: 2 space indent needed

.gitlab-ci.yml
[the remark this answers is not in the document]
sgnn7: @doodlesbykumbi This is still not the right image.

PR #524, opened 2018-11-26
Add PR template

izgeri:
Connected to #525

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,17 +3,17 @@
 Connected to [relevant GitHub issues, eg #76]
 #### Where should the reviewer start?
 #### What is the status of the manual tests?
-For any new manual tests that should be run to test this feature, have you created/updated a folder in `test/manual` that includes:
-- [ ] An updated README with instructions on how to manually test this feature
-- [ ] Utility `start` and `stop` scripts to spin up and tear down the test environments
-- [ ] A `test` script to run some basic manual tests (optional; if does not exist, the README should have detailed instructions)
-
-In addition, have you run the following manual tests to verify existing functionality continues to function as expected?
+Have you run the following manual tests to verify existing functionality continues to function as expected?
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+#### What does this PR do?
+#### Any background context you want to provide?
```
jonahx: Since big PR templates can be scary/demotivating, I'd combine this with first:
```
#### What does this PR do? (include background context, if relevant)
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+#### What ticket does this PR close?
+Contributes to [relevant GitHub issues, eg #76]
+#### Where should the reviewer start?
+#### How should this be manually tested?
```
jonahx: I'd like to have a system, just like the `start` `stop` `test` system we use for automated tests.  So the question could even be "Does this include a new folder in the `manual` test dir?", and there would be a detailed README, per our system.

Rather than putting that knowledge in the PR description.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+- [ ] Manually run [Kubernetes-Conjur demo](https://github.com/conjurdemos/kubernetes-conjur-demo) with a local Secretless Broker image build of your branch
+- [ ] Manually run the [full demo](https://github.com/cyberark/secretless-broker/tree/master/demos/full-demo) (optional)
+#### Screenshots (if appropriate)
+#### Link to build in Jenkins / GitLab (if not already visible on PR)
+#### Questions:
+> Does this work have automated integration and unit tests (if not, link to relevant GH issues to add them)?
+
+> Has this change been documented (Readme, docs, etc. - if not, link to relevant GH issues to add docs)?
+
+> Can we make a blog post, video, or animated GIF of this?
```
jonahx: I think I'd prefer to make these all `####` as well if we care about them.  They're not really related and all of the other ones are questions too.... I know we do this in our other templates but I don't think it makes sense, actually.

PR #529, opened 2018-11-27
refactor(pipeline): allow ./bin/test_integration to specify individual test_folders

doodlesbykumbi:

`./bin/test_integration` currently runs all the tests and has no mechanism for customisation.
The ability to specify individual test_folders and the local flag would improve developer experience.
Currently, the developer has to run the test steps manually.

~None. Small and spur of the moment.~
Connected to #530

`./bin/test_integration`

1. Run `./bin/test_integration` and ensure it runs as usual and covers all the tests. This is also confirmed by the green build.
2. Run `./bin/test_integration -x`, it should complain about an invalid option and provide help and exit with 1
3. Run `./bin/test_integration --help`, it should provide help and exit with 0
4. Add a dummy test to `kubernetes_provider`
5. Run `./bin/test_integration kubernetes_provider`, it should only run the previously baked into the container image `kubernetes_provider` tests
6. Run `./bin/test_integration -l kubernetes_provider`, the new test should be included in the output

N/A, this doesn't touch the source

### DIFF
--- a/bin/test_integration
+++ b/bin/test_integration
@@ -2,13 +2,46 @@
 
 set -eo pipefail
 
+help_and_exit() {
+  local retval=${1:-1}
+  cat <<EOF
+${0##*/} [-l|--local] [test_folder1] [test_folder2] [test_folder3] ...
+
+Runs suite of integration tests by calling ./start and ./stop scripts, if found, for subdirectories in the test subdirectory of the project root.
+Names of subdirectories of the project root can be optionally specified as arguments to limit the subdirectories whose test cases will be run.
+
+Options:
+  -l, --local sets the local flag when the test is called in the test folder
+EOF
+  exit "$retval"
+}
+
+non_flag_args=( )
+test_flag="";
+while (( $# )); do
+  case $1 in
+    -h|--help)   help_and_exit 0 ;;
+    -l|--local)  test_flag="-l" ;;
+    -*)          echo "Unknown option: $1";
+                 echo "";
+                 help_and_exit 1 ;;
+    *)           non_flag_args+=( "$1" ) ;;
+  esac
+  shift
+done
+
 project_dir=$PWD
-rm -f $project_dir/test/junit.output
-touch $project_dir/test/junit.output
+rm -f "${project_dir}/test/junit.output"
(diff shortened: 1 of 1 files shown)


bin/test_integration
```diff
@@ -2,13 +2,46 @@
...
+    -l|--local)  test_flag="-l" ;;
+    -*)          echo "Unknown option: $1";
+                 echo "";
+                 help_and_exit 1 ;;
+    *)           non_flag_args+=( "$1" ) ;;
+  esac
+  shift
+done
+
+test_folders=*/;
```
jonahx: See: https://github.com/koalaman/shellcheck/wiki/SC2125

I believe you want `test_folders=( */ )`.  The current code would, I believe, break further down in the script if dir names had spaces.

bin/test_integration
```diff
@@ -2,13 +2,46 @@
...
+if [[ ${#non_flag_args[@]} -gt 0 ]]; then
+  test_folders=${non_flag_args}
```
jonahx: See https://github.com/koalaman/shellcheck/wiki/SC2128

Again, this sort of works by accident now because the array only _has_ one element (see comment above), but better to use arrays properly here.

bin/test_integration
```diff
@@ -2,13 +2,46 @@
...
+fi
+
 project_dir=$PWD
 rm -f $project_dir/test/junit.output
 touch $project_dir/test/junit.output
```
jonahx: Both of these need double quotes per: https://github.com/koalaman/shellcheck/wiki/SC2086

bin/test_integration
```diff
@@ -21,7 +54,7 @@ pushd test
 
       # Run the tests
       set +e
-        ./test | tee -a $project_dir/test/junit.output
+        ./test "${test_flag}" | tee -a $project_dir/test/junit.output
```
jonahx: double quotes around `$project_dir` here too, again per https://github.com/koalaman/shellcheck/wiki/SC2086

doodlesbykumbi:
@jonahx Thanks for the review. Comments addressed. I agree that shellcheck is a definite must!

PR #531, opened 2018-11-28
DO NOT MERGE: plan for TLS Support for Database Handlers

doodlesbykumbi:

Plan for downstream TLS Support on Database Handlers

Connected to #460

### DIFF
--- a/docs/docs/reference/handlers/postgres.md
+++ b/docs/docs/reference/handlers/postgres.md
@@ -10,7 +10,7 @@ permalink: docs/reference/handlers/postgres.html
 The PostgreSQL handler authenticates incoming connections for a particular
 listener.
 
-**Handler in alpha - not suitable for production** 
+**WARNING: Handler in alpha - not suitable for production**
 
 ### Credentials
 - `address`  
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,11 +13,15 @@ Thanks for your interest in the Secretless Broker. Before contributing, please t
 
 Our [Waffle.io][waffle] is synchronized with GitHub and helps you navigate this workflow more easily.
 
+In addition to technical workflow descriptions available in Waffle / GitHub,
+some of the project's technical design documents can be found in the project [design][design] folder.
+
 ## Style guide
 
 Use [this guide][style] to maintain consistent style across the Secretless Broker project.
 
+[design]: https://github.com/cyberark/secretless-broker/tree/master/design
+[issues]: https://github.com/cyberark/secretless-broker/issues
 [style]: STYLE.md
 [tests]: README.md#testing
 [waffle]: https://waffle.io/cyberark/secretless
-[issues]: https://github.com/cyberark/secretless-broker/issues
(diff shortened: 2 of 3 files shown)


tls-support-plan.md
```diff
@@ -0,0 +1,81 @@
...
++ Extract sslmode params from credentials into backend connection options
+  + sslmode: Provide different levels of protection.
+  + sslca: The Certificate Authority (CA) certificate file. This option, if used, must specify the same certificate used by the server
+  + sslcert: The client public key certificate file.
+  + sslkey: The client private key file.
+  + sslcrl: The certificates revoked by certificate authorities.
++ Dial backend to get net.Conn
++ Default to secure connection if TLS supported
++ Ensure client TLS requirements match server requirements and support, FAIL if not.
+  + Default to insecure connection, regardless of support
```
izgeri: is this still correct?

tls-support-plan.md
```diff
@@ -0,0 +1,81 @@
...
+  + Use TLS if server demands it
++ Upgrade net.Conn to tls.Conn and use appropriate strategy for sslmode
```
izgeri: this step seems like it could be the bulk of the work. is it worth describing in more detail? or is what is described in the handler-specific sections below sufficient?

tls-support-plan.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Test cases
+
+Given DB container services with TLS support and without:
++ each sslmode works or errors as intended
```
izgeri: is it worth writing out exactly what this means right now?

docs/docs/reference/handlers/postgres.md
```diff
@@ -7,26 +7,136 @@ permalink: docs/reference/handlers/postgres.html
...
+  + `require` (default)
+  only try an SSL connection. If a root CA file is present, verify the certificate in the same way as if verify-ca was specified
+
+  + `verify-ca`
+  only try an SSL connection, and verify that the server certificate is issued by a trusted certificate authority (CA).
+
+  + `verify-full`
+  only try an SSL connection, verify that the server certificate is issued by a trusted CA and that the server host name matches that in the certificate
+
+_NOTE_: If `sslmode` is set to `require`, `verify-ca`, or `verify-full`, it may
```
izgeri: is it possible / sensible to break out the stories for adding this functionality into an initial opinionated config, and gradually add in options to specify some of the items below? at current it seems very flexible and configurable, but I'm wondering if that will also mean it is harder to implement in one go.

tls-support-plan.md
jonahx: I don't think @doodlesbykumbi has updated this yet....

tls-support-plan.md
doodlesbykumbi: I think the handler specific stuff below is sufficient for describing the handshakes. configuring TLS using the appropriate strategy is Golang thing more than anything else

docs/docs/reference/handlers/postgres.md
doodlesbykumbi: Yup. Latest `Stories` section shows the stories each handler should have in agreement with what you wrote

tls-support-plan.md
doodlesbykumbi: Stories provide more detail

PR #571, opened 2018-12-08
WIP: pg-handler supports sslmode up verify ca

doodlesbykumbi:


internal/app/secretless/handlers/pg/backend.go
```diff
@@ -52,6 +52,15 @@ func (h *Handler) ConfigureBackend() (err error) {
 	delete(values, "username")
 	delete(values, "password")
 
+	// TODO: remove because of hack
```
izgeri: maybe remove this line now?

internal/app/secretless/handlers/pg/backend.go
```diff
@@ -52,6 +52,15 @@ func (h *Handler) ConfigureBackend() (err error) {
...
+
+	// TODO: what are options for, it's weird that any additonal
```
izgeri: maybe this should be a gh issue instead of a comment? I've had this thought before too, but it's going to get lost here

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
+package protocol
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net"
+)
+
+type options map[string]string
```
jonahx: This threw me a bit, because we _call_ it like this in backend.go:
```go
tlsConf, err := protocol.ResolveTLSConfig(h.BackendConfig.QueryStrings)
```
Yet here we have a special type, which is private, called `options`.  Ofc it works because they're both just `map[string]string`.  But I'd say that you _either_ want to just omit this type, and use the `map[string]string` everywhere, _or_ expose the `protocol.Options` type, and use that at the calling site as well.

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+
+type TLSConfigWrapper struct {
+	UseTLS bool
+	tls.Config
+	VerifyCaOnly bool
+	Options options
+}
```
jonahx: I think, as a general rule, the name `XXXWrapper` indicates a problem.  It's telling you: I don't know really know what this thing _is_, so I'll name it after a thing that describes the way I'm _implementing_ it.

I _think_ in this case the missing concept is `SecretlessTLSConfig`....

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+
+func ResolveTLSConfig(o options) (TLSConfigWrapper, error) {
```
jonahx: I'm not a fan of the word "resolve" in general (there are better alternatives) but in this case it seems inaccurate as well: "resolving" implies to me a process of looking up, via a network or db, as in "the DNS name could not be resolved."  

Here we simply have a pure function that takes options and returns a `TLSConfigWrapper`.  That is, we have a constructor.  Let's stick with the convention and name it `NewTLSConfigWrapper(options)`.

*This comment is independent of my above suggestion that `TLSConfigWrapper` itself should be renamed :)

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+func sslVerifyCertificateAuthority(client *tls.Conn, tlsConf *tls.Config) error {
+	err := client.Handshake()
+	if err != nil {
+		return err
+	}
+	certs := client.ConnectionState().PeerCertificates
+	opts := x509.VerifyOptions{
+		DNSName:       client.ConnectionState().ServerName,
+		Intermediates: x509.NewCertPool(),
+		Roots:         tlsConf.RootCAs,
```
jonahx: The argument `tlsConf` is only used for the `RootCAs` member it contains.  We should instead pass in only the `RootCAs` as an argument.

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	for i, cert := range certs {
+		if i == 0 {
```
jonahx: What is special about the index `0` here?  This becomes dark knowledge without a comment or other semantic clarification.

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+			continue
+		}
+		opts.Intermediates.AddCert(cert)
+	}
+	_, err = certs[0].Verify(opts)
+	return err
+}
```
jonahx: We want to clarify what is special about the index `0` here... rather than making the reader figure it out.  After line 135, it might make sense to set `[0]` to a meaningful variable name, and set the tail to.... `intermediateCerts`

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return err
+	}
+
+	tlsConf.Certificates = []tls.Certificate{cert}
+	return nil
+}
+
+// sslCertificateAuthority adds the RootCA specified in the "sslrootcert" setting.
+func sslCertificateAuthority(tlsConf *tls.Config, o options) error {
```
jonahx: Since this is not returning a CertificateAuthority, and this implicitly does validation, I think `addRootCA`would be a more accurate name.

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+			return nil, err
+		}
+	}
+
+	return client, nil
+}
+
+// sslClientCertificates adds the certificate specified in the "sslcert" and
+// "sslkey" settings
+func sslClientCertificates(tlsConf *tls.Config, o options) error {
```
jonahx: As is, `addClientCertsToConfiguration` would be a more accurate name, since this function mutates `tlsConf` rather than returning the certificates, as I'd expect given the current name and the convention of naming returning funcs after what the return.

That said, in this case there's no good reason to pass in `tlsConf` afaict (in contrast the the funcs below), so I'd consider keeping the current name, but having the func do nothing but _return_ the `[]tls.Certificate{cert}`.  The calling code could then do:
```go
tlsConf.Certificates = clientCertificates(options)
```

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
...
+	default:
+		// TODO add verify-full below
+		return TLSConfigWrapper{}, fmt.Errorf(`unsupported sslmode %q; only "require" (default), "verify-ca", and "disable" supported`, mode)
+	}
+
+	return tlsConf, nil
+}
+
+// HandleSSLUpgrade upgrades a net.Conn using TLSConfigWrapper
+func HandleSSLUpgrade(connection net.Conn, tlsConf TLSConfigWrapper) (net.Conn, error) {
```
jonahx: I'd name this `SSLClientConnection` since it returns a client connection.

internal/app/secretless/handlers/pg/backend.go
```diff
@@ -72,11 +86,64 @@ func (h *Handler) ConnectToBackend() (err error) {
 	debug := util.OptionalDebug(h.GetConfig().Debug)
 	debug("Sending startup message")
 
-	connection, err = ssl(connection, h.BackendConfig.QueryStrings)
+	tlsConf, err := protocol.ResolveTLSConfig(h.BackendConfig.QueryStrings)
 	if err != nil {
 		return
 	}
 
+	if tlsConf.UseTLS {
```
jonahx: I'd pull this whole part out into a separate function.

internal/app/secretless/handlers/pg/protocol/ssl.go
```diff
@@ -0,0 +1,149 @@
+package protocol
```
jonahx: General comment that relates to this file and to backend.go... I'm writing this after having written all the other comments, and it's something larger that's still bothering me...

The "division of labor" among these files isn't clear... it's doesn't feel intuitive.  What are the responsibilities _really_?  This could use more design and thought imo, because it doesn't "stick" (and the names don't match with things really are).  I find myself hopping back to definitions a few minutes... even 30s after I've read them.

A few examples... what is currently called `ResolveTLSConfig` appears to be a mapping from user supplied "QueryStrings" to options in the native `tlsConf` config object.  That is, it's essentially a pure transformation function, and it's related to user input as it comes into secretless.  Yet it's currently house in the `protocol.ssl.go`, and I'd say it's a stretch to call this function part of the ssl protocol.  

OTOH, in `pg.backend.go`, we have snippets like this:

```go
		message := protocol.NewMessageBuffer([]byte{})
		message.WriteInt32(8)
		message.WriteInt32(protocol.SSLRequestCode)
```
which is a byte level details of the pg ssl protocol.

There are other similar bit of confusion.  I don't have a solution for this yet, but I think it deserves more thought and should be solved.  Otherwise it's very hard to form a proper mental model of the code.

test/util/test/types.go
```diff
@@ -28,6 +28,43 @@ func ServerTLSTypeValues()[]ServerTLSType {
 	return []ServerTLSType{TLS, NoTLS}
 }
 
+func (tlsType ServerTLSType) toConfigVariables(dbConfig TestDBConfigType) []config.Variable  {
```
jonahx: Not part of this PR, but we need to change `ListenerType` to `ProtocolType`.

test/mysql_handler/pkg/run_test_case.go
```diff
@@ -9,11 +10,26 @@ import (
...
+	if clientConfig.SSL != nil && *clientConfig.SSL {
+		args = append(args, "--ssl", "--ssl-verify-server-cert=TRUE")
+	}
+	if clientConfig.Username != nil {
+		args = append(args, fmt.Sprintf("--user=%s", *clientConfig.Username))
+	}
+	if clientConfig.Password != nil {
+		args = append(args, fmt.Sprintf("--password=%s", *clientConfig.Password))
+	}
+	switch connectPort.ListenerType {
```
jonahx: The need for this switch statement means we still haven't found the concepts with the right "natural joints".  I'll return to this tomorrow and elaborate.

Issue #629, opened 2019-02-05
A method exists for encoding HandshakeV10 as bytes


doodlesbykumbi:
To aid the refactor effort `HandashakeV10`, like all other domain specific entities, should have a utility method for encoding to bytes

PR #630, opened 2019-02-06
WIP - a method exists for encoding mysql HandshakeV10 as bytes

doodlesbykumbi:
Connected to #629 [relevant GitHub issues, eg #76]


internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -126,6 +116,19 @@ func (h *Handler) ConnectToBackend() (err error) {
 		return
 	}
 
+	// Remove Client SSL Capability from Server Handshake Packet
+	serverHandshake.ServerCapabilities &^= protocol.ClientSSL
```
jonahx: In general, clients (in the code sense) should never be manipulating object member data themselves.  Instead, we should make a method to do it.  So in this case, we could write:
```go
serverHandshake.removedClientSSL()
```
and the bit masking logic would then be hidden from client code.

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -126,6 +116,19 @@ func (h *Handler) ConnectToBackend() (err error) {
...
+	packetWithNoSSL, err := protocol.PackHandshakeV10(serverHandshake)
```
jonahx: Same idea here:

```go
serverHandshake.pack()
```

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -101,29 +101,31 @@ func (h *Handler) ConnectToBackend() (err error) {
...
 		return err
 	}
 
-	// Remove Client SSL Capability from Server Handshake Packet
-	packetWithNoSSL, err := protocol.RemoveSSLFromHandshakeV10(packet)
+	// STEP: ServerGreeting. Secretless => Client
+	//////////////////////////////////////////////
+
+	// Unpack server packet
+	serverHandshake, err := protocol.UnpackHandshakeV10(serverHandshakePacket)
```
jonahx: since this a ctor, let's change the name to `NewServerHandshake(rawPkt)`

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -101,29 +101,31 @@ func (h *Handler) ConnectToBackend() (err error) {
...
+	serverDoesntSupportSSL := serverHandshake.ServerCapabilities & protocol.ClientSSL == 0
```
jonahx: change to method: `serverHandshake.supportsSSL()`

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -251,11 +253,12 @@ func (h *Handler) ConnectToBackend() (err error) {
...
 // STEP: LoginResponse. Backend => Secretless
 	// Proxy server response
-	if packet, err = h.BackendRead(); err != nil {
+	loginResponsePacket, err := h.BackendRead();
+	if err != nil {
 		return
 	}
 
-	if _, err = protocol.UnpackOkResponse(packet); err != nil {
+	if _, err = protocol.UnpackOkResponse(loginResponsePacket); err != nil {
```
jonahx: Same here, create a ctor `NewServerLoginResponse`

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -130,14 +130,101 @@ func UnpackOkResponse(packet []byte) (*OkResponse, error) {
...
+	AuthPlugin             string
+	Salt                   []byte
 }
 
+// RemoveClientSSL removes Client SSL Capability from Server Handshake
+func (serverHandshake *HandshakeV10) RemoveClientSSL() {
+	serverHandshake.ServerCapabilities &^= ClientSSL
+}
+
+func (serverHandshake *HandshakeV10) Pack() ([]byte, error) {
```
jonahx: I'd like to see an ascii table above here that shows the structure of the packet at the byte level.  That way, I could match up each of the steps of the packing to this "picture" and both better understand and verify the code.  As is, the process of verification of this code would be to find the relevant docs myself on mysql code, and then go through them to find the relevant info about the packet structure, and then do the mapping between that and the code myself.

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -130,14 +130,101 @@ func UnpackOkResponse(packet []byte) (*OkResponse, error) {
...
+	if err := WriteNullTerminatedString(&buf, serverHandshake.ServerVersion); err != nil {
+		return nil, err
+	}
+
+	// Write Connection ID
+	connectionIDBuf := make([]byte, 4)
+	binary.LittleEndian.PutUint32(connectionIDBuf, serverHandshake.ConnectionID)
+	buf.Write(connectionIDBuf)
+
+	// Write AuthnPluginDataPart1
```
jonahx: I have a similar comment for most of these comments, which is that I'd like more info inline (just a one sentence, assume no prior knowledge, explanation, with an example if relevant).  Also a link in each case to the relevant docs.

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -130,14 +130,101 @@ func UnpackOkResponse(packet []byte) (*OkResponse, error) {
...
+	buf.Write(serverHandshake.Salt[:8])
+
+	// Filler
```
jonahx: Why do we need this?  Again, this might be clear  if the packet structure picture was above.

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -130,14 +130,101 @@ func UnpackOkResponse(packet []byte) (*OkResponse, error) {
...
+	buf.WriteByte(0)
+
+	// write ServerCapabilities
```
jonahx: What are these?  Link to docs + a few examples.

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -130,14 +130,101 @@ func UnpackOkResponse(packet []byte) (*OkResponse, error) {
...
+	buf.Write([]byte{
+		byte(serverHandshake.ServerCapabilities),
+		byte(serverHandshake.ServerCapabilities >> 8),
+	})
+
+	// Write ServerDefaultCollation
```
jonahx: Link to docs plus simple definition to save future readers googling.

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -130,14 +130,101 @@ func UnpackOkResponse(packet []byte) (*OkResponse, error) {
...
+	buf.WriteByte(serverHandshake.ServerDefaultCollation)
+
+	// Write StatusFlags
```
jonahx: Same for this and all comments below

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -673,6 +699,14 @@ func ReadNullTerminatedString(r *bytes.Reader) string {
 	}
 }
 
+func WriteNullTerminatedString(buf *bytes.Buffer, contents string) error {
```
jonahx: is there a std lib func for this?  seems like there might be...

Issue #627, opened 2019-02-01
MySQL handler itself has clear code


jonahx:
This is the task for the changes described here:

https://gist.github.com/jonahx/3a400bf335d546d7750804d5057d6262

That is, it is for refactoring the handler and backend files for clarity.

PR #632, opened 2019-02-07
Refactor MySQL handler for readability and consistency

jonahx:
This is large refactor, but most of it was in service of clarifying the
high-level logic of the authentication handshake:

1. Splitting that logic into its own object: `AuthenticationHandshake`
2. Changing the responsibilities `Handler.Run()` to reflect those
   changes.
3. Creating new types to make the above refactor possible and clear.
4. Moving new and existing types into their own files.
5. Renaming a bunch of methods and types for clarity.
6. Fixing many check style errors, notably adding correct Go style
   comments to all exported methods and types.

Connected to #627


internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -70,273 +118,310 @@ func (h *Handler) ConfigureBackend() (err error) {
...
-func (h *Handler) ConnectToBackend() (err error) {
+func (mc *ConnectionPhase) Run() error {
+	mc.readServerHandshake()
+	mc.writeHandshakeToClient()
+	mc.validateServerSSL()
+	mc.readClientHandshakeResponse()
+	mc.overrideClientCapabilities()
+	mc.injectCredentials()
+	mc.maybeUpgradeToSSL()
+	mc.writeClientHandshakeResponseToBackend()
```
izgeri: the sketch at the top of backend.go should be updated to reflect this flow - there seems to be a slight mismatch right now

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -70,273 +118,310 @@ func (h *Handler) ConfigureBackend() (err error) {
...
+	)
+}
+
+// TODO: All of this should be moved to kumbi's code
+func (mc *ConnectionPhase) maybeUpgradeToSSL() {
+	if mc.err != nil {
+		return
+	}
+
+	if !mc.clientRequestedSSL() {
```
izgeri: why not just call the method `upgradeToSSL` and move this check to `ConnectionPhase.Run()`? I don't love the use of `maybe` in a method name

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
+}
+
+type MySQLConnection struct {
+	conn net.Conn
+	sequenceId byte
+}
+
+
+func NewMySQLClientConnection(conn net.Conn) *MySQLConnection {
+	return &MySQLConnection{sequenceId: 0, conn: conn}
```
izgeri: justification for the sequenceId here?

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
+}
+
+func NewMySQLBackendConnection(conn net.Conn) *MySQLConnection {
+	return &MySQLConnection{sequenceId: 1, conn: conn}
```
izgeri: same as above - why is the sequenceId 1 here?

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
 	Host     string
 	Port     uint
 	Username string
 	Password string
 	Options  map[string]string
 }
 
-func (bc *BackendConfig) Address() string {
-	return bc.Host + ":" + strconv.FormatUint(uint64(bc.Port), 10)
+// TODO: all these comments will be moved and subsumed by kumbi's work
```
izgeri: it's not clear to me here which parts are going to be taken care of by KT's changes. are we just referring to `mysqlPacket` here?

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
 }
 
 // Handler connects a client to a backend. It uses the handler Config and Providers to
-// establish the BackendConfig, which is used to make the Backend connection. Then the data
+// establish the ConnectionDetails, which is used to make the Backend connection. Then the data
 // is transferred bidirectionally between the Client and Backend.
 //
 // Handler requires "host", "port", "username" and "password" credentials.
+//
 type Handler struct {
```
izgeri: Since this is the main type defined in this file (handler.go) it should be at the top of the doc, with its methods defined immediately underneath.

Other types currently in this file (`ConnectionDetails`, `mysqlPacket`, and `MySQLConnection`) should be defined either in a separate file that is part of the same package or defined below and organized in sections by type in order of decreasing importance, with the type definition at the top and the methods for that type below its definition

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -11,34 +70,23 @@ import (
...
-var (
-	ErrNoTLS = errors.New("SSL connection error: SSL is required but the server doesn't support it")
-)
-
 // ConfigureBackend resolves the backend connection settings and credentials and sets the
-// BackendConfig field.
-func (h *Handler) ConfigureBackend() (err error) {
-	result := BackendConfig{Options: make(map[string]string)}
+// ConnectionDetails field.
+func (h *Handler) fetchConnectionDetails() (result *ConnectionDetails, err error) {
```
izgeri: this belongs in handler.go

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
+
+func (pkt *mysqlPacket) SequenceId() byte {
+	return (*pkt)[3]
+}
+
+func (pkt *mysqlPacket) SetSequenceId(id byte) {
+	(*pkt)[3] = id
+}
+
+type MySQLConnection struct {
```
izgeri: the package is called mysql - why not just have this be `connection`?

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
+	conn net.Conn
+	sequenceId byte
+}
+
+
+func NewMySQLClientConnection(conn net.Conn) *MySQLConnection {
```
izgeri: why not just have one method `NewConnection` that accepts a sequenceId parameter?

izgeri: though there is something about this that's off - why is the sequenceId so intricately tied to the connection?

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
+//  |    Type     |      Name      |                 Description                 |
+//	+-------------+----------------+---------------------------------------------+
+//  | int<3>      | payload_length | Length of the payload. The number of bytes  |
+//  |             |                | in the packet beyond the initial 4 bytes    |
+//  |             |                | that make up the packet header.             |
+//  | int<1>      | sequence_id    | Sequence ID                                 |
+//  | string<var> | payload        | [len=payload_length] payload of the packet  |
+//  +-------------+----------------+---------------------------------------------+
+//
+type mysqlPacket []byte
```
izgeri: the package is called mysql - why not just have this be `packet`?

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -11,48 +11,127 @@ import (
...
+		mysqlError = genericProtocolError(err)
+	}
+
+	// TODO: log something if this fails
+	//
+	h.mySQLClientConn.write(mysqlError.GetMessage())
+}
 
-		h.ClientWrite(mysqlError.GetMessage())
+func genericProtocolError(goErr error) *protocol.Error {
```
izgeri: this feels like it belongs somewhere in the protocol (protocol/error.go? protocol/protocol.go?) and not here

sgnn7:
Drive-by comment: would be good to clean up the commit history on this before a merge

internal/app/secretless/handlers/mysql/backend.go
jonahx: `Run()` is only concerned with the higher level logic, very much by design.  I'm open to other names or strategies but I don't want to mar `Run()` with an an if statement relevant only to one method.  I considered `OptionallyUpgradeToSSL` or `HandleClientSSLRequests`, etc -- do you like those better?  I do not like `UpgradeToSSLIfRequested`, though...

internal/app/secretless/handlers/mysql/handler.go
jonahx: That's just part of the protocol definition.  I'll add a comment to that effect.

internal/app/secretless/handlers/mysql/handler.go
jonahx: Yes all the details about the packet structure will exist in that code.  The code you're reviewing now now is "client code" -- it will interact only with our custom Go types that represent these things, and won't know anything about the internals of packet structure.  

All this will be documented.

internal/app/secretless/handlers/mysql/handler.go
jonahx: yes this is correct... iirc there may have been some other ambiguity i was addressing, but that should probably be dealt with differently.  I'll revisit and fix.

internal/app/secretless/handlers/mysql/handler.go
jonahx: > why not just have one method `NewConnection` that accepts a sequenceId parameter?

You definitely do not want to do that.   That would make your client code (in the sense of "code that uses this code") aware of protocol impl details.  To the client code there is only a "client" connection and a "backend" connection.  That's how it thinks of it.  

This information may *really* belong in the protocol code I'm working  on -- you're right to notice it's not quite right that *this* protocol detail lives here and all the rest will be tidily hidden away.  I'll have to think about that.  What you see now is a cleanup of the previous code, where this mixing was much worse, so that old design may still be exerting some pressure on my thinking.  Worse case, this will documented much better with comments.  Best case, I'll find a way to bring all the protocol code together neatly.

internal/app/secretless/handlers/mysql/handler.go
jonahx: Yes that was the plan when we do the merge.

jonahx:
> Drive-by comment: would be good to clean up the commit history on this before a merge

Oh yeah, this is full on "private work" mode atm.  Though you'll probably have to settle for a squash :(

jonahx:
@izgeri Great PR so far.  You seem to be noticing all the things I knew were wrong... the best kind of critique.

internal/app/secretless/handlers/mysql/handler.go
izgeri: there's something interesting here about the connection between the `packet` (which has a sequenceId) and the `connection` (which right now also has a sequenceId). I feel this property belongs more with the packet, but it will be interesting if we can find a way to define things so that the implementation remains correct while the objects are more logical

internal/app/secretless/handlers/mysql/backend.go
izgeri: I think I prefer `HandleClientSSLRequests`

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -89,37 +168,69 @@ func stream(source, dest net.Conn, callback func([]byte)) {
...
 		return
 	}
 
-	if err = h.ConnectToBackend(); err != nil {
-		h.abort(err)
+	h.mySQLBackendConn = NewMySQLBackendConnection(rawBackendConn)
+
+	// 3. Runs through the connection phase steps to authenticate.
+	//
+	connPhase := NewConnectionPhase(h.mySQLClientConn, h.mySQLBackendConn, connDetails)
```
izgeri: this naming seems off to me. what it's doing is negotiating the authentication handshake to create a new authenticated connection. wdyt about `ConnectionHandshake` instead of `ConnectionPhase`?

izgeri: or maybe `AuthenticationHandshake`

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -11,34 +70,23 @@ import (
...
-//
-type mysqlPacket []byte
-
-func (pkt *mysqlPacket) SequenceId() byte {
-	return (*pkt)[3]
-}
-
-func (pkt *mysqlPacket) SetSequenceId(id byte) {
-	(*pkt)[3] = id
+var ErrNoTLS = &protocol.Error{
```
izgeri: this should go wherever `genericProtocolError` goes

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -70,273 +118,310 @@ func (h *Handler) ConfigureBackend() (err error) {
 		result.Options[k] = string(v)
 	}
 
-	h.BackendConfig = &result
+	return result, nil
+}
+
+// https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase.html
+//
+type ConnectionPhase struct {
```
izgeri: I'd love to see this moved into the protocol

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -483,7 +484,7 @@ func UnpackHandshakeResponse41(packet []byte) (*HandshakeResponse41, error) {
 		PacketTail:      packetTail}, nil
 }
 
-// InjectCredentials takes in a HandshakeResponse41 from the client, the
+// injectCredentials takes in a HandshakeResponse41 from the client, the
```
izgeri: I think `go lint` enforces the capital here, actually

internal/app/secretless/handlers/ssl/ssl.go
```diff
@@ -9,14 +9,14 @@ import (
 
 type options map[string]string
 
-type SecretlessTLSConfig struct {
+type DbSSLMode struct {
```
izgeri: what's the thought behind this change? it seems like this struct contains more than just the ssl mode, esp if I think of ssl mode as the string parameter from the secretless config that defines how to handle SSL in the secretless -> backend connection. I would have thought `DbTLSConfig` was sufficient

internal/app/secretless/handlers/mysql/handler.go
jonahx: > I feel this property belongs more with the packet

@izgeri I see why'd you'd think so, because the data does live in the packet.  And there _is_ more cleanup to be done here probably (see below).  But I do believe it's a higher level concern than the packet.

The MySQL protocol is a stateful protocol, and as the mysql server and client send packets back and forth to each other they track what part of the process they're in using this sequence id.  That seq id is part of the _header_ (which we don't _yet_ have a proper abstraction for -- part of the other refactor).  But in any case, we want this part of the protocol to be transparent to the code that is writing the packet _payloads_ -- ie, the ConnectionPhase stuff.  I just want to say "write" this payload, and have the header with the proper sequence id inserted.  Which is why it makes sense to live at the connection level.  

Or more simply, the sequence id is, conceptually, a property of _the entire process of sending packets back and forth_ -- not of any particular packet (even though each packet does include that information).  The "Connection" is precisely what represents that entire process of sending back and forth.

internal/app/secretless/handlers/ssl/ssl.go
jonahx: Your instinct here is spot on.  The problem is that it's the wrong abstraction, and it's doing multiple things, so every name will be somewhat wrong.  This will be fixed during the ssl refactoring issue, already in waffle.

For now I do think this is slightly "less wrong" than `DbTLSConfig` because if you look at the logic around this, most of it has to do with ssl mode stuff, even though it also has the certs themselves.  so i'd like to label this `WONT_FIX_UNTIL_SSL_TICKET`.

internal/app/secretless/handlers/mysql/backend.go
jonahx: So the design is: protocol provides the abstractions over the low level protocol details.  This thing (`ConnectionPhase` above but soon to be `AuthenticationHandshake`) uses those abstractions to actually broker the connection.  That is, this code is not _part_ of the protocol, it _uses_ the protocol.  Or again: This code is the heart of the secretless handler.  A secretless handler is  not part of the MySQL protocol.  

So we don't want to move this into protocol.  That said, the `backend` vs `handler` distinction is a bit vague.

internal/app/secretless/handlers/mysql/connection.go
```diff
@@ -0,0 +1,60 @@
...
+// Importantly, putting the sequence id, which is part of the *header*, in
+// Connection allows it to be transparent to the code that is writing the packet
+// *payloads*.
+//
+type Connection struct {
+	conn       net.Conn
+	sequenceId byte
+}
+
+func NewMySQLClientConnection(conn net.Conn) *Connection {
```
izgeri: can we drop the `MySQL` from here and below? just `NewClientConnection`?

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -89,37 +128,69 @@ func stream(source, dest net.Conn, callback func([]byte)) {
...
+func (h *Handler) pipe() {
 	plugin_v1.PipeHandlerWithStream(h, stream, h.EventNotifier, func() {
 		h.ShutdownNotifier(h)
 	})
 }
 
-// Run configures the backend connection info, connects to the backend to
-// complete the connection phase, and pipes the data between the client and
-// the backend
+// This is the main handler method. It:
```
izgeri: I would guess that `go lint` would complain about this comment not starting with the (exportable) method name: `Run`

internal/app/secretless/handlers/mysql/backend.go
```diff
@@ -1,342 +1,387 @@
...
+// between a MySQL client and server during which authentication occurs.
+// Note this is distinct from the various specific handshake packets that are
+// sent back and forth during this process.
+//
+// Note this is process is referred to as the "connection phase" in the MySQL
+// docs:
+//
+// https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase.html
+//
+type AuthenticationHandshake struct {
```
izgeri: is this file really authentication_handshake.go or handshake.go now?

internal/app/secretless/handlers/mysql/authentication_handshake.go
```diff
@@ -0,0 +1,380 @@
+package mysql
+
+import (
+	"github.com/cyberark/secretless-broker/internal/app/secretless/handlers/mysql/protocol"
+	"github.com/cyberark/secretless-broker/internal/app/secretless/handlers/ssl"
+	"net"
+)
+
+var errNoTLS = &protocol.Error{
```
izgeri: wondering if this should be moved to the protocol with `NewGenericError`

internal/app/secretless/handlers/mysql/authentication_handshake.go
```diff
@@ -0,0 +1,380 @@
...
+	h.readServerHandshake()
+	h.writeHandshakeToClient()
+	h.validateServerSSL()
+	h.readClientHandshakeResponse()
+	h.overrideClientCapabilities()
+	h.injectCredentials()
+	h.handleClientSSLRequest()
+	h.writeClientHandshakeResponseToBackend()
+
+	// TODO: Kumbi: what does this mean? AuthSwitchRequest, AuthSwitchResponse
```
izgeri: is this getting left in still? looking at the code without any context, I'm not sure why this is here

internal/app/secretless/handlers/mysql/authentication_handshake.go
```diff
@@ -0,0 +1,380 @@
...
+	// TODO: change this to method call on clientHandshakeResponse when Kumbi's work done
+	h.err = protocol.InjectCredentials(
+		h.clientHandshakeResponse,
+		h.serverHandshake.Salt,
+		h.connectionDetails.Username,
+		h.connectionDetails.Password,
+	)
+}
+
+// TODO: All of this should be moved to kumbi's code
```
izgeri: just to verify, that's within the protocol? because if we're getting into manipulating packets, I think that's where it belongs.

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -2,58 +2,97 @@ package mysql
...
+	// deal with these decorated versions.
+
+	mySQLClientConn *Connection
+	mySQLBackendConn *Connection
 	plugin_v1.BaseHandler
-	BackendConfig *BackendConfig
+	connectionDetails *ConnectionDetails
 }
 
-func (h *Handler) abort(err error) {
```
izgeri: my sense is that we're dropping this because we've implemented even better error handling - is this fair? I know we were aborting the connection attempt in certain circumstances before

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -2,58 +2,97 @@ package mysql
...
+// "Unknown" MySQL protocol error, because the client understands only those.
+//
+func (h *Handler) sendErrorToClient(err error) {
+	mysqlError, isProtocolErr := err.(*protocol.Error)
+
+	if !isProtocolErr {
+		mysqlError = protocol.NewGenericError(err)
+	}
+
+	// TODO: log something if this fails
```
izgeri: if the write fails? why not add that now?

internal/app/secretless/handlers/mysql/authentication_handshake.go
jonahx: addressed

internal/app/secretless/handlers/mysql/authentication_handshake.go
jonahx: removed

internal/app/secretless/handlers/mysql/authentication_handshake.go
jonahx: Removed, although this code will likely change when we do the ssl utility code changes

internal/app/secretless/handlers/mysql/handler.go
jonahx: yes this has essentially been moved around
the logic still exists to abort

PR #634, opened 2019-02-11
shared library exposing secret providers

doodlesbykumbi:
Connected to #635

An experiment in exposing the internals of Secretless as a C shared library. Static would be best but for now dynamic linking is all we've got.

Used in this PR https://github.com/doodlesbykumbi/envoy/pull/1 to allow secret injection via Envoy


cmd/shared/README.md
```diff
@@ -0,0 +1,54 @@
+# C shared library
+
+This is an experiment in exposing the internals of Secretless as a C shared library.
+Static would be best but for now this needs dynamic linking since some Secretless components use CGO.
+
```
izgeri: It should include something about its state here - I'd call it `pre-alpha` bc it has no tests, wdyt?

cmd/shared/main.go
```diff
@@ -0,0 +1,77 @@
...
+	secrets, err := GetSecrets([]config.StoredSecret{ref})
+	if err != nil {
+		fmt.Println("Error fetching secret")
+		return C.CString("")
+	}
+	return C.CString(string(secrets[ref.Name]))
+}
+
+//export NativePassword
+func NativePassword(cRef C.struct_StoredSecret, salt *C.char) (*C.char) {
```
izgeri: Is this MySQL-specific?

izgeri: can we just use this method here? https://github.com/cyberark/secretless-broker/blob/master/internal/app/secretless/handlers/mysql/protocol/protocol.go#L713

cmd/shared/README.md
doodlesbykumbi: Sounds good

cmd/shared/main.go
doodlesbykumbi: same functionality but different signatures. i'm trying to avoid the Go strings because I can't control when their memory is freed.

izgeri: so, if we change the input for the existing one to `[]byte` instead of string, you would use that? bc I am 100% for that, rather than having the functionality defined in 2 places

izgeri: the method converts it to a byte array very early on anyway

doodlesbykumbi: fair point. done

internal/app/secretless/handlers/mysql/protocol/protocol.go
```diff
@@ -711,9 +711,11 @@ func CheckPacketLength(expected int, packet []byte) error {
 // https://dev.mysql.com/doc/internals/en/secure-password-authentication.html#packet-Authentication::Native41
 // SHA1( password ) XOR SHA1( "20-bytes random data from server" <concat> SHA1( SHA1( password ) ) )
 func NativePassword(password string, salt []byte) (nativePassword []byte, err error) {
-
+	return NativePasswordWithBytes([]byte(password), salt)
```
izgeri: Rather than doing this, can we just update where it's called to pass in a byte array? It's just [one place](https://github.com/cyberark/secretless-broker/blob/master/internal/app/secretless/handlers/mysql/protocol/protocol.go#L513) in the code and [one test](https://github.com/cyberark/secretless-broker/blob/master/internal/app/secretless/handlers/mysql/protocol/protocol_test.go#L429) that would need to be updated. I don't see the value in keeping the original method with the string input if we are trying to get away from that - in fact, I see the very clear drawback that if we leave it here, someone may use it as we're trying to get away from passing strings around.

PR #638, opened 2019-02-14
Fixed minor typo in security.md

sgnn7:
Seems like there was a spelling problem :(

### DIFF
--- a/docs/docs/overview/security.md
+++ b/docs/docs/overview/security.md
@@ -47,7 +47,7 @@ It is _very_ important to configure the OS properly so that unauthorized process
 
 ## Plugin checksum verification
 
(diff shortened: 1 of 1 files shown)


sigalsax:
I want to double check, this file and its contents will be deleted because we are moving docs off the secretless site. This typo has led to a merge conflict in my branch

izgeri:
@sigalsax Srdjan updated the security page here and on the docs site. Since this is still publishing, it makes sense to continue editing here. Hopefully the rebase is easy enough to do, and it's ok to wait until just before go-live to do it again.

sigalsax:
found my way through the tunnel of rebasing and changes (: we good!

PR #660, opened 2019-03-15
Fix all non-TODO CodeClimate issues

jonahx:

This fixes a whole bunch of codeclimate shellcheck errors

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -22,7 +22,8 @@ fi
 
 echo "Building secretless-broker:$TAG Docker image"
 # NOTE: the latest tag is required by downstream pipeline stages
-# shellcheck disable=SC2086 # (we want the flags to be word split here)
+# (we want the flags to be word split here)
+# shellcheck disable=SC2086
 docker build -t "secretless-broker:${TAG}" \
              -t "secretless-broker:latest" \
              $DOCKER_FLAGS \
@@ -31,7 +32,8 @@ docker build -t "secretless-broker:${TAG}" \
 
 echo "Building secretless-dev:$TAG Docker image"
 # NOTE: the latest tag is required by downstream pipeline stages
-# shellcheck disable=SC2086 # (we want the flags to be word split here)
+# (we want the flags to be word split here)
+# shellcheck disable=SC2086
 docker build -t "secretless-dev:${TAG}" \
              -t "secretless-dev:latest" \
              $DOCKER_FLAGS \
@@ -40,7 +42,8 @@ docker build -t "secretless-dev:${TAG}" \
 
 echo "Building secretless-broker-quickstart:$TAG Docker image"
 # NOTE: the latest tag is required by downstream pipeline stages
-# shellcheck disable=SC2086 # (we want the flags to be word split here)
+# (we want the flags to be word split here)
+# shellcheck disable=SC2086
 docker build -t "secretless-broker-quickstart:${TAG}" \
              -t "secretless-broker-quickstart:latest" \
              $DOCKER_FLAGS \
--- a/bin/build
+++ b/bin/build
@@ -22,24 +22,27 @@ fi
 
 echo "Building secretless-broker:$TAG Docker image"
 # NOTE: the latest tag is required by downstream pipeline stages
+# shellcheck disable=SC2086 # (we want the flags to be word split here)
 docker build -t "secretless-broker:${TAG}" \
              -t "secretless-broker:latest" \
-             "$DOCKER_FLAGS" \
+             $DOCKER_FLAGS \
              -f "$TOPLEVEL_DIR/Dockerfile" \
              "$TOPLEVEL_DIR"
 
 echo "Building secretless-dev:$TAG Docker image"
 # NOTE: the latest tag is required by downstream pipeline stages
+# shellcheck disable=SC2086 # (we want the flags to be word split here)
 docker build -t "secretless-dev:${TAG}" \
              -t "secretless-dev:latest" \
-             "$DOCKER_FLAGS" \
+             $DOCKER_FLAGS \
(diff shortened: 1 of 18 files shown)


bin/build
```diff
@@ -6,7 +6,7 @@ set -ex
 
 . bin/build_utils
 
-CURRENT_DIR=$($(dirname $0)/abspath)
+CURRENT_DIR=$("$(dirname "$0")/abspath")
```
izgeri: thoughts about setting `prefix=$(dirname "$0")` before this line? quotes inside quotes look weird to me

izgeri: wdyt about setting `prefix=$(dirname "$0")` before this line? having a command inside a command makes it harder to read imo

bin/build_website
```diff
@@ -22,11 +23,12 @@ rm -rf "${DEST_SITE_DIR}"
...
 
 echo "Cleaning up generated files..."
-rm -rf $DOCS_DIR/generated/*.html
+rm -rf "$DOCS_DIR"/generated/*.html
 
 echo "Reading and generating godoc pages..."
-cat $DOCS_DIR/godoc_packages.txt | while read package; do
-  if [[ "${package}" =~ ^\#.* ]] || [[ "${package}" == "" ]]; then
+while read -r package "$DOCS_DIR/godoc_packages.txt"; do
+  # Note: "${package:?}" instead of "${package}" to avoice shellcheck 2154
```
izgeri: avoice?

bin/cpu_profiling
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash -e
 
-for i in {1..10}; do
+#Note: use _ instead of i to avoid shellcheck error
```
izgeri: is this because we don't use `i`? I think it's fine to make this change without the note

bin/build
jonahx: all this stuff will be changed when we do:

> Add ticket to cleanup the way we’re determining path per the discussion in about bash today.

ie, the way we're doing abspath is wrong anyway (we shouldn't be using `$0`, eg), so i'm not sure this worth is fixing since all this code will go away when that's fixed.  i will if you feel strongly.

bin/build_website
jonahx: that's how the cool kids spell "avoid" ldo...

bin/build_website
```diff
@@ -22,11 +23,13 @@ rm -rf "${DEST_SITE_DIR}"
 mkdir -p "${DEST_SITE_DIR}"
 
 echo "Cleaning up generated files..."
-rm -rf $DOCS_DIR/generated/*.html
+rm -rf "$DOCS_DIR"/generated/*.html
 
 echo "Reading and generating godoc pages..."
-cat $DOCS_DIR/godoc_packages.txt | while read package; do
-  if [[ "${package}" =~ ^\#.* ]] || [[ "${package}" == "" ]]; then
+while read -r package "$DOCS_DIR/godoc_packages.txt"; do
```
izgeri: it seems to be hanging here when I run this script

bin/prefill_changelog
```diff
@@ -11,47 +17,49 @@ if [[ "$1" == "--dry-run" ]]; then
...
 change_date=$(date +%Y-%m-%d)
 
-changelog_snippet="\n"
-changelog_snippet+="## [${desired_version}] ${change_date}\n\n"
-changelog_snippet+="### TODO: Don't commit this without proofreading!!!\n\n"
-changelog_snippet+="### Added\n\n"
-changelog_snippet+="### Changed\n\n"
+changelog_snippet='\n'
+changelog_snippet+="## [${desired_tag}] ${change_date}"'\n\n'
+changelog_snippet+="### TODO: Don't commit this without proofreading!!!"'\n\n'
```
izgeri: I think the single quotes are causing this text to be parsed incorrectly - running the script results in output that looks like this:
```
\n## [0.6.5] 2019-03-18\n\n### TODO: Don't commit this without proofreading!!!\n\n### Added\n\n### Changed\n\n- Address a couple of Geri's comments
```

bin/codegen
```diff
@@ -1,10 +1,9 @@
 #!/bin/bash -e
 
-CURRENT_DIR=$($(dirname $0)/abspath)
```
izgeri: this script doesn't work, but I don't think it's your changes ...

bin/publish_website
```diff
@@ -18,6 +20,6 @@ docker run --rm \
   -e AWS_ACCESS_KEY_ID \
   -e AWS_SECRET_ACCESS_KEY \
   mesosphere/aws-cli \
-    s3 sync --delete /_site s3://${S3_BUCKET}
+    s3 sync --delete "/_site s3://${S3_BUCKET}"
```
izgeri: I don't know how to test this - but I worry about this failing when it gets merged to master.

jonahx: i am like 99% confident that change won’t break anything, so if it was working before it will still be working.  the previous would fail if the S3_BUCKET name has spaces in it, which is probably not allowed anyway, but these rules exists so you don’t you think about that stuff.

bin/codegen
jonahx: new issue for this then?

jonahx: https://github.com/cyberark/secretless-broker/issues/663

bin/prefill_changelog
jonahx: I rewrote this and it worked for me, you should re-run the test yourself though to be 100% sure.

What happened was I forgot that in my refactor `changlog_snippet` was supposed to be the printf format string, not the data string.  I updated it to reflect that.

bin/build_website
jonahx: Fixed.

Issue #666, opened 2019-03-25
Tutorial boilerplate code framework exists in secretless


sigalsax:
The secretless site needs a way to present information in its tutorials in a more focused way that is not overwhelming for the user.

The design is a carousel where as users advance through the tutorial, they will only be shown a certain amount of information at a time (ex: 100 lines worth of info). When user presses next, the current section of the page will disappear and another section will appear its place. That way, the user will stay on the same page but toggle through information as they see fit.

PR #674, opened 2019-03-28
Add tutorial framework and design secretless

sigalsax:
This PR implemented the latest k8s tutorial design for the Secretless site 

Connected to #666 

link to start: http://localhost:4000/docs/tutorials/kubernetes/kubernetes-tutorial-base.html

Design to be implemented:
computer screens:
Screen Shot 2019-03-28 at 9 52 51

mobile:
mobile


docs/_data/navigation.yml
```diff
@@ -4,7 +4,7 @@ main:
     target: target="_blank" rel="noopener noreferrer"
     section: /docs/
   - title: Tutorial
-    url: /docs/get_started/kubernetes_tutorial.html
+    url: /docs/tutorial-tiles/kubernetes/kubernetes-tutorial-base.html
```
izgeri: - [ ] can we just call this directory `tutorials/`?
- [ ] can docs/get_started/kubernetes_tutorial.html redirect to the new tutorial start page?
- [ ] can you log an issue in the docs repo to update the tutorial links there and note `Dependent on cyberark/secretless-broker#666`

docs/_data/navigation.yml
```diff
@@ -38,3 +38,13 @@ footer-resources:
   - title: FAQ
     url: https://docs.secretless.io/Latest/en/Content/FAQ/faq.htm
     target: target="_blank" rel="noopener noreferrer"
+
+tutorial-tiles:
```
izgeri: shouldn't this just be referencing the k8s tutorial? what if we have more tutorials down the line?

docs/_layouts/tutorials.html
```diff
@@ -0,0 +1,70 @@
...
+            <p class="page-subtitle-right">Products used: {{ page.products-used }}</p>
+          </div>
+
+          <div class="row page-wrapper">
+            <!-- left-side of tutorial with content -->
+            <div class="col-xs-12 col-s-12 col-md-3 tutorial-left">
+              <div class="progressbar-index">
+                <ul class="progressbar">
+
+                  {% for item in site.data.navigation.tutorial-tiles %}
```
izgeri: should `tutorial-tiles` here be a parameter (not sure if that's possible, but I'm shooting for the moon!) so that multiple tutorials could use this layout, and this could reference the layout for any given tutorial?

docs/_sass/_tutorial.scss
```diff
@@ -0,0 +1,242 @@
+// ***********************
+// Tutorial styling
+// ***********************
+.tutorial-right {
+	background-color: #F7F8FA;
```
izgeri: can we use consistent indentations for this file?

docs/javascript/tutorial.js
```diff
@@ -0,0 +1,28 @@
+var nodeList = $('.progressbar-bubble');
+
+var forwardBtn = document.getElementById('continue');
```
izgeri: why not use jquery here too?

docs/javascript/tutorial.js
```diff
@@ -0,0 +1,28 @@
...
+    nodeList[i].className += 'completed';
+  }
+}
+
+// Fill progress bar when at final, full state
+function fullState() {
+  for (let j = 0; j < nodeList.length; j++) {
+    nodeList[j].className += 'completed';
+  }
+}
```
jonahx: Per our slack convo, this really belongs as logic in jekyll, since the page reloads and the state of the nav can be determined statically.

I haven't used Jekyll a ton, but just glancing at the docs one way you might accomplish this by making the nav an "include".  That is, pulling the nav  stuff out of `ocs/_layouts/tutorials.html`, and [making it an include that takes a parameter](https://jekyllrb.com/docs/includes/#passing-parameters-to-includes).  The parameter would be the "active" page.

jonahx: ACTUALLY…

I think there’s a better pure CSS way…

Check out https://css-tricks.com/child-and-sibling-selectors/#article-header-id-2

and the related selectors.  I think you can style “all siblings that come before”.

If so, you can do this with 1 CSS statement, delete your js file, _and_ not have to mess with jekyll

docs/_data/navigation.yml
sigalsax: Updated and created issue but regarding pt 3, should we update links in docs if we have a redirect in place?

docs/_layouts/tutorials.html
sigalsax: I tried a couple of ways to accomplish this (collections, passing parameters with assign from specific tile and assign content in tutorial layout) but it seems that somehow it is being converted into a string (I think).
@jonahx @typaulhus do you know another way to pass a parameter into a for loop with jekyll?

docs/_data/navigation.yml
sigalsax: We could create 'Collections' but that would just be the same thing. For more tutorials, we would add another set of navigation like this one. @typaulhus do you know a work around?

docs/javascript/tutorial.js
sigalsax: I have trying to play with it but ~ is only for siblings after the element. 
I found this: https://stackoverflow.com/questions/1817792/is-there-a-previous-sibling-css-selector, the one with 190 votes but unsure really how they did that...

sigalsax: implementation is a bit longer than expected so I logged an issue for us to come back to: #679

docs/_layouts/tutorials.html
sigalsax: I really like this idea! 
I realize that this is taking longer than expected for implementation so I have logged as issue #680  and when we start to think about adding more tutorials, we can revisit this one

docs/javascript/tutorial.js
sigalsax: I also liked the idea of progress bar being sticky. I logged an issue for this one #681

sigalsax:
I would also like a review on the content I added: 'Up next' and 'You're done!' summary. Might need a better way to end it that inspires action @jonahx @izgeri @typaulhus  (:"

docs/javascript/tutorial.js
jonahx: @sigalsax  

> Do not try to find a backward selector...
> That's impossible.
> Instead, only try to realize the truth...
> There is no direction.
> Then you'll see that it is not the CSS that bends
> It is only yourself.

-- [Spoon Boy](https://www.youtube.com/watch?v=uAXtO5dMqEI)

https://jsfiddle.net/cd6w4yfp/

docs/_data/navigation.yml
sigalsax: addressed in issue created #680

sigalsax: changed to `tutorial-k8s:`

docs/javascript/tutorial.js
sigalsax: bingo bango...

sigalsax: that was embarassingly easy... I was thinking about it all wrong. 
ME: "How do I get it to reverse itself..."
Just put normal styling and cover it with `.active ~ .progressbar-bubble`. DUUUUH ^.^

sigalsax: #pushed

sigalsax:
@izgeri  would you be able to manually rerun the Gitlab tests? I don't think these tests are rerun automatically with force pushes

demos/k8s-demo/README.md
```diff
@@ -6,15 +6,15 @@ It's intended for advanced users and is light on explanations.
 
 For a friendlier version of this tutorial, with breakdowns of every step, try:
 
-[Our Detailed Introduction to Secretless in Kubernetes](https://secretless.io/docs/get_started/kubernetes_tutorial.html)
+[Our Detailed Introduction to Secretless in Kubernetes](https://secretless.io/docs/tutorials/kubernetes/kubernetes-tutorial-base.html)
```
izgeri: can we get this out of the `docs` folder? it seems odd now to have that on the path

izgeri: this link is still outdated

docs/docs/get_started/kubernetes_tutorial.md
```diff
@@ -4,15 +4,17 @@ id: kubernetes_tutorial
 layout: subpages
 description: Secretless Broker Documentation
 permalink: /docs/get_started/kubernetes_tutorial.html
+redirect_to: /tutorials/kubernetes/kubernetes-tutorial-base.html
+
 ---
 
-This is a detailed, step-by-step tutorial.
```
izgeri: do we still need the content here if we're redirecting? probably best to delete it, so that people don't get confused if they have to edit this tutorial

docs/tutorials/kubernetes/app-dev.md
```diff
@@ -0,0 +1,217 @@
...
+
+Again, the application has no knowledge of the database credentials it's using.
+
+For usage examples, please see [Test the Application](#test-the-application).
+
+### Create Application Deployment Manifest
+
+We're ready to deploy our application.
+
+A detailed explanation of the manifest below is featured in the next step, <a href="/docs/get_started/kubernetes_tutorial.html">Appendix - Secretless
```
izgeri: this isn't the right link anymore

docs/docs/get_started/kubernetes_tutorial.md
sigalsax: good catch

sigalsax:
This PR seems to have gotten a bit too messy to rebase and reorganize commits. Created new branch #683

sigalsax:
closing b/c merged here: https://github.com/cyberark/secretless-broker/pull/684

PR #678, opened 2019-04-02
Update contributor info for documentation

izgeri:
- Update CONTRIBUTING.md with info on documentation sources / doc dev
- Move docs/README content into CONTRIBUTING
- More clearly describe which docs go where


CONTRIBUTING.md
```diff
@@ -134,6 +135,46 @@ cd test/manual/k8s_crds
...
+
+### Website
+The [website](https://secretless.io) source is in the [docs](docs/) folder in this repository. It is generated using Jekyll.
+
+The source includes:
+- the website main page
+- some old pages that redirect to the documentation subdomain
+- tutorials
+- godocs for the plugin API
+
```
izgeri:
```suggestion
- Secretless blog
- community info page
```

CONTRIBUTING.md
```diff
@@ -134,6 +135,47 @@ cd test/manual/k8s_crds
 ```
 This test currently does not run as part of the test suite.
 
+## Documentation
+Secretless has a few sources for documentation: a website, a documentation subdomain, and godocs.
```
jonahx: I would be confused here what the difference is between "the website" and "the documentation subdomain".  I know this is explained below but it would be nice if the outlines of the differences could be conveyed here in a phrase.

Eg, "the website and the documentation subdomain are split for no good reason.  They are an aritfact of Conway's law"

:)

PR #685, opened 2019-04-12
automated testing of CRDs

doodlesbykumbi:
Automates testing of CRDs
1. Uses an image built from google/cloud-sdk, with Docker CLI and kubectl
2. Docker in Docker via socket sharing
3. Uses summon to fetch credentials for connecting to GKE
4. Pushes local Secretless image to GKE registry
5. Runs tests against GKE cluster
6. modify CRD to allow concurrent tests runs through arbitrary naming of CRD `CRDGroupName = "secretless" + os.Getenv("SECRETLESS_CRD_SUFFIX") + ".io"`


k8s-ci/utils.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Sets additional required environment variables that aren't available in the
+# secrets.yml file, and performs other preparatory steps
+function prepareTestEnvironment() {
+  # Prepare Docker images
+  docker build --tag "gke-utils:latest" \
```
hughsaunders: I think `UNIQUE_TEST_ID` or similar would be useful here to prevent collisions between builds.

k8s-ci/test
```diff
@@ -0,0 +1,41 @@
...
+# Set environment variables
+export UNIQUE_TEST_ID="$(uuidgen | tr "[:upper:]" "[:lower:]" | head -c 10)"
+export SECRETLESS_IMAGE="${DOCKER_REGISTRY_PATH}/secretless-broker:${UNIQUE_TEST_ID}"
+
+# Clean up when script completes
+function finish {
+  announce 'Wrapping up and removing test environment'
+
+  # Delete registry images that were used
+  deleteRegistryImage "${SECRETLESS_IMAGE}" > /dev/null
```
hughsaunders: also the gke-utils image?

k8s-ci/test
```diff
@@ -0,0 +1,41 @@
...
+trap finish EXIT
+
+function testCRDs() {
+  runDockerCommand "
+# push the local secretless-broker image to the remote DOCKER_REGISTRY_PATH
+docker tag 'secretless-broker:latest' '${SECRETLESS_IMAGE}' > /dev/null;
+docker push '${SECRETLESS_IMAGE}' > /dev/null;
+
+# run CRD tests
+cd test/manual/k8s_crds;
```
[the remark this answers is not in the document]
izgeri: good catch!

k8s-ci/utils.sh
doodlesbykumbi: `gke-utils:latest` should actually be a globally accessible image. it generally shouldn't change. i'll get rid o the docker build arg since it's not necessary

k8s-ci/test
doodlesbykumbi: nah the gke-utils image is for the CI host. if having it around can speed up subsequent builds that can only be a good thing

k8s-ci/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM google/cloud-sdk
```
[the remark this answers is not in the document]
doodlesbykumbi: done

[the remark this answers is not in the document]
doodlesbykumbi: actually not done. this is because `google/cloud-sdk` doesn't have a tag that fixes jessie repo issue aside from latest

k8s-ci/test
```diff
@@ -0,0 +1,56 @@
...
+}
+trap finish EXIT
+
+function testCRDs() {
+  runDockerCommand "
+# push the local secretless-broker image to the remote DOCKER_REGISTRY_PATH
+docker tag 'secretless-broker:latest' '${SECRETLESS_IMAGE}' > /dev/null;
+docker push '${SECRETLESS_IMAGE}' > /dev/null;
+
+export SECRETLESS_CRD_SUFFIX='${UNIQUE_TEST_ID}';
```
hughsaunders: ++ A great improvement for test isolation

k8s-ci/test
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -euo pipefail
+cd "$(dirname "$0")"
+. ./utils.sh
+
+# Set environment variables
+export UNIQUE_TEST_ID
+UNIQUE_TEST_ID="$(uuidgen | tr "[:upper:]" "[:lower:]" | head -c 12 | tr -d -)"
+export SECRETLESS_IMAGE="${DOCKER_REGISTRY_PATH}/secretless-broker:${UNIQUE_TEST_ID}"
```
hughsaunders: Would it be useful to build and tag the secretless-broker image here? I had to manually build it to get the tests to pass.

doodlesbykumbi: This is normally run preceded by a build stage in Jenkins so it assumes the image has been built already. I'm wondering if there's a better way to make that relationship explicit. Perhaps a check before the script runs that complains if there's no such image

hughsaunders: Ah of course, it runs in an earlier pipeline stage!

PR #699, opened 2019-05-03
fix:ssh: fail without exiting program when server config can not be resolved

doodlesbykumbi:
Prevents the Secretless process being terminated when server configuration can not be resolved by SSH handler

Some context:
> Initially, I thought about that for a bit and concluded that it’s just a bug, precisely because I assumed that handlers should never panic. My assumption was grounded in the fact that handlers don’t explicitly panic anywhere nor do they use fatalf. This was the exception.
>
> With that reasoning there wouldn’t be a next time as long as we document clearly that handlers should never panic and that perhaps listeners is when panics might happen because of failed config validation.
>
> But now that I’m here thinking about the UX I wonder if  we might want the whole process to terminate if any given handler is unable to do it’s job (e.g. secret resolution fails). At the moment all handlers log the issue and some terminate the connection to the client (e.g. there’s a presumably unaddressed TODO for the ssh handler to terminate the client connection on errors). An argument can be made for not panicking as much as panicking.
>
> It’d be good to have a consistent error handling stategy

### DIFF
--- a/internal/app/secretless/handlers/ssh/handler.go
+++ b/internal/app/secretless/handlers/ssh/handler.go
@@ -94,7 +94,8 @@ func (h *Handler) Run() {
 	var server ssh.Conn
 
 	if serverConfig, err = h.serverConfig(); err != nil {
-		log.Fatalf("ERROR: Could not resolve server config\n", err)
+		log.Printf("ERROR: Could not resolve server config\n", err)
+		return
 	}
 
 	// TODO: Ensure that we don't print credentials here before uncommenting


jonahx:
Why was it `Fatalf` previously?

doodlesbykumbi:
@jonahx I'm not sure but it shouldn't be. `Fatalf` causes the whole process to panic.

izgeri:
@doodlesbykumbi should we proceed with this PR? I feel like if the issue would be helped by a better error handling strategy, then maybe that's the right approach to ultimately take, rather than making this small one-off change whose impact is hard to understand.

doodlesbykumbi:
@izgeri I'm going to close this for now. We need to think about a better and more consistent error handling approach. I also have too many dangling PRs

Issue #644, opened 2019-02-26
Bump Go version to 1.12


izgeri:
AC:
- [x] Bump Go version to 1.12
- [x] Convert from using `mitchellh/homedir` to `os.UserHomeDir`

PR #709, opened 2019-05-22
Upgrade to Golang 1.12.5

sgnn7:

Upgrades the broker to Golang 1.12.5

Connected to https://github.com/cyberark/secretless-broker/issues/644

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/644-golang-1.12/)

- [x] Manually tested [Keychain provider](https://github.com/cyberark/secretless-broker/tree/master/test/manual/keychain_provider)
- [x] Manually run the [K8s demo](https://github.com/cyberark/secretless-broker/tree/master/demos/k8s-demo)
- [x] Manually run [Kubernetes-Conjur demo](https://github.com/conjurdemos/kubernetes-conjur-demo) with a local Secretless Broker image build of your branch

Connected to https://github.com/cyberark/secretless-broker/issues/644


internal/app/summon/command/temp_factory_test.go
```diff
@@ -0,0 +1,99 @@
+package command
+
+import (
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
```
jonahx: I thought we made our decision yesterday about this:

> We use std lib `testing` for everything, with the addition of assertion func’s provided by `testify`, and mock’s from `testify` when needed

If that's the decision, we should not go any deeper down the Go convey hole....

internal/app/summon/command/temp_factory.go
```diff
@@ -4,12 +4,10 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
-
-	"github.com/mitchellh/go-homedir"
 )
 
 // DEVSHM is the location of a memory-mapped directory on Linux.
-const DEVSHM = "/dev/shm"
+var DEVSHM = "/dev/shm"
```
jonahx: Let's give this a meaningful name.  `DEVSHM` is totally confusing when you just see it -- I had no idea what was going on in the tests until I came back and found this comment. 

Side Note: Go conventions dictate [CamelCase for constants](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps) -- the stdlib follows this as well.

I know this is a `var` and not a `const`, but that appears to have been done only so that it could be changed in the unit tests.  But it actually *is* a constant afaict, and so this seems to be a hack?  Let me know if I've misunderstood this one.

This one already existed I know, but since we're changing this code we should rename `TempFactory` to `TempFileFactory` since that's what it really is. 

Also, importantly, I'd like brief comments explaining the reasons this type (`TempFileFactory`) even exists at the top of the file.  Obviously it creates temp files, but why did we have to create it?  And also why we need to use the memory-mapped directory (is it just for speed?  are there other considerations?).  

More and more, just generally speaking now,  I feel the "why" and the context are extremely important to capture in the code, and having them will significantly improve the experience of future devs.

internal/app/summon/command/temp_factory_test.go
```diff
@@ -0,0 +1,99 @@
...
+		})
+
+		Convey("tries using os.TempDir as last resort", func() {
+			// Override DEVSHM
+			devshm := DEVSHM
+			DEVSHM = ""
+			defer func() {
+				DEVSHM = devshm
+			}()
+
```
jonahx: Per my other comment, making a const into a var so we can fiddle with it in tests feels hacky.  What this is telling us that `os.Stat` is a dependency.  So you can inject it (variadic args allow you to do this without messing up the current ctor interface).  Then in your tests you can inject a double that will return a mock `fi` object, allowing you to do all this cleanly.

demos/k8s-demo/01_security_admin_steps
```diff
@@ -1,22 +1,24 @@
...
-. ./security_admin_secrets.sh
+. ./env.sh
+
 
 ##################################################
 step "Create a new namespace"
 
-delete_ns_and_cleanup "quick-start-backend-ns"
-kubectl create namespace "quick-start-backend-ns"
+kubectl create namespace "${BACKEND_NAMESPACE}"
```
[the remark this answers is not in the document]
sgnn7: Yes. `./stop` step is now included at line 3.

demos/k8s-demo/02_app_developer_steps
```diff
@@ -1,32 +1,34 @@
-#!/usr/bin/env bash
+#!/bin/bash -e
```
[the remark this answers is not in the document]
sgnn7: I have not seen a single *nix OS where bash was not in `/usr/bin` plus the `/usr/bin/env` is a hardcoded path anyways so we're depending on a hardcoded path one way or another.

demos/k8s-demo/02_app_developer_steps
```diff
@@ -1,32 +1,34 @@
...
+
 # Since wait_for_app and kubectl rollout status are both not working...
 # We revert to simply pinging the endpoint
 printf 'Waiting for application to boot up\n'
 printf '(This may take more than 1 minute)\n'
 
 until curl --connect-timeout 3 --silent "${APPLICATION_URL}/pets" >/dev/null; do
-  printf "."
-  sleep 3
+  echo -n "."
```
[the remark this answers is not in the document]
sgnn7: no complaints out of codeclimate and `echo`s are much more common then `printf` :shrug:

demos/k8s-demo/02_app_developer_steps
```diff
@@ -1,32 +1,34 @@
...
+  sleep 2
 done
 
-printf 'Done\n'
+echo "OK"
```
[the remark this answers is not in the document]
sgnn7: Either is fine but I've seen "OK" as a more-common phrase. We are "done with the waiting task" but you cal also consider it as "waiting task went OK". I can change if needed.

internal/app/summon/command/temp_factory_test.go
sgnn7: I implemented this before that decision :/

demos/k8s-demo/02_app_developer_steps
```diff
@@ -42,4 +44,9 @@ step "Retrieve All Pets"
 
 curl -include "${APPLICATION_URL}/pets"
 
-printf '\n\nAll finished!  Secretless is working!\n\n'
+echo
+echo
+step "*** All finished! Secretless is working! ***"
```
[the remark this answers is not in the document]
sgnn7: k - will change

demos/k8s-demo/02_app_developer_steps
```diff
@@ -42,4 +44,9 @@ step "Retrieve All Pets"
...
+
+step "Cleaning up"
+./stop || true
```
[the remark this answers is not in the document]
sgnn7: Sure but more common scenario is that they just want to test things. I'd rather we clean up than leave stuff hanging in their env but that's a personal leaning. Dunno - your call on that.

demos/k8s-demo/utils.sh
```diff
@@ -18,39 +22,14 @@ wait_for_app() {
     --output jsonpath='{$.items[0].status.containerStatuses.*.ready}')" =~ (true ?)+ ]]
   do
     if [[ "$waiting" != "true" ]]; then
-      printf "Waiting for %s to be ready" "$1"
+      echo "Waiting for $1 to be ready"
```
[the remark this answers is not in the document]
sgnn7: yes, this passes codeclimate shellcheck

demos/k8s-demo/utils.sh
```diff
@@ -18,39 +22,14 @@ wait_for_app() {
...
       waiting=true
     fi
-    printf "."
+    echo -n "."
```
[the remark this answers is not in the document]
sgnn7: ditto - passes shellcheck on codeclimate

demos/k8s-demo/02_app_developer_steps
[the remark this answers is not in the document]
jonahx: eh, i guess i don't care about this one too much

internal/app/summon/command/temp_factory.go
```diff
@@ -21,25 +19,43 @@ type TempFactory struct {
...
+// sharedMemoryDir. If sharedMemoryDir is empty, we use the default path for it.
+// defer Cleanup() if you want the files removed.
+func NewCustomTempFactory(path string, sharedMemoryDir string) TempFactory {
 	if path == "" {
-		path = DefaultTempPath()
+		path = defaultTempPath(sharedMemoryDir)
+	}
+
+	return TempFactory{
+		path: path,
```
jonahx: 👍

internal/app/summon/command/temp_factory.go
```diff
@@ -21,25 +19,43 @@ type TempFactory struct {
 // NewTempFactory creates a new temporary file factory.
 // defer Cleanup() if you want the files removed.
 func NewTempFactory(path string) TempFactory {
+	return NewCustomTempFactory(path, "")
```
jonahx: Can't remember if I mentioned this last time, but per the comment above, this should be `NewTempFileFactory` and the filename should be `temp_file_factory.go` and the change should be made whereever else it's relevant, since a `TempFactory` is too ambiguous.

internal/app/summon/command/temp_factory.go
```diff
@@ -21,25 +19,43 @@ type TempFactory struct {
...
+	home, err := os.UserHomeDir()
 	if err == nil {
-		dir, _ := ioutil.TempDir(home, ".tmp")
-		return dir
+		dir, err := ioutil.TempDir(home, ".tmp")
+		if err == nil {
+			return dir
+		}
 	}
+
```
jonahx: 👍

internal/app/summon/command/temp_factory_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (e *envSnapshot) RestoreEnv() {
+	ClearEnv()
+	for _, s := range e.env {
+		k, v := splitEq(s)
+		os.Setenv(k, v)
+	}
+}
+
```
jonahx: Just a thought, not sure if possible, but seems like it would be a lot cleaner to just execute the tests in a subshell rather snapshotting and restoring the env.  Could that be done somehow?

internal/app/summon/command/temp_factory_test.go
```diff
@@ -0,0 +1,88 @@
...
+	Convey("When constructor path is not provided", t, func() {
+		env := ClearEnv()
+		defer env.RestoreEnv()
+
+		Convey("tries using shared memory path first", func() {
+			tempFactory := NewTempFactory("")
+
+			So(tempFactory, ShouldResemble, TempFactory{
+				files: []string(nil),
+				path:  "/dev/shm",
```
jonahx: This goes against my comment above to make `sharedMemoryDir` private, but if we are going to write this test, then the `path` value should def not be hardcoded.  Instead it shoud be `path: TempFactory.SharedMemoryDir`.  

*However*, I'm wondering what we're really testing here?  This feels like a test of impl detail.  Got me thinking about what the purpose of this type really is.  Then I looked at the  definition and noticed the summon ext, and realized that it's not really clear.  So I'd like to know what you think about that.

internal/app/summon/command/temp_factory_test.go
jonahx: You can resolve this.  New tests look good 👍

internal/app/summon/command/temp_factory.go
sgnn7: We discussed this to be pushed to another story: 
https://github.com/cyberark/secretless-broker/issues/721
https://conjurhq.slack.com/archives/CBTETAE20/p1559053808056500?thread_ts=1559052741.053300&cid=CBTETAE20

demos/k8s-demo/README.md
```diff
@@ -88,12 +104,12 @@ configmap "quick-start-application-secretless-config" created
 to the database.**
 
 In particular, as the application developer you do not know any of the secrets
-in **security_admin_secrets.sh**.
+in **security_admin_vars.sh**.
```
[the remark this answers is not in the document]
sgnn7: I don't quite think I agree with either this comment nor the one about `env.sh` since "vars" and "env" are somewhat universal but at this point we're needlessly bikeshedding this PR so I'll make the change.

internal/app/summon/command/temp_factory_test.go
sgnn7: Either way, you have to somewhere test the exact value of the variable. Either you:
- Expose `TempFactory.SharedMemoryDir`
- Test that `TempFactory.path` == `TempFactory.SharedMemoryDir`
- Test that `TempFactory.SharedMemoryDir` == `/dev/shm`
or you 
- Don't care about where the path value comes from as long as it's the expected value

In both cases, you have to assert in the tests that the returned shared memory dir matches the one you expect and that can easily be done by string cmp since the value is a static const.

Since I realize now that we don't need the `TempFactory.SharedMemoryDir` public as nothing is using that outside of the class, I'll swap it to private.

internal/app/summon/command/temp_factory_test.go
sgnn7: I haven't researched it but given that it would require a whole subshell-launching harness for running and collecting the results which would also be pretty fragile, I don't think we should try to do that as part of this PR. We also use this exact same setup [elsewhere](https://github.com/cyberark/conjur-api-go/blob/f60f81f91778842e743b098f84210ba7d5f5d637/conjurapi/config_test.go#L120-L121).

demos/k8s-demo/env.sh
```diff
@@ -0,0 +1,2 @@
+export BACKEND_NAMESPACE="quick-start-backend-ns"
+export APP_NAMESPACE="quick-start-application-ns"
```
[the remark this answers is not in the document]
sgnn7: k

internal/app/summon/command/temp_factory.go
```diff
@@ -4,12 +4,10 @@ import (
...
 	"os"
 	"strings"
-
-	"github.com/mitchellh/go-homedir"
 )
 
-// DEVSHM is the location of a memory-mapped directory on Linux.
-const DEVSHM = "/dev/shm"
+// This is the location of a memory-mapped directory on Linux.
+const SHARED_MEMORY_DIR = "/dev/shm"
```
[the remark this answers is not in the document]
sgnn7: Description added to this const but going deeper into the module itself regarding "why we have it" is really outside of the scope of this task ("upgrade codebase to golang 1.12").

internal/app/summon/command/temp_factory_test.go
jonahx: > Since I realize now that we don't need the TempFactory.SharedMemoryDir public as nothing is using that outside of the class, I'll swap it to private.

internal/app/summon/command/temp_factory.go
```diff
@@ -4,12 +4,13 @@ import (
...
-	"github.com/mitchellh/go-homedir"
 )
 
-// DEVSHM is the location of a memory-mapped directory on Linux.
-const DEVSHM = "/dev/shm"
+// This is the location of a memory-mapped directory on Linux. We want
+// to use this location where possible since files saved there are not
+// saved on disk but kept in transient memory which is more secure than
+// the filesystem.
+const defaultSharedMemoryDir = "/dev/shm"
```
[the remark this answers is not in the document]
sgnn7: I get where you're coming from and we could clean this up at some point (I didn't write any of the original code for it) though to be fair it's probably of dubious value compared to all the other stuff we have to do imo. Ideally, this class wouldn't really care about `/dev/shm` since `/tmp` is almost always on `tmpfs` anyways. The module is also there not to make sure stuff secure but to ensure cleanup after temp fles are used from what I can tell.

> Anyway, you are right that technically it's not in scope, but in general I like to clean things up whenever I'm there anyway

This is the big issue to me with this PR: originally the only change needed in this file was a _single line of logic_ that was then: expanded to add tests since it had none (my own choice there), that was then expanded to rename shm variable, that was then expanded to add a secondary constructor, that then could have been cleaned up more, etc, etc which really isn't the way that we should work with PRs for issues that are unrelated to the task at hand especially with long CR cycles. Incremental fixes are fine but we can't make everything we touch perfect first time around.

internal/app/summon/command/temp_factory_test.go
sgnn7: > what is the behavior of this class, really? 

Looks to me to be in charge of creation and cleanup of all tempfiles created by it. It uses some logic to try to find a secure-ish place for created files.

> If you don't want to make that big a change at this point, that's ok with me.

Bigger change will touch other parts of the system and would open it up to more review changes and manual test invalidation for originally a very small change into the codebase which goes counter to the Git workflow in my opinion. If we need something specific fixed we should open new issues up.

> Test that TempFactory.path == TempFactory.SharedMemoryDir

I think that it doesn't need to be public but it would be good if we could use some Golang built-in to ensure that it's the right value. Hardcoded value like we use here is not really portable nor flexible.

internal/app/summon/command/temp_factory.go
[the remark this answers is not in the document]
jonahx: > to be fair it's probably of dubious value compared to all the other stuff we have to do imo.

I agree.  I don't care so much about this particular file and making sure that we keep our quality bar really high for new code or updates to code.

> This is the big issue to me with this PR: originally the only change needed in this file was a single line of logic that was then: expanded to add tests since it had none (my own choice there), that was then expanded to rename shm variable, that was then expanded to add a secondary constructor, that then could have been cleaned up more, etc, etc which really isn't the way that we should work with PRs for issues that are unrelated to the task at hand especially with long CR cycles. Incremental fixes are fine but we can't make everything we touch perfect first time around.

I agree here too.  I wasn't aware of all that.  My rule is: If you're there doing something, and notice something that should be fixed that will take anywhere from 15m - 2hrs or so (the higher times being justified only for bigger problems), then just do it.  If it's bigger than that, create an issue.

It sounds like you had already gone beyond the call of duty, and then added another beyond the call of duty on top of that, in this particular case.  So I think we're on the same page.

Issue #720, opened 2019-05-24
Update project status to remove beta


izgeri:
AC:
- [x] Homepage update (removing BETA icon, content updates)
- [x] README update (to remove beta warning, add alpha/beta labels)
- [x] blog post
- [ ] google group post

izgeri:
Blog post draft: https://cyberark365.sharepoint.com/:w:/s/Conjur/EZGSTIYobgFKmb3oKAez0m0BFoNlCtCS4g8LpigMXwhJNQ?e=b3Wkkd

Issue #713, opened 2019-05-22
All Secretless configuration code samples are updated to reflect new yml config format


jonahx:
Depends on #708

This story focuses on updating secretless.yml samples in code, demos, tutorials, and documentation. Updating the documentation on the listener / handler concept will happen as part of a separate effort in cyberark/secretless-docs#135.

## Repos to be updated
- [x] Secretless (all samples, docs, demos, etc - tests updated in #712)
- [x] Secretless Docs (all samples & references to config)
- [x] Kubernetes-Conjur-Demo (config used)
- [x] [Sidecar Injector](https://github.com/cyberark/sidecar-injector/blob/master/README.md#secretless-sidecar-injection-example) - README only

# AC:
- [x] All relevant references to secretless.yml are updated to use the new format
- [x] All demos and tutorials are updated to use the new secretless.yml format in documentation and code samples

izgeri:
Note - based on the [design doc](https://github.com/cyberark/secretless-broker/tree/master/design/simpler-configuration.md) we decided to continue supporting the old syntax going forward, so we _do not necessarily need to update all instances of old Secretless config YAML_. In particular, old versions in unused private repos are low priority right now.

izgeri:
There are draft PRs in:
- [Secretless](https://github.com/cyberark/secretless-broker/pull/736)
- [Kubernetes-Conjur-Demo](https://github.com/conjurdemos/kubernetes-conjur-demo/pull/64)
- [Sidecar Injector]()

that can be reviewed now, but should not be merged until the version bump in Secretless is ready.

sgnn7:
@izgeri All PRs look good, just need to wait for merge, real PR opening, and rebuilds.

PR #736, opened 2019-06-07
Update configuration samples in demos / docs

izgeri:

- Updates the configuration examples in `docs/`, `demos/` and the README to use the v2 format.
- Removes the deprecated `demos/full-demo`

Connected to #713
Connected to #720


sgnn7:
Edit: builds are failing for it so just need to be ensured that they pass when the other stuff is merged.

demos/k8s-demo/etc/secretless.yml
```diff
@@ -1,6 +1,6 @@
-version: 2
+version: "2"
```
sgnn7: All of the ymls should use `version: 2` without quotes (it's not needed)

izgeri: why does it matter if we have the quotes or not? it works both ways

Issue #725, opened 2019-05-30
Secretless has a test app to exercise the database handlers


izgeri:
# Description
In this task, we will write an application that can be deployed as a container to OpenShift. We will use this application to validate:
- The stability of Secretless Broker in handling database connections
- The request throughput is within 20% of direct-to-DB speed

## Requirements for the application
- Can be deployed as a container (eg there is a Docker image)
- Is reusable, so the code should be added to an existing GitHub repository or a new repository should be created
- Can connect to PostgreSQL and/or MySQL
- Can retrieve its database connection info from its environment, which may include address, username, and password or just address (if connecting via Secretless)
- The app should support the following connection modes:
  1. Open a connection, keep the connection open indefinitely, send regular / frequent queries, and print clear errors to stderr if the connection fails
  1. Open multiple connections at once, add query load, and print clear errors to stderr if the connection fails
  1. Continuously open connections and send queries, and print clear errors to stderr if the connection fails
  1. Open a connection, send a number of queries, measure the time for each query request, and print clear query time info to stdout

## Application tracking requirements
The application should track:
- The direct-to-DB query time data vs the query data via Secretless
- Any errors that the application logs, and the connection mode / database backend that caused that error (and any other relevant data)

# Open Questions

- Can the app be monitored so that we are alerted about errors reported in its logs?
- Are there other connection modes that the app should support? Are the current modes well enough defined?
- Are there other indicators that we need to capture and compare? Return codes? Connection properties?
- Are there ways to make this pluggable, so we can drop in new connection modules for things like HTTP, SSH, etc.?

# AC:
- [ ] An application exists that fulfills the functionality described above
- [ ] A manifest exists to deploy this application to OpenShift (without Secretless), retrieving the credentials for the database connection from Kubernetes Secrets

andytinkham:
If possible, in addition to the above, it would be nice to have 

- the application track times to open & close a DB connection, with & without Secretless
- compare the results of the query made through Secretless to that made without Secretless to ensure the same results come back. I'm not sure what the best result set for this would be - do we need to have a large query result? A query result with a variety of data types? Not sure exactly how this could fail if there's a bug (truncation feels more likely to me than mutation of the results, though).
- if there are other values like return codes from methods or connection properties that we think might look different between the with Secretless and without Secretless modes, compare those
- the ability to rotate the password during app execution to ensure Secretless keeps functioning. This might complicate the connect without Secretless portion.

**Additional Open Questions**
[...]

PR #737, opened 2019-06-07
725 Add perfagent runner (juxtaposer)

sgnn7:

Adds an agent that can test speed comparisons between direct DB connections and secretless

Connected to #725 

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/725-add-perfagent-runner/) though no tests should have changed

N/A


bin/build_perf_tool
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
+CURRENT_DIR=$("$(dirname "$0")/abspath")
+
+PERF_TEST_IMAGE="juxtaposer"
+PERF_TEST_IMAGE_TAG="$PERF_TEST_IMAGE:latest"
+
+echo "Current dir: $CURRENT_DIR"
```
izgeri: does this line need to persist?

bin/build_perf_tool
```diff
@@ -0,0 +1,14 @@
...
+
+pushd "$CURRENT_DIR/juxtaposer"
```
izgeri: This seems like the time to bring up naming questions 😬

- I'm looking at this and `juxtaposer` is just not evocative of what it is that this actually doing. I _think_ you called it `juxtaposer` because it's comparing direct-to-db vs via secretless (eg comparing them side-by-side), but that's not all it's doing - it's also validating long-time performance and validating performance under load.

  Infra has scripts to perf test Conjur, and they call them `exercise` and `pummel` (which is pretty evocative of what they're doing). I wonder if we could come up with a better name here.

- I wonder whether there are better names for the `formatter` and `tester` components - thinking of @jonahx's recent repost of [this link](http://www.carlopescio.com/2011/04/your-coding-conventions-are-hurting-you.html) warning against components with `-er` names.

- Putting this stuff in `bin/` seems odd. I'm wondering if we should be putting the integration tests in `test/integration` and putting this in `test/performance`, or some other place? Just in `performance-testing` maybe?

bin/juxtaposer/deploy/secretless.yml
```diff
@@ -0,0 +1,84 @@
+listeners:
```
izgeri: are you planning to upgrade to the v2 config?

bin/juxtaposer/deploy/juxtaposer_deployment_template.yml
```diff
@@ -0,0 +1,111 @@
...
+
+        volumeMounts:
+          - mountPath: /etc/${APP_NAME}
+            name: ${APP_NAME}-config
+            readOnly: true
+          - mountPath: /sock
+            name: ${APP_NAME}-sockets
+            readOnly: false
+      - name: secretless
+        image: cyberark/secretless-broker:latest
```
izgeri: can we have another image as a param? `cyberark/secretless-broker:0.7.1-becf48d` is the image from the latest commit on master - it would be great to run some tests against that with the v2 config

bin/build_perf_tool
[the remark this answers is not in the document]
sgnn7: I'll remove it

bin/juxtaposer/deploy/juxtaposer_deployment_template.yml
sgnn7: sure

bin/juxtaposer/deploy/secretless.yml
sgnn7: indeed it's just that the broker didn't have the v2 config merged yet

bin/juxtaposer/README.md
```diff
@@ -0,0 +1,113 @@
+# Perf Test Agent
+
+## Description
+
+This tool is used to compare timing data between arbitrary number of similar
+services to evaluate differences in speed between a specified baseline backend
+and the other backends.
+
```
izgeri: This is still missing:

- Status of this tool (eg `alpha`)
- Description of what it does, exactly (eg "Defines a perf test agent that can be deployed alongside Secretless to run through the following test scenarios: ...")

PR #739, opened 2019-06-11
734 Additional Performance Tests

sgnn7:

This PR adds the following perf agent (juxtaposer) capabilities:
- Ability to test reopened connections
- Ability to test backends with user-set number of parallel threads

Additional features added/changed:
- Added ability to stream-process results asynchronously
- Added ability to stream-process test runs
- Cleaned up timing-related logic
- Made round-based limit be the minimum rounds for all backends (ensures proper comparison)

Connected to #739 

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/734-additional-performance-tests/)

N/A - this _is_ the testing tool

### DIFF
--- a/bin/juxtaposer/config/config.go
+++ b/bin/juxtaposer/config/config.go
@@ -3,7 +3,6 @@ package config
 import (
 	"fmt"
 	"io/ioutil"
-	"strings"
 
 	"gopkg.in/yaml.v3"
 
@@ -34,27 +33,16 @@ type Backend struct {
 type Comparison struct {
 	BaselineBackend             string `yaml:"baselineBackend"`
 	BaselineMaxThresholdPercent int    `yaml:"baselineMaxThresholdPercent"`
+	RecreateConnections         bool   `yaml:"recreateConnections"`
 	Rounds                      string `yaml:"rounds"`
 	Silent                      bool   `yaml:"silent"`
-	Style                       string `yaml:"style"`
+	SqlStatementType            string `yaml:"sqlStatementType"`
 	Threads                     int    `yaml:"threads"`
-	Type                        string `yaml:"type"`
 }
 
 func (configuration *Config) verify() error {
-	if !strings.HasPrefix(configuration.Comparison.Type, "sql/") {
-		return fmt.Errorf("comparison type not supported: %s", configuration.Comparison.Type)
-	}
-
-	connectionType := configuration.Comparison.Type[4:]
-	if connectionType != "persistent" &&
-		connectionType != "recreate" {
-		return fmt.Errorf("comparison connection type not supported: %s",
-			connectionType)
-	}
-
-	if configuration.Comparison.Style != "select" {
-		return fmt.Errorf("comparison style supported: %s", configuration.Comparison.Style)
+	if configuration.Comparison.SqlStatementType != "select" {
+		return fmt.Errorf("comparison style supported: %s", configuration.Comparison.SqlStatementType)
 	}
 
 	if configuration.Comparison.Threads < 1 {
@@ -89,9 +77,9 @@ func NewConfiguration(configFile string) (*Config, error) {
 	configuration := Config{
 		Comparison: Comparison{
 			BaselineMaxThresholdPercent: 120,
+			RecreateConnections:         false,
 			Rounds:                      "1000",
-			Style:                       "select",
-			Type:                        "sql/persistent",
+			SqlStatementType:            "select",
 			Threads:                     1,
 		},
 		Formatters: map[string]formatter_api.FormatterOptions{
--- a/bin/juxtaposer/config/config.go
+++ b/bin/juxtaposer/config/config.go
@@ -37,6 +37,7 @@ type Comparison struct {
 	Rounds                      string `yaml:"rounds"`
 	Silent                      bool   `yaml:"silent"`
 	Style                       string `yaml:"style"`
+	Threads                     int    `yaml:"threads"`
 	Type                        string `yaml:"type"`
 }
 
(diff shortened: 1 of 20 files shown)


bin/juxtaposer/main.go
```diff
@@ -3,167 +3,74 @@ package main
...
-	log.Println("Backends:", len(config.Backends))
 	for backendName, backendConfig := range config.Backends {
 		backendNames = append(backendNames, backendName)
 
 		log.Printf("Setting up backend: %s", backendName)
 
-		if config.Comparison.Type != "sql" {
-			return nil, nil, fmt.Errorf("ERROR: Comparison type supported: %s", config.Comparison.Type)
+		// Sanity check
+		if !strings.HasPrefix(config.Comparison.Type, "sql/") {
```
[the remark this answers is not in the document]
sgnn7: yeah I guess that could be collapsed 👍

bin/juxtaposer/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+
+type Comparison struct {
+	BaselineBackend             string `yaml:"baselineBackend"`
+	BaselineMaxThresholdPercent int    `yaml:"baselineMaxThresholdPercent"`
+	Rounds                      string `yaml:"rounds"`
+	Silent                      bool   `yaml:"silent"`
+	Style                       string `yaml:"style"`
+	Threads                     int    `yaml:"threads"`
+	Type                        string `yaml:"type"`
+}
```
jonahx: The names `Type` and `Style` aren't descriptive.

I'd suggest:

- `Type` => `RecreateConnections` (true / false). 
- `Style` => `SqlStatementType`

bin/juxtaposer/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+	Password    string `yaml:"password"`
+	Port        string `yaml:"port"`
+	SslMode     string `yaml:"sslmode"`
+	Socket      string `yaml:"socket"`
+	Username    string `yaml:"username"`
+}
+
+type Comparison struct {
+	BaselineBackend             string `yaml:"baselineBackend"`
+	BaselineMaxThresholdPercent int    `yaml:"baselineMaxThresholdPercent"`
```
jonahx: Do we need this?  The stats we were doing yesterday -- % of requests over threshold A, B, C, D, etc -- seems to be a superset of this stat.  

Or would that be too big a change at this point?

bin/juxtaposer/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	err = yaml.Unmarshal(yamlFile, &configuration)
+	if err != nil {
+		return nil, err
+	}
+
+	// Slice out any backends which are ignored
+	filteredBackends := map[string]Backend{}
+	for backendName, backendConfig := range configuration.Backends {
+		if backendConfig.Ignore == false {
```
jonahx: This is probably a change to save for later, but I've given this more thought and I really think cluttering up our config object and the logic which uses it with the `Ignore` field is a mistake.  If users want to ignore something, commenting it out of their config file is not too onerous.  This is something where the convenience of the feature (and I'm not even sure I like the feature)  is nowhere near worth the extra complexity in the code.

bin/juxtaposer/formatter/stdout/stdout.go
```diff
@@ -14,22 +15,22 @@ func NewFormatter(options formatter_api.FormatterOptions) (formatter_api.OutputF
...
-		map[string]string{"name": "Errors", "nameFormat": "%7s", "valueFormat": "%7d"},
+		map[string]string{"name": "Total", "nameFormat": "%18s", "valueFormat": "%18v"},
+		map[string]string{"name": "Rounds", "nameFormat": "%9s", "valueFormat": "%9d"},
+		map[string]string{"name": "Errors", "nameFormat": "%9s", "valueFormat": "%9d"},
 		map[string]string{"name": "Succ %", "nameFormat": "%9s", "valueFormat": "%9.2f"},
 		map[string]string{"name": "Thresh %", "nameFormat": "%9s", "valueFormat": "%9.2f"},
 	}
 
-	dividerString := strings.Repeat("-", 146)
+	dividerString := strings.Repeat("-", 158)
```
jonahx: Similar to my comment about rolling your own stats functions, I don't think we should be rolling your own ascii table formatter.  This is stock utility code.  Eg, 

https://github.com/olekukonko/tablewriter

would probably work well.

This change doesn't need to be made now.

bin/juxtaposer/config/config.go
sgnn7: Fixed

bin/juxtaposer/config/config.go
sgnn7: Yeah without this field, none of the data works anymore ☹️ . We'll need to update the tool to perform the same operations that our python/ruby tools do soon to fix this.

bin/juxtaposer/config/config.go
sgnn7: That's fair though to be fair it's only 3 lines. Either way, we can change this later.

bin/juxtaposer/formatter/stdout/stdout.go
sgnn7: Yup, we can figure out how to do the numbers after we also include the rolling avg for the baseline too.

jonahx: I have some architectural ideas around the formatter, but those will def wait for the full review.  Just planting a flag.

bin/juxtaposer/main.go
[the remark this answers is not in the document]
sgnn7: Resolved by followup commits

Issue #756, opened 2019-06-18
Add more documentation to Juxtaposer


sgnn7:
Perf agent needs more info about how it will be run to be really useful

AC:
- [ ] Juxtaposer has information about how to run it locally
- [ ] Juxtaposer has information about how to deploy it

sgnn7:
Linked PR: https://github.com/cyberark/secretless-broker/pull/757

PR #757, opened 2019-06-18
Added info to README about how to run perf agent locally

sgnn7:
This change adds more information needed to run the perf agent
(juxtaposer) locally to evaluate its functionality.

Connected to #756 

README.md

N/A

### DIFF
--- a/bin/juxtaposer/README.md
+++ b/bin/juxtaposer/README.md
@@ -44,7 +44,7 @@ Note: More comparison types may be added in the future.
 
 ---
 
-## Running the code
+## Running the code locally
 
 You can run and test this code on your machine easily with a few things:
 - [Docker](https://docker.com)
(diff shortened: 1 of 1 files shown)


bin/juxtaposer/README.md
```diff
@@ -44,6 +44,180 @@ Note: More comparison types may be added in the future.
 
 ---
 
+## Running the code
+
+You can run and test this code on your machine easily with a few things:
+- [Docker](https://docker.com)
+- [Golang](https://golang.org/dl/) 1.12 or higher
+- [Secretless-broker](https://github.com/cyberark/secretless-broker) (optional)
+- A writable folder for shared socket files (e.g. `/sock` in these configurations)
```
jonahx: Is this optional too?  What if you were only testing `tcp` backends?

bin/juxtaposer/README.md
```diff
@@ -44,6 +44,180 @@ Note: More comparison types may be added in the future.
...
+
+- Clone the [secretless-broker repository](https://github.com/cyberark/secretless-broker) (`git clone https://github.com/cyberark/secretless-broker`).
+- Create the shared socket folder if it's missing (`sudo mkdir /sock`).
+- Create the following `secretless.yml` file in that folder:
+<details>
+  <summary><code>secretless.yml</code></summary>
+  <p>
+
+```yaml
+listeners:
```
jonahx: how come we're using a v1 config here?

bin/juxtaposer/README.md
```diff
@@ -44,6 +44,180 @@ Note: More comparison types may be added in the future.
...
+
+comparison:
+  baselineBackend: pg_direct
+#  baselineBackend: mysql_direct
+#  type: sql/persistent
+#  style: select
+#  rounds: 1000
+#  rounds: infinity
+#  baselineMaxThresholdPercent: 120
+  threads: 5
```
jonahx: why are these commented out?

bin/juxtaposer/README.md
sgnn7: It's optional too - I'll fix in followup commit

bin/juxtaposer/README.md
sgnn7: This is my local config - I'll change to v2 in followup commit

bin/juxtaposer/README.md
sgnn7: For reference so that you don't need to hunt them through the readme. It's similar to lots of files in /etc/ that have example variables.

Issue #707, opened 2019-05-22
Secretless configuration is intuitive and simple


jonahx:
ETA for completion of this epic: June 6, 2019
Confidence level of ETA accuracy: 90%
[Design document](https://github.com/cyberark/secretless-broker/blob/master/design/simpler-configuration.md)

## Link to Aha
https://cyberark.aha.io/features/AAM-92

### Secretless UX is intuitive and simple: improved secretless.yml and docs
This epic is for simplifying the user-facing experience of configuring Secretless.  It removes unnecessary concepts and presents better chrome to the user.  This improvement comes in two main flavors:

1. Shorter, simpler `secretless.yml` configuration files
2. Shorter, simpler "concept" documentation

The linked tasks break down how this will be achieved.

### Development
- [ ] Parent of #708 - Configuration parsing in config package handles new config format
- [ ] Parent of #714 - Secretless configuration syntax is versioned 

### Testing
- [ ] Parent of #712 - All Secretless test cases use new yml config format

### Documentation
- [ ] Parent of #713 - All Secretless configuration code samples are updated to reflect new yml config format
- [ ] Parent of cyberark/secretless-docs#135 - Listener / Handler references are updated to refer to single handler construct

## Additional Dev Notes

1. Open question: How should Secretless fail on invalid configuration when watching ?
2. How should authenticators handle parsing and validation of service configuration.  Idea:
    - Secretless handle the global configuration in the config package
    - Proxy delegates validation to authenticator (fail fast)
    - Proxy passes the `map[string]interface` service configuration to the authenticator
3. In implementing, multiple authenticators under http protocol (see #708)

## Future work
- We have planned future work to further simplify and improve the plugin interface. This will be done in a follow-on effort.
- Update the configuration CRD to follow the new model (#715)
- Update the configuration data structures (#710)

PR #765, opened 2019-06-24
Draft of plugin refactor design doc

jonahx:
Feedback on plugin feature doc

### DIFF
--- a/plugin-feature-doc.md
+++ b/plugin-feature-doc.md
@@ -1,0 +1,111 @@
+# Plugin API is simple and well-documented
+
+The current plugin API has an overly large surface area, poor documentation,
+and unclear names.  The objective is to improve this, focusing on the UX of
+primary use-case: creating new service authenticators.
+
+Aha Card: https://cyberark.aha.io/features/AAM-92
+
+- [Objective](#objective)
+- [Experience (Summary)](#experience-summary)
+- [Experience (Detailed)](#experience-detailed)
+- [Technical Details](#technical-details)
+- [Testing](#testing)
+- [Documentation](#documentation)
+- [Open Questions](#open-questions)
+- [Stories](#stories)
+- [Future Work](#future-work)
+
+### Objective
+
+- Simplify Plugin API so authors of new service authenticators have a good
+  experience which requires learning only concepts relevant to their goal.
+- Write clear documentation to make the simplified API accessible to new
+  contributors
+
+### Experience (Summary)
+
+The new plugin API will support writing new service authenticators only.  Only
+a minimal interface will be required.
+
+### Experience (Detailed)
+
+Authors will need to implement only `PluginInfo` and a new
+`ServiceAuthenticator` interface.
+
+`APIVersion` will be moved into `PluginInfo`, and the `plugin_type` will be
+explicitly specified:
+
+```go
+var PluginInfo = map[string]string{
+  "api_version": "2.0",
+  "plugin_types": "service_authenticator",
+  "version":     "0.0.7",
+  "id":          "test-plugin",
+  "name":        "Test Plugin",
+  "description": "Test plugin to demonstrate plugin functionality",
+}
+```
+
+Implementing authenticators will look like:
+
+```go
+func ServiceAuthenticators() map[string]func(plugin_v2.AuthenticatorOptions) plugin_v2.Authenticator {
+  return map[string]func(plugin_v2.AuthenticatorOptions) plugin_v2.Authenticator{
+    "sample-authenticator-plugin": samplePlugin.NewAuthenticator,
+  }
+}
+```
+
+With the precise interface of an `Authenticator` TBD.
+
+##### Assumptions
+
+- The primary interest in plugins, at least at first, will be in writing new service authenticators.
+
+##### Workflow
+
+1. Write and compile one or more authenticators according to the above
+1. Place the resulting `.so` files in the correct directory (currently `/usr/local/lib/secretless`)
+1. Restart secretless
+
+### Technical Details
+
+- Per above, we'll refactor so that only the new, above described minimal API is public
+- All the old interfaces will be moved into internal
+- We'll have a generic type doing what's currently repeated by all listeners
+- We'll need to deal with the exceptions represented by the http handler --
+  what we were calling "sub-handlers" in previous discussions (as a placeholder
+  name).
+- Once handlers and listeners are combined, go back and update our config code
+  to use them directly.
+
+### Testing
+
+**Integration**
+
+- A new example plugin to replace current one 
+- Tests showing it loads and can be used successfully as authenticator specified in a v2 secretless.yml
+
+**Unit**
+
+A major aim of this refactor will be to have components that are unit testable.
+Thus every new nontrivial type or func created will likely have unit tests.
+What those types are is still TBD.
+
+### Open Questions
+
+- Per above, the "sub handler" question
+- Definitive list of all places in `plugin.Manager` and elsewhere that we'll need to update
+- API of `ServiceAuthenticator` interface
+- Design of the generic "listener engine"
+
+### Stories
+
+The epic has links to all current stories:
+
+- https://app.zenhub.com/workspaces/secretless-5c9d073c270ed03f5cb98c95/issues/cyberark/secretless-broker/693
+
+### Future Work
+
+- Further related refactoring of the plugin.Manager, Proxy, etc.


plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
+# Plugin API is simple and well-documented
+
+The current plugin API has an overly large surface area, poor documentation,
+and unclear names.  The objective is to improve this, focusing on the UX of
+primary use-case: creating new service authenticators.
+
+Aha Card: https://cyberark.aha.io/features/AAM-92
```
izgeri: It's not clear to me whether this is correct - I've asked PM for clarification, but I haven't heard back.

I think AAM-92 was revised to refer only to the configuration refactoring we completed in #707. I'm not sure there is an Aha card for this, since it's engineering driven work.

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+```go
+var PluginInfo = map[string]string{
+  "api_version": "2.0",
+  "plugin_types": "service_authenticator",
+  "version":     "0.0.7",
+  "id":          "test-plugin",
+  "name":        "Test Plugin",
+  "description": "Test plugin to demonstrate plugin functionality",
+}
+```
```
izgeri: there's nowhere else in the doc where we dig into the `PluginInfo` and define its components - that seems worth doing here

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+
+```go
+func ServiceAuthenticators() map[string]func(plugin_v2.AuthenticatorOptions) plugin_v2.Authenticator {
+  return map[string]func(plugin_v2.AuthenticatorOptions) plugin_v2.Authenticator{
+    "sample-authenticator-plugin": samplePlugin.NewAuthenticator,
+  }
+}
+```
+
+With the precise interface of an `Authenticator` TBD.
```
izgeri: Is that a TODO for this document? when will that be defined?

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+  "id":          "test-plugin",
+  "name":        "Test Plugin",
+  "description": "Test plugin to demonstrate plugin functionality",
+}
+```
+
+Implementing authenticators will look like:
+
+```go
+func ServiceAuthenticators() map[string]func(plugin_v2.AuthenticatorOptions) plugin_v2.Authenticator {
```
izgeri: this might be a "too long since I've been in the code" question, but is this what will replace [this part](https://github.com/cyberark/secretless-broker/blob/master/test/plugin/example/cmd/main.go#L27) of the example plugin?

can we break this out in the design doc like we did for the config, where we said "here is what the example plugin looked like before" and "here is what we propose to look like it now"? I feel like that would make the improvement clearer and would give us an opportunity to also address how we are dropping support for other plugin types for now.

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+##### Workflow
+
+1. Write and compile one or more authenticators according to the above
+1. Place the resulting `.so` files in the correct directory (currently `/usr/local/lib/secretless`)
+1. Restart secretless
+
+### Technical Details
+
+- Per above, we'll refactor so that only the new, above described minimal API is public
+- All the old interfaces will be moved into internal
```
izgeri: Do we want to refactor those based on our current (updated) understanding of them, or leave that until there is actually a demand for those types of public plugins?

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+- We'll have a generic type doing what's currently repeated by all listeners
+- We'll need to deal with the exceptions represented by the http handler --
+  what we were calling "sub-handlers" in previous discussions (as a placeholder
```
izgeri: we called them different `authentication strategies` for the HTTP service authenticator

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+  name).
+- Once handlers and listeners are combined, go back and update our config code
+  to use them directly.
+
```
izgeri: I don't have a clear picture from this of what's happening in the internal code. we're going to change the interface - move some parts of it to `internal/`, revise the public pieces that remain. but:

- are we also updating the internal code to implement the v2 interface?
- in a related question, are we going to reorganize the internal code to remove the distinction between listeners and handlers? eg put code in a single (maybe) `authenticators` directory, remove internal references to listeners/handlers, etc?
- is the v2 config going to become part of the public v2 plugin interface, or is it going to remain its own package?
- are we going to update the internal components that still use the v1 config, or will that tech debt remain after this effort?
- what other tech debt are we not capturing in what is described above, especially around the fundamental concepts that are part of the code? it would be good to capture that in more detail in the `future work` section

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+**Integration**
+
+- A new example plugin to replace current one 
+- Tests showing it loads and can be used successfully as authenticator specified in a v2 secretless.yml
+
+**Unit**
+
+A major aim of this refactor will be to have components that are unit testable.
+Thus every new nontrivial type or func created will likely have unit tests.
+What those types are is still TBD.
```
izgeri: do we need some spikes to try to determine the types? imo the design doc isn't really "complete" until we have the full design defined (or at least reasonably close to it). what info are we missing right now?

plugin-feature-doc.md
jonahx: > this might be a "too long since I've been in the code" question, but is this what will replace this part of the example plugin?

Yep.

> can we break this out in the design doc like we did for the config, where we said "here is what the example plugin looked like before" and "here is what we propose to look like it now"?

Yep, but to be clear as of now this one is a simple rename.  So there's no improvement beyond that.  And  it looks like that name will actually be `ServiceConnecters()` or `Connecters()`

plugin-feature-doc.md
jonahx: Not for this document (at least in my mind).

What I meant there is that when, internally in the refactor, we combine the artists currently called `Listeners` and `Handlers` we'll have a new thing called `Authenticator` and precisely what fields and methods it has will be determined as part of that refactoring.  So I'm considering this an "in the weeds" question that I wasn't going to answer here.

plugin-feature-doc.md
jonahx: You're right.  We should mirror this language in the code.  "sub handler" was just a temp name Kumbi and I were using during some previous conversations.

plugin-feature-doc.md
jonahx: We want to refactor `Listeners` and `Handlers`.  I'm not sure if want to touch the other stuff yet.  I mean, it could use some love to but I'm wary of scope creep so it may make sense to just not touch those yet.  I think that's my official position atm, with the caveat it may change once I'm in the code and can make better estimates.

plugin-feature-doc.md
jonahx: > are we also updating the internal code to implement the v2 interface?

yes.  the current "adapter" approach we're using was intended to be temporary, to keep us focused on just delivering the config changes.

> in a related question, are we going to reorganize the internal code to remove the distinction between listeners and handlers? eg put code in a single (maybe) authenticators directory, remove internal references to listeners/handlers, etc?

yep, that's the basic plan

> is the v2 config going to become part of the public v2 plugin interface, or is it going to remain its own package?

i haven't thought super deeply about this, and atm i don't feel strongly one way or the other.

> are we going to update the internal components that still use the v1 config, or will that tech debt remain after this effort?

i think we should completely eliminate it completely from the code, and no longer support it.  if ending support for it isn't an option for some reason, we should still end support for it.  if it's *really* not an option, we can support it using an adapter approach: ie, parse it and convert it internally into a v2 config object.

> are we going to update the internal components that still use the v1 config, or will that tech debt remain after this effort?

update.

> what other tech debt are we not capturing in what is described above, especially around the fundamental concepts that are part of the code? it would be good to capture that in more detail in the future work section

there's debt around the plugin manager, the connection managers, as well as others.  i have notes and even some local branches where i've started work.  i can update the document to include that stuff?   but technically it's not part of the "plugin" refactor (again, at least in my mind).  i can think about which parts are "most adjacent" though...

plugin-feature-doc.md
jonahx: Hmm.... Yeah we could.  As I said, some of this is really "in the weeds" / "won't know for sure until we do it" stuff.  But we could take the spike approach to minimize that as much as possible.  Is that the preferred approach?

izgeri: I think it would be useful to make a list of "open questions" for things that we don't have enough information to decide right now so that we can look at that list and decide what merits a spike or not.

This document is our chance to collaboratively plan beforehand what we're going to do, and the more we can explicitly plan in the document, the less time we'll have to spend regrouping on the design mid-implementation.

plugin-feature-doc.md
```diff
@@ -0,0 +1,111 @@
...
+
+`APIVersion` will be moved into `PluginInfo`, and the `plugin_type` will be
+explicitly specified:
+
+```go
+var PluginInfo = map[string]string{
+  "api_version": "2.0",
+  "plugin_types": "service_authenticator",
+  "version":     "0.0.7",
+  "id":          "test-plugin",
```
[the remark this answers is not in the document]
sgnn7: Yeah the protocol support info is provided by the handler itself

plugin-feature-doc.md
sgnn7: just a heads up that it should be `Connectors` and not `Connecters`.

jonahx:
closing.  this will be re-opened in a separate pull.  the work on the feature doc has gone on a long time and changed the scope of this.

PR #781, opened 2019-07-03
Replace check_style linting with codeclimate golint engine

jonahx:
#### What does this PR do (include background context, if relevant)?

From the commit message:

    - Change bin/check_style script to call codeclimate via Docker
    - Remove check_style task from Jenkinsfile, since it will now be run by
    Code Climate
    - Turn on codeclimate golint engine and configure it to match the
    directories we were testing with check_style
    - Update CONTRIBUTING.md to document the new codeclimate use

I also created a test branch and PR to verify that Code Climate will in fact now fail on linting errors:

https://github.com/cyberark/secretless-broker/pull/783

This confirmed CC is failing linting errors as expected.

#### What is the status of the manual tests?

Not performed.

PR #794, opened 2019-07-13
Integration tests run in parallel

jonahx:
This makes all integration tests run in parallel, so they are much faster (total running on Jenkins is now about 3 minutes)

Don't think it has one.

Definitely start on the Commits tab.

All commits have been broken down into small, logically coherent pieces that can be reviewed in isolation.  This will make reviewing easy.

Not run.

#781 should probably be merged before this, though it's not required as long as both are merged relatively quickly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,7 +98,7 @@ $ ./bin/build_darwin
 
 ## Testing
 
-**Prerequisites**
+### Prerequisites
 
 * **Docker** You need Docker to run the tests.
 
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -113,6 +113,38 @@ Then run the test cases:
...
+- `./start` (required) - Performs setup work -- eg, spinning up test
+   containers, populating a database with test data, etc.
+- `./test` (required) - Runs the actual tests.  It is assumed to produce go
+   test output on stdout.
+- `./stop` (optional) - Performs cleanup work.
+
+Hence, adding a new integration requires two steps:
+
+1. Creating a folder with test scripts as described above. 
+1. Adding a new entry to the `Jenkinsfile` to exercise it using the
```
doodlesbykumbi: TIL to put a comma before "which" in nonrestrictive phrases.

More a suggestion than anything ...
I wonder if we can rephrase this so as to reduce the length of the sentence? It might make it simpler to parse.
For some reason the comma at `output file, which will` was throwing me off a bit.

jonahx: Sure, I can do that.

Jenkinsfile
```diff
@@ -19,49 +19,97 @@ pipeline {
...
+            junit 'test/conjur/junit.xml'
           }
         }
 
-        stage('Demo tests') {
+        stage('Integration: HTTP Basic Auth') {
           steps {
-            sh './bin/test_demo'
+            sh './bin/run_integration http_basic_auth'
+            // junit 'test/http_basic_auth/junit.xml'
```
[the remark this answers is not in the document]
jonahx: It wasn't working, and iirc it wasn't producing output.  Like, it looked like it never was.  I'll take another look and if that's correct, I'll just remove the commented line unless you think otherwise...

Issue #727, opened 2019-05-31
Secretless has a NOTICES.txt


izgeri:
Secretless has a NOTICES.txt that details the open source licenses of its dependencies. Ideally, this is auto-updated as the dependencies change.

PR #797, opened 2019-07-15
Add NOTICES.txt to the project

izgeri:
Adds a NOTICES.txt to the project with OSS dependency license information and some utilities for tracking dependency changes. At this stage, the dependency license tracking is still largely manual, but ideally we can continue to progress to automated dependency license tracking pre-releases.

Connected to #727


bin/check_dependencies
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -uo pipefail
```
sgnn7: Why no `-e`?

bin/check_dependencies
```diff
@@ -0,0 +1,29 @@
...
+# get a new list of project licenses
+# this command syntax is wild but it's what the license_finder image requires
+docker run --rm \
+  -v $toplevel_dir:/scan \
+  licensefinder/license_finder \
+  /bin/bash -lc "cd /scan && license_finder" \
+  > $toplevel_dir/assets/updated_licenses.txt
+
+# check if there are any differences; if not, exit
+diff=$(diff $toplevel_dir/assets/updated_licenses.txt $toplevel_dir/assets/license_finder.txt)
```
sgnn7: Is diff common enough to expect it? If so, ignore the comment.

bin/check_dependencies
```diff
@@ -0,0 +1,29 @@
...
+
+if [ -z "$diff" ]; then
+  echo "No dependencies have changed."
+else
+  echo -e "\nWarning - dependencies have changed. NOTICES.txt will need to be updated.\n"
+  echo "$diff"
+fi
```
sgnn7: I don't see any error-type exits here - I'd assume that we eventually want this automated so the exit code of "1" would go a long way to helping automate this even in a simple checking script on Jenkins.

NOTICES.txt
```diff
@@ -0,0 +1,1494 @@
+
```
sgnn7: After re-reading this, I think we really need something better than this and split up into parts.

sgnn7:
PS: CodeClimate is also complaining about something.

bin/check_dependencies
izgeri: With it the script would die after the docker run or after diff

NOTICES.txt
izgeri: I based this on the doc that Pivotal requires for us, and I think it's much easier to read than the typical one our org uses. What specifically would you suggest that we change?

bin/check_dependencies
izgeri: For maintainers? I think so

bin/check_dependencies
izgeri: This isn't currently part of the pipeline, I don't want it to be until the end to end of generating the file is actually automated, so for now this is a util to check if the txt needs to be updated pre-release.

I'm fine with maybe adding those in later, but right now this is manual and informational.

izgeri:
@sgnn7 I fixed the code climate issues, this should be gtg now - but some of your comments I responded to rather than changing what I did. lmk what you think.

bin/check_dependencies
sgnn7: Can you just add a comment or a TODO on this for that then? I'd hate for it to be lost in this PR and never fixed up later.

PR #803, opened 2019-07-29
DO NOT MERGE: Only exists for sharing inline notes

jonahx:
Some inline notes I'm talking while reviewing the codebase for the refactor.


jonahx:
This should have been close a while ago.

Issue #805, opened 2019-07-30
MySQL connector port is defaulted to 3306


sgnn7:
AC:
- [ ] MySQL does not need a `port` field defined if the backend is on the default 3306 port

PR #806, opened 2019-07-31
Made MySQL connector able to guess the 3306 port if it is not provided

sgnn7:
Since in most cases, we know that the MySQL port is 3306, it is
superfluous to need that in the config if the user is connecting to the
default port setting. This commit also reorganizes the code a bit for
better testability and adds a few tests around this feature.

Connected to https://github.com/cyberark/secretless-broker/issues/805

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/805-mysql-default-port-set/)

### DIFF
--- a/internal/app/secretless/handlers/mysql/connection_details.go
+++ b/internal/app/secretless/handlers/mysql/connection_details.go
@@ -12,6 +12,46 @@ type ConnectionDetails struct {
 	Options  map[string]string
 }
 
+const DefaultMySQLPort = uint(3306)
+
+// NewConnectionDetails is a constructor of ConnectionDetails structure from a
+// map of credentials.
+func NewConnectionDetails(credentials map[string][]byte) (
+	*ConnectionDetails, error) {
+
+	connDetails := &ConnectionDetails{Options: make(map[string]string)}
+
+	if host := credentials["host"]; host != nil {
+		connDetails.Host = string(credentials["host"])
+	}
+
+	connDetails.Port = DefaultMySQLPort
+	if credentials["port"] != nil {
+		port64, _ := strconv.ParseUint(string(credentials["port"]), 10, 64)
+		connDetails.Port = uint(port64)
+	}
+
+	if credentials["username"] != nil {
+		connDetails.Username = string(credentials["username"])
+	}
+
+	if credentials["password"] != nil {
+		connDetails.Password = string(credentials["password"])
+	}
+
+	delete(credentials, "host")
+	delete(credentials, "port")
+	delete(credentials, "username")
+	delete(credentials, "password")
+
+	connDetails.Options = make(map[string]string)
+	for k, v := range credentials {
+		connDetails.Options[k] = string(v)
+	}
+
+	return connDetails, nil
+}
+
 // Address returns a string representing the network location (host and port)
 // of a MySQL server.  This is the string you would would typically use to
 // connect to the database -- eg, in cmd line tools.
--- a/internal/app/secretless/handlers/mysql/connection_details_test.go
+++ b/internal/app/secretless/handlers/mysql/connection_details_test.go
@@ -1,0 +1,91 @@
+package mysql
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExpectedFields(t *testing.T) {
+	credentials := map[string][]byte{
+		"host":     []byte("myhost"),
+		"port":     []byte("1234"),
+		"username": []byte("myusername"),
+		"password": []byte("mypassword"),
+	}
+
+	expectedConnDetails := ConnectionDetails{
+		Host:     "myhost",
+		Port:     1234,
+		Username: "myusername",
+		Password: "mypassword",
+		Options:  make(map[string]string),
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	if err == nil {
+		assert.EqualValues(t, expectedConnDetails, *actualConnDetails)
+	}
+}
+
+func TestDefaultPort(t *testing.T) {
+	credentials := map[string][]byte{
+		"host":     []byte("myhost"),
+		"username": []byte("myusername"),
+		"password": []byte("mypassword"),
+	}
+
+	expectedConnDetails := ConnectionDetails{
+		Host:     "myhost",
+		Port:     DefaultMySQLPort,
+		Username: "myusername",
+		Password: "mypassword",
+		Options:  make(map[string]string),
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	if err == nil {
+		assert.EqualValues(t, expectedConnDetails, *actualConnDetails)
+	}
+}
+
+func TestUnexpectedFieldsAreSavedAsOptions(t *testing.T) {
+	credentials := map[string][]byte{
+		"host":     []byte("myhost"),
+		"port":     []byte("1234"),
+		"foo":      []byte("5432"),
+		"username": []byte("myusername"),
+		"bar":      []byte("data"),
+		"password": []byte("mypassword"),
+	}
+
+	expectedOptions := map[string]string{
+		"foo": "5432",
+		"bar": "data",
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	if err == nil {
+		assert.EqualValues(t, expectedOptions, (*actualConnDetails).Options)
+	}
+}
+
+func TestAddress(t *testing.T) {
+	credentials := map[string][]byte{
+		"host": []byte("myhost2"),
+		"port": []byte("12345"),
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	if err == nil {
+		assert.EqualValues(t, "myhost2:12345", (*actualConnDetails).Address())
+	}
+}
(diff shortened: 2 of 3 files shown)


internal/app/secretless/handlers/mysql/connection_details.go
```diff
@@ -12,6 +12,44 @@ type ConnectionDetails struct {
 	Options  map[string]string
 }
 
+// NewConnectionDetails is a constructor of ConnectionDetails structure from a
```
jonahx: I like pulling this out into its own file 👍

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
+package mysql
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConnectionDetailsCorrectlyParsesConnectionDetails(t *testing.T) {
```
jonahx: Since this is the test file for connection details (ie, we're in`connection_details_test`), we can dispense with `ConnectionDetails` in any of the names.  After reading through all the tests, it seems what we're actually testing here is that the basic, expected information is parsed correctly.

Therefore, suggested name: `TestExpectedFields`.

Additionally, all the tests here are testing different aspects of `NewConnectionDetails`.  We might consider:

```go
func TestNewConnectionDetails(t *testing.T) {
    t.Run("Test Expected Fields", func(t *testing.T) { ... })
    t.Run("Test Default Port", func(t *testing.T) { ... })
    // etc
}
```
though that's probably a premature optimization at this point, and improving the current names and relying on the file context is enough.

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+	credentials := make(map[string][]byte)
+
+	credentials["host"] = []byte("myhost")
+	credentials["port"] = []byte("1234")
+	credentials["username"] = []byte("myusername")
+	credentials["password"] = []byte("mypassword")
+
```
jonahx: Let's use a literal here, ie, like:
```go
	creds := map[string][]byte{
		"host": []byte("myhost"),
		"port": []byte("1234"),
                "username": []byte("myusername"),
                "password": []byte("mypassword"),
	}
```

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+func TestPortDefaultsTo3306IfNotProvided(t *testing.T) {
+	credentials := make(map[string][]byte)
+
+	credentials["host"] = []byte("myhost")
+	credentials["username"] = []byte("myusername")
+	credentials["password"] = []byte("mypassword")
+
+	expectedConnDetails := ConnectionDetails{
+		Host:     "myhost",
+		Port:     3306,
```
jonahx: Similar to my other comment, this [magic number](https://refactoring.guru/replace-magic-number-with-symbolic-constant) should refer to a public constant on the impl file, rather than be hardcoded in both places.

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	assert.EqualValues(t, expectedConnDetails, *actualConnDetails)
+}
+
+func TestPortDefaultsTo3306IfNotProvided(t *testing.T) {
+	credentials := make(map[string][]byte)
+
+	credentials["host"] = []byte("myhost")
+	credentials["username"] = []byte("myusername")
+	credentials["password"] = []byte("mypassword")
```
jonahx: use single assignment to map literal

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Port:     1234,
+		Username: "myusername",
+		Password: "mypassword",
+		Options:  make(map[string]string),
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	assert.EqualValues(t, expectedConnDetails, *actualConnDetails)
```
jonahx: I could be wrong, but iirc this statement will error if the `err` is not nil.  Ie, it won't fail as an assertion, but will error with a go runtime error.  You might expect that it would stop processing if an assertion fails, but it doesn't work that way (again, iirc).  So you have to wrap the 2nd assertion in an `if`, sadly.

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Port:     3306,
+		Username: "myusername",
+		Password: "mypassword",
+		Options:  make(map[string]string),
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	assert.EqualValues(t, expectedConnDetails, *actualConnDetails)
```
jonahx: Same comment about wrapping in `if` here...

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+func TestConnectionDetailsAddsUnknownCredentialsToOptions(t *testing.T) {
+	credentials := make(map[string][]byte)
+
+	credentials["host"] = []byte("myhost")
+	credentials["port"] = []byte("1234")
+	credentials["foo"] = []byte("5432")
+	credentials["username"] = []byte("myusername")
+	credentials["bar"] = []byte("data")
+	credentials["password"] = []byte("mypassword")
```
jonahx: map literal

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Options:  make(map[string]string),
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	assert.EqualValues(t, expectedConnDetails, *actualConnDetails)
+}
+
+func TestPortDefaultsTo3306IfNotProvided(t *testing.T) {
```
jonahx: I think `IfNotProvided` is redundant (implied by the definition of "default") and for the same magic number argument I noted in other comments we shouldn't have 3306 in the name.  

Suggested improvement: `TestDefaultPort`

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Options:  make(map[string]string),
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	assert.EqualValues(t, expectedConnDetails, *actualConnDetails)
+}
+
+func TestConnectionDetailsAddsUnknownCredentialsToOptions(t *testing.T) {
```
jonahx: Suggested improvement: `TestUnexpectedFieldsAreSavedAsOptions`

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	expectedOptions := map[string]string{
+		"foo": "5432",
+		"bar": "data",
+	}
+
+	actualConnDetails, err := NewConnectionDetails(credentials)
+	assert.Nil(t, err)
+
+	assert.EqualValues(t, expectedOptions, (*actualConnDetails).Options)
```
jonahx: `if` guard

internal/app/secretless/handlers/mysql/connection_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestConnectionDetailsAddressIsProperlyGenerated(t *testing.T) {
```
jonahx: Suggested improvement: `TestAddress`

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -25,60 +24,27 @@ type Handler struct {
...
 }
 
-// fetchConnectionDetails looks up the provider credentials and returns them
-// in a ConnectionDetails.  This almost could have been called
-// "NewConnectionDetails", but it's not quite a constructor.  I used "fetch" to
-// emphasize that this is not a pure contructor, but a side effectful one that
-// makes a network request and can return an error.  And, of course, it's a
-// method on handler.
+// fetchConnectionDetails looks up the provider credentials and returns a
+// ConnectionDetails object.
```
jonahx: Good edit

internal/app/secretless/handlers/mysql/handler.go
```diff
@@ -25,60 +24,27 @@ type Handler struct {
...
-	delete(connectionDetails, "username")
-	delete(connectionDetails, "password")
-
-	result.Options = make(map[string]string)
-	for k, v := range connectionDetails {
-		result.Options[k] = string(v)
-	}
-
-	return result, nil
+	return NewConnectionDetails(credentials)
```
jonahx: This was a good refactor.

internal/app/secretless/handlers/mysql/connection_details_test.go
sgnn7: Line 26 should take care of that sanity check

internal/app/secretless/handlers/mysql/connection_details_test.go
sgnn7: Line 47 should sanity check that

internal/app/secretless/handlers/mysql/connection_details_test.go
sgnn7: Also dealt with by the assert.Nil above it

Issue #812, opened 2019-08-05
README is clear about releases


izgeri:
Update the README to be clear about:
- The project's definition of `stable`
- The difference between `latest` and `stable` Docker images
- What’s in / when to use GitHub releases
- What’s in the project on `master` in GitHub, and when to use it

PR #813, opened 2019-08-06
Update README to add information about releases

izgeri:
This PR updates the README:
- To remove old content documenting components and refer users to the documentation website instead
- To make minor revisions to the language describing the project
- To add notes about where to find releases, and when to use which one
- To add notes about what "stability" means in the context of this project

Connected to #812

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-[![pipeline status](https://gitlab.com/cyberark/secretless-broker/badges/master/pipeline.svg)](https://gitlab.com/cyberark/secretless-broker/commits/master)
-
 # Table of Contents
 
 - [Secretless Broker](#secretless-broker)
@@ -7,9 +5,7 @@
 - [Quick Start](#quick-start)
 - [Additional demos](#run-more-secretless-demos)
 - [Using Secretless](#using-secretless)
-  - [Service Connectors](#service-connectors)
-  - [Credential Providers](#credential-providers)
-  - [Health Checks](#health-checks)
+  - [About our releases](#about-our-releases)
 - [Community](#community)
 - [Performance](#performance)
 - [Development](#development)
@@ -18,11 +14,11 @@
 
 # Secretless Broker
 
-The Secretless Broker is a connection broker which relieves client applications of the need to directly handle secrets to target services such as databases, web services, SSH connections, or any other TCP-based service.
+Secretless Broker is a connection broker which relieves client applications of the need to directly handle secrets to target services such as databases, web services, SSH connections, or any other TCP-based service.
 
 ![Secretless Broker Architecture](https://github.com/cyberark/secretless-broker/blob/master/docs/img/secretless_architecture.svg)
 
-The Secretless Broker is designed to solve two problems. The first is **loss or theft of credentials from applications and services**, which can occur by:
+Secretless is designed to solve two problems. The first is **loss or theft of credentials from applications and services**, which can occur by:
 
 - Accidental credential leakage (e.g. credential checked into source control, etc)
 - An attack on a privileged user (e.g. phishing, developer machine compromises, etc)
@@ -42,11 +38,11 @@ When the client connects to a target service through the Secretless Broker:
 
 - **The client doesn’t have to manage secret rotation**
 
-  The Secretless Broker is responsible for establishing connections to the backend, and can handle secrets rotation in a way that’s transparent to the client.
+  Secretless is responsible for establishing connections to the backend, and can handle secret rotation in a way that’s transparent to the client.
 
-To provide the Secretless Broker access to a target service, a [Service Connector](#service-connectors) implements the protocol of the service, replacing the authentication handshake. The client does not need to know or use a real password to the service. Instead, it proxies its connection to the service through a local connection to the Secretless Broker. Secretless Broker obtains credentials to the target service from a secrets vault (such as Conjur, a keychain service, text files, or other sources) via a [Credential Provider](#credential-providers). The credentials are used to establish a connection to the actual service, and the Secretless Broker then rapidly shuttles data back and forth between the client and the service.
+To provide Secretless access to a target service, a [Service Connector](#service-connectors) implements the protocol of the service, replacing the authentication handshake. The client does not need to know or use a real password to the service. Instead, it proxies its connection to the service through a local connection to Secretless. Secretless obtains credentials to the target service from a secrets vault (such as Conjur, a keychain service, text files, or other sources) via a [Credential Provider](#credential-providers). The credentials are used to establish a connection to the actual service, and Secretless then rapidly shuttles data back and forth between the client and the service.
 
-The Secretless Broker is currently licensed under [ASL 2.0](#license)
+Secretless Broker is currently licensed under [ASL 2.0](#license)
 
 ## Currently supported services
 
@@ -59,7 +55,7 @@ With many others in the planning stages!
 
 If there is a specific target service that you would like to be included in this project, please open a [GitHub issue](https://github.com/cyberark/secretless-broker/issues) with your request.
 
-For specific guidelines about using a particular service, please see our instructions for [using the Secretless Broker](#using-secretless).
+For specific guidelines about using a particular service, please see our [documentation](https://docs.secretless.io/Latest/en/Content/References/connectors/overview.htm).
 
 # Quick Start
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+Stable components (eg service connectors, credential providers, etc.) of Secretless meet the core acceptance criteria:
+- Synchronous user actions must complete in a reasonable amount of time
+- The component must guard against threats from all parts of the [STRIDE threat model](https://en.wikipedia.org/wiki/STRIDE_(security))
+- Documentation exists that clearly explains how to set up and use the component as well as providing troubleshooting information for anticipated common failure cases.
+- A suite of automated tests that exercise the component exists, and code coverage has been measured. The entire team is in mutual agreement that any code not tested by the unit test suite is acceptable to leave untested. Tests that are deferred to be implemented at a later time are tracked so that we don't lose visibility into that work.
+  
+In addition, the following must be true for any stable release:
+- Secretless has had security review, and has no high or critical bugs
+  - Secretless has been scanned for CVEs and security vulnerabilities with no high issues
+  - Secretless has undergone threat modeling
```
sgnn7: Do we even have a clear idea what this entails? If not, we should remove it.

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+### Stable release definition
+Stable components (eg service connectors, credential providers, etc.) of Secretless meet the core acceptance criteria:
+- Synchronous user actions must complete in a reasonable amount of time
+- The component must guard against threats from all parts of the [STRIDE threat model](https://en.wikipedia.org/wiki/STRIDE_(security))
+- Documentation exists that clearly explains how to set up and use the component as well as providing troubleshooting information for anticipated common failure cases.
+- A suite of automated tests that exercise the component exists, and code coverage has been measured. The entire team is in mutual agreement that any code not tested by the unit test suite is acceptable to leave untested. Tests that are deferred to be implemented at a later time are tracked so that we don't lose visibility into that work.
+  
+In addition, the following must be true for any stable release:
+- Secretless has had security review, and has no high or critical bugs
+  - Secretless has been scanned for CVEs and security vulnerabilities with no high issues
```
sgnn7: We should specify which tools we will use for scanning

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+  - Secretless has undergone threat modeling
+-	Secretless has a stable API and configuration spec and there will be no near-term backwards-incompatible changes
```
sgnn7: This should be covered by following SemVer rules so we don't need to explicitly state it.

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+- For use cases involving stable components of Secretless (eg service connectors, credential providers, etc.):
+  -	Secretless is stable while running
+    -	It does not randomly drop connections while running
```
sgnn7:
```suggestion
    -	It does not drop connections while running
```

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+    -	It can run without failure for an extended period of time (~4 days)
+    -	It can consistently open connections while running
```
sgnn7: This is covered by the item above

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+    -	When it fails in Kubernetes / OpenShift, the pod is destroyed and rescheduled
```
sgnn7: define "fails"

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+  -	Secretless has minimal impact on connection speeds
+    -	Request throughput is within 20% of direct-to-DB speed
+  -	Secretless performs under load in a realistic environment
```
sgnn7: We should specify what we consider a realistic environment

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+  -	Secretless handles connections transparently, so that requests from the client are appropriately transmitted to the server and messages from the server are propagated back to the client
+-	Secretless is easy to set up, and setup takes less than 30 minutes
```
sgnn7: This needs a broader definition on what that setup is

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
-HTTP/1.1 200 OK
-Content-Type: application/json; charset=utf-8
-Date: Thu, 27 Dec 2018 17:13:22 GMT
-Content-Length: 45
-
-{
-    "listening": "OK",
-    "ready": "OK"
-}
-```
```
sgnn7: I don't agree that we should be removing this content. While we do have that content on our documentation website, it will probably prevent many developers from adopting our product since you can't just take a look at the README itself to get an idea on how to use/run this thing.

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+
+### GitHub releases
+Post-1.0, GitHub releases are created for GitHub tagged versions that have undergone additional quality activities to ensure that Secretless continues to meet its [stability requirements](#stable-release-definition). When a GitHub release is created for a stable Secretless version, an additional Docker image is pushed to DockerHub with the `stable` tag.
+
+### GitHub repository
+The code on `master` in the project's GitHub repository represents the development work done since the previous GitHub tag. It is possible to build Secretless from source (see [our contributing guidelines](#development) for more info), but for regular use we recommend using the `stable` Docker image from DockerHub or an official GitHub release.
+
+### Stable release definition
+Stable components (eg service connectors, credential providers, etc.) of Secretless meet the core acceptance criteria:
+- Synchronous user actions must complete in a reasonable amount of time
```
sgnn7: What are the synchronous user actions here? We do most of our work on startup so I'm a bit confused as to what's meant here.

README.md
izgeri: maybe this should be dropped, or a reference to the STRIDE model should also be added here.

README.md
izgeri: You are right, but it's also explicitly stating an aspect of "run without failure" that is beneficial to specify, imo

Maybe it should be "It can run without failure (eg, it can consistently open connections and maintain existing open connections) ..."

README.md
izgeri: the process dies

README.md
izgeri: hm - any thoughts for how to define that? points for simplicity

README.md
izgeri: "and [adding the configured secretless broker sidecar to your application]"?

README.md
izgeri: our README is so long, even with this removed. looking at other repos, eg [kubernetes](https://github.com/kubernetes/kubernetes/), [envoy](https://github.com/envoyproxy/envoy/), [linkerd](https://github.com/linkerd/linkerd2), there has been a movement in this space to make the readme as simple as possible and refer out to other docs. there are also major benefits in terms of keeping things up-to-date of having a single source-of-truth for any given piece of data.

I just don't think there's a real benefit from having this info in the README anymore.

README.md
izgeri: maybe @andytinkham can comment / give some context here? this is pulled straight from the core acceptance criteria that he drafted.

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
-    "listening": "OK",
-    "ready": "OK"
-}
-```
+## About our releases
+
+### Docker images
+The primary source of Secretless Broker releases is our [DockerHub](https://hub.docker.com/r/cyberark/secretless-broker/tags/).
+
+Every new GitHub tag (`x.y.z`) added to the project produces a set of Docker images in DockerHub: (`x.y.z`, `x.y`, `x`, `latest`). The `latest` image tag therefore always corresponds to the latest GitHub tag.
```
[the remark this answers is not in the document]
izgeri: you are correct. so I could add to the last sentence:
> and the `x.y` and `x` tags alway correspond to the latest `x.y.z` image.

README.md
```diff
@@ -136,267 +132,43 @@ For an even more in-depth demo, check out our [Deploying to Kubernetes](https://
...
+
+### GitHub releases
+Post-1.0, GitHub releases are created for GitHub tagged versions that have undergone additional quality activities to ensure that Secretless continues to meet its [stability requirements](#stable-release-definition). When a GitHub release is created for a stable Secretless version, an additional Docker image is pushed to DockerHub with the `stable` tag.
```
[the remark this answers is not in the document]
izgeri: the idea is that there is one `stable` tag, that always corresponds to the latest stable release.

every git tag produces a `latest` image
only git tags that have undergone additional quality testing will produce a `stable` image

I am open to suggestions for other regimes, where we might append `-stable` to a git tag when that specific tag has undergone more stability testing, but right now what I've suggested seems the most simple. you can always find historically stable release versions by looking at the github releases for versions over 1.0 - if there is no release, the tag was not promoted to stable.

README.md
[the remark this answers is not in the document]
izgeri: what if we logged all low / medium issues in the backlog? then there would be visibility to see what remained unaddressed.

README.md
andytinkham: The "synchronous" word specifically, I think was added at @alexkalish's suggestion because we may not be able to control the timing of anything happening asynchronously. In general, the intent of this core criteria right now is "think about performance and don't make things be slow". As we discuss and refine, this should become something more measurable and less hand-wavey. For this page, I'd be OK dropping 'synchronous' though I don't know if there's anything to do yet around "reasonable".

README.md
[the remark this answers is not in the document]
izgeri: maybe this implies that for now we should leave it as 
> Secretless has had security review (including static code analysis), and all known high and critical issues have been addressed. Any low or medium issues that have not been addressed have been logged in the [github issue backlog](https://github.com/cyberark/secretless-broker/issues) with a label of the form `security/X`

README.md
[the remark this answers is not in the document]
andytinkham: The addition is moving in the right direction. I think one of the things I stumbled on leading to the comment was that `x` and `x.y` may very well not point to the same release except for when `y` is latest-y.  I think the phrasing above still implies they point to the same image. Maybe something like:

The `latest` image tag therefore always corresponds to the latest GitHub tag, while `x` and `x.y.z` each point to the latest matching release.

Still doesn't feel exactly right to me, and it's a bit splitting hairs, but that's where my thoughts are going.

README.md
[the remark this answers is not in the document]
andytinkham: OK, so to make sure that I got it. In DockerHub, there'll be a number of tags. One `latest`, one `stable` and all the major version/minor version/full version tags (x, x.y, and x.y.z). `latest` and `stable` will sometimes point to the same release, but there'll be windows while a version is released but we're still doing the stability testing where `latest` will be a newer version. 

In addition, every time we push a new version to the `stable` tag, we'll make a GH release that's tagged numerically. Any GitHub release is considered stable, so if I need an old one for some reason, I just come here and get a stable release. 

That makes sense to me. I'm good with that approach.

README.md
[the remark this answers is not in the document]
andytinkham: I think that would help, particularly if we logged each individually. We can close issues we don't intend to fix and leave the pending ones open. Then, we have the path of users being able to comment on a particular issue if it is blocking them somehow or they otherwise disagree. 

I'd like to see us get to the point where even these issues don't stay on the backlog for long, but this is a good way to be transparent about what still remains.

README.md
sgnn7: stride model reference would work - I'm just trying to avoid us having flexible/fungible standards for these if we're writing them down

README.md
sgnn7: any wording that will give precise meaning of what the failure mode is would be helpful here but in theory both lines are saying the same thing to me though 🤷‍♂ :
- broker doesn't fail
- broker works

README.md
sgnn7: I figured as much but it should be in the wording of the text

README.md
sgnn7: k8s environment as a sidecar with DAP?

README.md
sgnn7: LGTM!

README.md
alexkalish: This AC feels a little out of place in the list, especially since this product doesn't have any synchronous user actions, in and of itself.  I think that the 20% max performance penalty added farther below captures the spirit of this criteria.

izgeri: ok, so it sounds like y'all would be good with me dropping this one?

README.md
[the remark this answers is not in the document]
izgeri: and just to further clarify, I'm distinguishing between github tags and github releases. I can create a tag `x.y.z` without creating a release for that tag.

README.md
sgnn7: yup - seems like it

andytinkham: I'm ok with removing it.

Issue #542, opened 2018-11-30
Automate demo testing prior to PR merge


doodlesbykumbi:
Update `k8s-ci` to run through the `demos/k8s-demo` scripts to ensure they don't error as part of the CI pipeline.

PR #820, opened 2019-08-13
542 Automate k8s demo

sgnn7:

Runs k8s-demo on each Jenkins build

Connected to #542 

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/542-automate-k8s-demo/)

### DIFF
--- a/demos/k8s-demo/02_app_developer_steps
+++ b/demos/k8s-demo/02_app_developer_steps
@@ -16,8 +16,53 @@ step "Start application"
 kubectl --namespace "${APP_NAMESPACE}" \
   apply --filename "etc/quick-start-application.yml"
 
-echo "Detecting app url..."
-APPLICATION_URL="$(minikube service quick-start-application --namespace "${APP_NAMESPACE}" --wait 15 --url)"
+if [ ! "${SECRETLESS_IMAGE_NAME:-}" == "" ]; then
+  image_name="${SECRETLESS_IMAGE_NAME}:${SECRETLESS_IMAGE_TAG}"
+
+  step "Patching deployment with ${image_name}"
+
+  kubectl patch deployment "quick-start-application" \
+    --namespace="${APP_NAMESPACE}" \
+    --patch """
+  spec:
+    template:
+      spec:
+        containers:
+          - name: secretless
+            image: ${image_name}
+"""
+fi
+
+step "Making service externally visible"
+
+if is_minikube; then
+  APPLICATION_URL="$(minikube service quick-start-application --namespace "${APP_NAMESPACE}" --wait 15 --url)"
+else
+  readonly external_service_name="${APP_NAMESPACE}-external-svc"
+  kubectl expose deployment quick-start-application \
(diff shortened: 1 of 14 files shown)


demos/k8s-demo/02_app_developer_steps
```diff
@@ -16,8 +16,53 @@ step "Start application"
...
+      spec:
+        containers:
+          - name: secretless
+            image: ${image_name}
+"""
+fi
+
+step "Making service externally visible"
+
+if is_minikube; then
```
doodlesbykumbi: After some thought, everything under this line isn't necessary. We shouldn't be bothered whether it's minikube or not. For the purposes of this demo we don't care to expose the application externally. 

Without any loss in generality, we could move the curl calls below to `kubectl run -it ...` and use the `quick-start-application` service DNS name `quick-start-application.${APP_NAMESPACE}.svc.local`.

sgnn7: While you're right that we can test/use the code in that way and I considered it, I opted to leave it as-is since it would miss the the point of being an advanced example of usage that's close to the actual deployment strategy.

demos/k8s-demo/utils.sh
```diff
@@ -33,3 +33,7 @@ wait_for_app() {
     echo "OK"
   fi
 }
+
+is_minikube() {
```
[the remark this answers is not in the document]
sgnn7: It seems stable enough from the docs: https://kubernetes.io/docs/setup/learning-environment/minikube/#starting-a-cluster. Thoughts?

demos/k8s-demo/02_app_developer_steps
doodlesbykumbi: I don't think there's any added value to making the application publicly accessible. It does nothing for the point we're trying to put across with the demo.

There are too many variables that go into how an application can be made publicly available.  Instead of the demo breaking for some why not circumvent that altogether. There's no loss in generality in not making the application publicly available.  If anything we make things less complicated for ourselves by removing the need to figure out if the use on minikube or not. 

I think we should overhaul this.

demos/k8s-demo/utils.sh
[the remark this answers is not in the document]
doodlesbykumbi: Context names are mutable so this call wouldn't be conclusive. I'd rather we avoid this altogether though.

demos/k8s-demo/02_app_developer_steps
sgnn7: Fair enough. I don't know if I fully agree about the added value but I'm not too opinionated on it. I'll make the changes.

demos/k8s-demo/utils.sh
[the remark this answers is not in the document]
sgnn7: With the convo from below, this might be removable - I'll see how that shakes out

sgnn7:
@doodlesbykumbi Comments fixed and build is green

PR #824, opened 2019-08-19
cross-compile linux and darwin using cross-compile goreleaser image

doodlesbykumbi:
An image of goreleaser build on top of dockercore/golang-cross allows us to cross-compile binaries, that use CGO, to both linux and darwin.

### DIFF
--- a/bin/Dockerfile.releaser
+++ b/bin/Dockerfile.releaser
@@ -1,0 +1,23 @@
+FROM dockercore/golang-cross
+
+RUN apt-get update && \
+    apt-get install -y bash \
+                       build-essential \
+                       curl \
+                       docker \
+                       git \
+                       mercurial \
+                       rpm
+
+ENV GORELEASER_VERSION=0.116.0 \
+    GORELEASER_SHA=34b7e3b843158bd0714d1be996951685496491adab4524fb1198ae144ab06ba3 \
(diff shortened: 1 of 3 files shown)


bin/Dockerfile.releaser
```diff
@@ -0,0 +1,23 @@
...
+ENV GORELEASER_VERSION=0.116.0
+ENV GORELEASER_SHA=34b7e3b843158bd0714d1be996951685496491adab4524fb1198ae144ab06ba3
+
+ENV GORELEASER_DOWNLOAD_FILE=goreleaser_Linux_x86_64.tar.gz
+ENV GORELEASER_DOWNLOAD_URL=https://github.com/goreleaser/goreleaser/releases/download/v${GORELEASER_VERSION}/${GORELEASER_DOWNLOAD_FILE}
+
+RUN  wget ${GORELEASER_DOWNLOAD_URL}; \
+			echo "$GORELEASER_SHA $GORELEASER_DOWNLOAD_FILE" | sha256sum -c - || exit 1; \
+			tar -xzf $GORELEASER_DOWNLOAD_FILE -C /usr/bin/ goreleaser; \
+			rm $GORELEASER_DOWNLOAD_FILE;
```
sgnn7: I think that overuse of `;` here will result in broken images vs `&&`. Is there a way to get this arranged so that they only chain on success?

PS: I know it's a bit tricky given the `exit` but it's better to have an error here than later in the build process.

Issue #826, opened 2019-08-19
Configuration adds support for universal `connector` spec


izgeri:
The v2 config currently requires that the specific connector protocol that's used is specified in the `protocol` field unless you are using an `http` connector. In that case the `protocol` is specified as `http` and there is a `config.authenticationStrategy` field where the specific protocol implementation is referenced.

```
version: "2"
services:
  aws-proxy:
    protocol: http
    listenOn: tcp://0.0.0.0:8080
    credentials:
      accessKeyId:
        from: env
        get: AWS_ACCESS_KEY_ID
      secretAccessKey:
        from: env
        get: AWS_SECRET_ACCESS_KEY
    config:
      authenticationStrategy: aws
      authenticateURLsMatching:
        - .*
```

This is inconsistent and could be improved. In this effort, we will add support for a new field `connector` that accepts a protocol identifier for _any_ connector type, even a specific instance of an `http` connector. With these changes, the config above would become:

```
version: "2"
services:
  aws-proxy:
    connector: aws
    listenOn: tcp://0.0.0.0:8080
    credentials:
      accessKeyId:
        from: env
        get: AWS_ACCESS_KEY_ID
      secretAccessKey:
        from: env
        get: AWS_SECRET_ACCESS_KEY
    config:
      authenticateURLsMatching:
        - .*
```

AC:
- [x] Config v2 is updated to support specifying `connector` instead of `protocol` or `protocol`/`config.authenticationStrategy`, while maintaining support for the old `protocol` usage
- [x] References to `protocol` and/or `authenticationStrategy` in tests and sample configuration files in the main project repo are updated to use the single `connector` reference instead
- [x] An issue is filed in the docs repo for updating the YAML examples / documentation to use `connector` instead of `protocol` https://github.com/cyberark/secretless-docs/issues/185

izgeri:
Note: we will probably have to consider what happens if both `protocol` and `connector` are specified. My instinct is that `connector` should override, perhaps with a warning. Wdyt? This case should also be tested.

PR #829, opened 2019-08-20
Configuration adds support for universal connector spec

doodlesbykumbi:
1. Configuration adds support for universal `connector` spec.
1. Maintains backwards compatibility with `protocol` spec and http `authenticationStrategy`.
1. Adds unit test cases for `connector` spec.
1. References to protocol and/or authenticationStrategy in tests and sample configuration files in the main project repo are updated to use the single connector reference instead

Connected to https://github.com/cyberark/secretless-broker/issues/826

### DIFF
--- a/pkg/secretless/config/v2/config.go
+++ b/pkg/secretless/config/v2/config.go
@@ -1,8 +1,11 @@
 package v2
 
 import (
+	"fmt"
+	"log"
 	"sort"
 
+	validation "github.com/go-ozzo/ozzo-validation"
 	"gopkg.in/yaml.v2"
 
 	config_v1 "github.com/cyberark/secretless-broker/pkg/secretless/config/v1"
@@ -19,11 +22,11 @@ type Config struct {
 // location of its required credentials, and (optionally) any additional
 // protocol specific configuration.
 type Service struct {
-	Credentials    []*Credential
-	ListenOn       string
-	Name           string
-	Protocol       string
-	ProtocolConfig []byte
+	Connector       string
+	ConnectorConfig []byte
+	Credentials     []*Credential
+	ListenOn        string
+	Name            string
 }
 
 // NewV1Config converts the bytes of a v2 YAML file to a v1.Config.  As such,
@@ -88,9 +91,42 @@ func NewConfig(v2YAML []byte) (*Config, error) {
 		services = append(services, svc)
 	}
 
+	// sort Services
+	sort.Slice(services, func(i, j int) bool {
+		return services[i].Name < services[j].Name
+	})
+
 	return &Config{
 		Services: services,
 	}, nil
+}
+
+// connectorFromLegacyHTTPConfig extracts authenticationStrategy.
+// This function is useful when the deprecated 'protocol' field equals 'http'
+// and you want to determine the connector name
+func connectorFromLegacyHTTPConfig(connectorConfigBytes []byte) (string, error) {
+	tempCfg := &struct {
+		AuthenticationStrategy string `yaml:"authenticationStrategy"`
+	}{}
+	err := yaml.Unmarshal(connectorConfigBytes, tempCfg)
+	if err != nil {
+		return "", err
+	}
+
+	err = validation.ValidateStruct(
+		tempCfg,
+		validation.Field(
+			&tempCfg.AuthenticationStrategy,
+			validation.Required,
+			validation.In(HTTPAuthenticationStrategies...),
+		),
+	)
+
+	if err != nil {
+		return "", err
+	}
+
+	return tempCfg.AuthenticationStrategy, nil
 }
 
 // NewService creates a named v2.Service from yaml bytes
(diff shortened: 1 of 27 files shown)


pkg/secretless/config/v2/config_http_test.go
```diff
@@ -0,0 +1,70 @@
+package v2
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
```
sgnn7: import sorting needed

pkg/secretless/config/v2/config_http_test.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func TestNewHTTPConfig(t *testing.T) {
+	t.Run("valid config with 'authenticateURLsMatching' string", func(t *testing.T) {
```
sgnn7: The test names in this file don't really explain what you're testing (is `valid config with 'authenticateURLsMatching' string` good? is it bad? does it return an error?)

pkg/secretless/config/v2/config_http_test.go
```diff
@@ -0,0 +1,70 @@
...
+	t.Run("config with bad type for 'authenticateURLsMatching'", func(t *testing.T) {
+		configFileContents := []byte(
+			`
+authenticationStrategy: aws
+authenticateURLsMatching: 
+ - true
+ - "meow"
+`)
+		_, err := newHTTPConfig(configFileContents)
+		assert.Error(t, err)
```
sgnn7: Not a blocker but probably the error type/message should be asserted too.

pkg/secretless/config/v2/v1_service.go
```diff
@@ -2,6 +2,8 @@ package v2
 
 import (
 	"fmt"
+	validation "github.com/go-ozzo/ozzo-validation"
+	"gopkg.in/yaml.v2"
 	"sort"
 	"strings"
```
sgnn7: Import list sorting needed

pkg/secretless/config/v2/config_http.go
```diff
@@ -2,24 +2,31 @@ package v2
...
 
 type httpConfig struct {
-	AuthenticationStrategy   string   `yaml:"authenticationStrategy"`
 	AuthenticateURLsMatching []string `yaml:"authenticateURLsMatching"`
 }
 
 // HTTPAuthenticationStrategies are the different ways an http service
 // can authenticate.
-var HTTPAuthenticationStrategies = []string{
+var HTTPAuthenticationStrategies = []interface{}{
```
sgnn7: This list should be dynamically generated from plugin list that has `http/` as prefix for now

pkg/secretless/config/v2/v1_service.go
```diff
@@ -71,35 +78,60 @@ func newV1Service(v2Svc Service) (ret *v1Service, err error) {
...
+			return err
+		}
 
-	v1Svc.Handler.Match = httpCfg.AuthenticateURLsMatching
-	v1Svc.Handler.Type = httpCfg.AuthenticationStrategy
+		v1Svc.Handler.Match = httpCfg.AuthenticateURLsMatching
+		v1Svc.Handler.Type = httpHandlerType
 
+		return nil
+	}
```
sgnn7: Not a blocker but this method is a bit gnarly though I don't have any specific guidance on how to improve it :/

pkg/secretless/config/v2/config.go
```diff
@@ -88,6 +89,11 @@ func NewConfig(v2YAML []byte) (*Config, error) {
 		services = append(services, svc)
 	}
 
+	// sort Services
+	sort.Slice(services, func(i, j int) bool {
+		return services[i].Name < services[j].Name
```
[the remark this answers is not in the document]
jonahx: @sgnn7 on this one, @doodlesbykumbi is following the convention.  if you check out the example docs in the `sort` package you'll see:

```
	sort.Slice(people, func(i, j int) bool {
		return people[i].Age > people[j].Age
	})
```

pkg/secretless/config/v2/config_http_test.go
jonahx: We should definitely avoid asserting on message string text.  I am personally okay with asserting on a custom error type, though this is a topic of controversy in the community, and Dave Cheney recommends against even this (at least in most cases):

pkg/secretless/config/v2/v1_service.go
jonahx: @doodlesbykumbi You can set your IDE to automatically do this for you.

pkg/secretless/config/v2/config_http.go
jonahx: We'd have to append our internal ones to that list, correct?

pkg/secretless/config/v2/v1_service.go
jonahx: One small thing: I've read now in multiple Go best practice guides that you can (almost always) rewrite logic to remove `else` clauses from your logic, and I agree that this improves readability.

pkg/secretless/config/v2/config.go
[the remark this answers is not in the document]
sgnn7: Convention of single-letter variables aside, using `i, j` is pretty uninformative of what you're working with other than integers. Definitely not a blocker but I don't think the convention is advantageous here vs `i1`/`i2` even.

pkg/secretless/config/v2/config_http.go
doodlesbykumbi: Yup. I'm going to defer that to another PR. There's a broader issue where we need to have validation happen in v2 instead of delegating it to v1. We also need to have this list of http service connectors but in a way that doesn't coupled the v1 way of doing things. A lot of this should be resolved as part of the plugin refactor.

pkg/secretless/config/v2/config_http_test.go
sgnn7: Error type should be for sure asserted on even if it's not custom but I agree that error message assertion might be too much of the detail to worry about.

jonahx:
> Some of this is a bit over my head

Which parts?  We should probably make it clearer / add comments / etc if this is the case.

pkg/secretless/config/v2/config_http.go
```diff
@@ -36,49 +43,45 @@ func newHTTPConfig(cfgBytes []byte) (*httpConfig, error) {
...
-	// It's a string, let's convert it to a []string
-	cfg.AuthenticationStrategy = tempCfg.AuthenticationStrategy
-	cfg.AuthenticateURLsMatching = []string{ tempCfg.AuthenticateURLsMatching }
+	// Populate actual http config from tempCfg
+	switch v := tempCfg.AuthenticateURLsMatching.(type) {
+	case string:
+		cfg.AuthenticateURLsMatching = []string{ v }
+	case []interface{}:
+		urlMatchStrings := make([]string, len(v))
+		for urlMatchIndex, urlMatch := range v {
```
jonahx: I'd change `urlMatchIndex` -> `i`.  No need to go against convention unless you have a really long loop or something.

pkg/secretless/config/v2/config_test.go
```diff
@@ -42,37 +87,54 @@ func TestNewConfig(t *testing.T) {
 		assert.Error(t, err)
 	})
 
-	t.Run("basic hydration", func(t *testing.T) {
-		cfg, err := sampleConfig()
+	RunNewConfigTestCases(t, "with-protocol", sampleConfigWithProtocolStr)
+	RunNewConfigTestCases(t, "with-connector", sampleConfigWithConnectorStr)
+}
```
jonahx: I like how you did this.

pkg/secretless/config/v2/config.go
```diff
@@ -88,31 +91,82 @@ func NewConfig(v2YAML []byte) (*Config, error) {
...
+		// When the deprecated 'protocol' field equals 'http'
+		// the connector name must be extracted from the http config
+		if connector == "http" {
+			connector, err = connectorFromLegacyHTTPConfig(connectorConfigBytes)
+			if err != nil {
+				return nil, fmt.Errorf("error on http config for service '%s': %s", svcName, err)
+			}
+		}
+	} else if svcYAML.Protocol != "" {
+		log.Printf("WARN: 'connector' and 'protocol' keys found on service '%s'. 'connector' key takes precendence, 'protocol' is deprecated.", svcName)
```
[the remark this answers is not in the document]
doodlesbykumbi: Thanks. I extrapolated from your example. Flattening conditional logic is useful.

pkg/secretless/config/v2/config_http.go
```diff
@@ -2,24 +2,31 @@ package v2
 
 import (
 	"errors"
-
```
sgnn7: keep the removed LF

sgnn7:
> Which parts? We should probably make it clearer / add comments / etc if this is the case.

The new code @doodlesbykumbi pushed is easier to follow but overall I think we have a serious root issue that our configs are growing to be these big context-specific/logic-heavy objects (~700 lines of code at the moment to parse the configs) which is going to make them very hard to maintain and evolve, especially in the CRD world. We need to find a strategy to reduce this down before we add any more changes after this one.

Issue #852, opened 2019-08-28
Plugin and Connector interfaces exist


jonahx:
## Description

See the [PLUGIN README](https://gist.github.com/jonahx/6892ddeaca03f6a5c281eb38218687eb) for exact definitions of these interfaces.  Note, however, that minor changes will likely be required per the notes in item #846.

AC:
- [ ] `tcp.Plugin` and `tcp.Connecter`  interfaces exists in `secretless.plugin.connector.tcp`
- [ ] `http.Plugin` and `http.Connector` interfaces exist in `secretless.plugin.connector.http`

PR #880, opened 2019-09-06
Added http and tcp plugin interfaces

sgnn7:
Since we have a new way to interact with our plugins, the
interface is the first needed step to initiate the conversion
so this commits adds those for both http and tcp.

Connected to #852 

Diff page

### DIFF
--- a/pkg/secretless/plugin/connector/http/http.go
+++ b/pkg/secretless/plugin/connector/http/http.go
@@ -1,0 +1,25 @@
+package http
+
+import (
+	"net/http"
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+)
+
+// Plugin is the main interface that HTTP plugins need to implement
+// to be loaded by our codebase.
+type Plugin interface {
+	// PluginInfo contains a key-value map of informational fields
+	// about the plugin.
+	PluginInfo() map[string]string
+
+	// NewConnector creates a new Connector based on the ConnectorResources
+	// passed into it.
+	NewConnector(connector.Resources) Connector
+}
+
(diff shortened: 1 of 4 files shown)


pkg/secretless/plugin/connector/http/http.go
```diff
@@ -0,0 +1,24 @@
...
+
+	// NewConnector creates a new Connector based on the ConnectorResources
+	// passed into it.
+	NewConnector(connector.Resources) Connector
+}
+
+// Connector is the function that will be invoked when a matching
+// request comes in. It uses both the request object and the credentials
+// map to authenticate the client.
+type Connector func(request *http.Request, credentials map[string][]byte) error
```
jonahx: I'm thinking we ought to create a meaningfully named type for `map[string][]byte`.  Something like:
```go
type SecretsByName map[string][]byte
// OR
type RawSecrets map[string][]byte
```
This won't affect usage at all.  See:

https://play.golang.org/p/0AdmgxVweOO

pkg/secretless/plugin/connector/tcp/tcp.go
```diff
@@ -0,0 +1,26 @@
...
+	// passed into it.
+	NewConnector(connector.Resources) Connector
+}
+
+// Connector is the function that will be invoked when a matching
+// TCP request comes in. It uses both the initiating connection and the
+// credentials map to authenticate the client, returning the backend
+// network connection.
+type Connector func(clientConn net.Conn,
+	credentials map[string][]byte) (backendConn net.Conn, err error)
```
jonahx: Nitpicky, but I feel like once you wrap a function sig, it's cleaner to just go all the way:
```go
type Connector func(
    clientConn net.Conn,
    credentials map[string][]byte
) (backendConn net.Conn, err error)
```
wdyt?

Issue #840, opened 2019-08-27
Create ConnectorResources Interface


jonahx:
## Description

AC:
- [ ] Public package `secretless/plugin/connector` is created, if it doesn't already exist.
- [ ] `ConnectorResources` interface is defined in that package
- [ ] Has contructor `NewConnectorResources`

`NewConnectorResources` will probably look like:

```go
`NewConnectorResources(
  fetchCreds func() (map[string][]byte, error),
  cfgBytes []byte, 
  logger secretless.Logger
)
```

PR #882, opened 2019-09-09
Added a method to create a new connector Resources interface

sgnn7:
We now have a way to instantiate a connector Resources interface via a
backing object. This should allow us to easily invoke the NewConnector
methods on plugins.

Connected to #840

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/840-connector-resources-int/)

N/A

### DIFF
--- a/pkg/secretless/plugin/connector/resources.go
+++ b/pkg/secretless/plugin/connector/resources.go
@@ -15,3 +15,28 @@ type Resources interface {
 	// to record logging messages from the plugin.
 	Logger() log.Logger
 }
+
+type _resources struct {
+	config []byte
+	logger log.Logger
+}
+
+func (res *_resources) Config() []byte {
+	return res.config
+}
+
+func (res *_resources) Logger() log.Logger {
+	return res.logger
+}
+
+// NewResources creates a new Resources interface from a backing object struct
+// to be used in invocation of a new Connector
+func NewResources(
+	config []byte,
+	logger log.Logger) Resources {
+
+	return &_resources{
+		config: config,
+		logger: logger,
+	}
+}
--- a/pkg/secretless/plugin/connector/resources_test.go
+++ b/pkg/secretless/plugin/connector/resources_test.go
@@ -1,0 +1,22 @@
+package connector
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


pkg/secretless/plugin/connector/resources.go
```diff
@@ -0,0 +1,42 @@
...
+	// Config returns the content of addition configuration
+	// parameters passed to the connector.
+	Config() []byte
+
+	// Logger returns an instance of a Logger that can be used
+	// to record logging messages from the plugin.
+	Logger() log.Logger
+}
+
+type resourcesHolder struct {
```
izgeri: is the stuff from here down meant to be here?

sgnn7: Yeah. We cannot return an interface without a backing object that implements that interface, thus we have the `resourcesHolder`

jonahx: Just call this `resources` or if you're worried about confusion with `Resources`, append or prepend an `_` too.

pkg/secretless/plugin/connector/resources_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/cyberark/secretless-broker/internal/pkg/log"
+)
+
+func TestNewReourcesWrapper(t *testing.T) {
```
jonahx: typo.  also, i'd remove `Wrapper`.  It's a test of `NewResources`, so `TestNewResources` is exactly right.  You can to the test suite subtests with `t.Run` per normal if you need to add more.

pkg/secretless/plugin/connector/resources.go
```diff
@@ -15,3 +15,28 @@ type Resources interface {
...
+
+func (res *_resources) Config() []byte {
+	return res.config
+}
+
+func (res *_resources) Logger() log.Logger {
+	return res.logger
+}
+
+// NewResources creates a new Resources interface from a backing object stuct
```
izgeri:
```suggestion
// NewResources creates a new Resources interface from a backing object struct
```

PR #885, opened 2019-09-10
add ProxyService interface

doodlesbykumbi:
Adds ProxyService interface
Check the diff

### DIFF
--- a/internal/app/secretless/proxy_service.go
+++ b/internal/app/secretless/proxy_service.go
@@ -1,0 +1,17 @@
+package secretless
+
+// ProxyService is the interface the Secretless framework uses to start and stop
+// ProxyServices. This interface allows ProxyServices to be implemented
+// independent of each other. While operational, ProxyServices are expected to
+// serve connections by routing them to the appropriate service connector
+type ProxyService interface {
+	// Start is a synchronous method responsible for carrying out the steps
+	// necessary for a ProxyService to become ready to service client
+	// connections.
+	Start() error
+	// Stop is a synchronous method responsible for terminating a ProxyService.
+	// Stop is expected to both terminate the ProxyService and carry out any
+	// necessary clean-up of resources held and consumed by the ProxyService
+	// while operating.
+	Stop() error
+}


internal/app/secretless/proxy_service.go
```diff
@@ -0,0 +1,18 @@
+package secretless
+
+// ProxyService is the interface the Secretless framework uses to
+// start and stop ProxyServices. 
+// This interface allows ProxyServices to be implemented
+// independent of each other.
+// While operational, ProxyServices are expected to serve client connections
+// by routing them to the appropriate service connector
```
sgnn7: Nit: You should wrap only on line length and not sentence starts since it makes the reading easier (paragraph form)

internal/app/secretless/proxy_service.go
```diff
@@ -0,0 +1,18 @@
...
+type ProxyService interface {
+	// Start is a synchronous method responsible for carrying out the steps
+	// necessary for a ProxyService to become ready to service client connections.
+	Start() error
+	// Stop is a synchronous method responsible for terminating a ProxyService.
+	// Stop is expected to both terminate the ProxyService and
+	// carry out any necessary clean-up of resources held and consumed by the ProxyService
+	// while operating.
```
sgnn7: Nit: Related to the previous comment, use a consistent word wrap boundary, especially for the same comment over multiple lines..

internal/app/secretless/proxy_service.go
```diff
@@ -0,0 +1,18 @@
...
+	Stop() error
```
sgnn7: So I'm not sure about this but: 
We know that `Stop` may take a while to finish. To use it properly in async manner you need to wrap it in some sort of go subroutine that you would have a wait group or channel around to keep track of when to return to other work, With all that said, should we return not just an error but also either a `WaitGroup` or a `chan bool` too to indicate when the operation is complete?

doodlesbykumbi: Sync for now. You're absolutely right about async.

Issue #879, opened 2019-09-05
Assorted small plugin.Manager responsibilities exist as independent types for StartSecretless


jonahx:
The idea here is to break the high-level pieces of Secretless into de-coupled, independnet types.  This list will likely include:

AC:
- [ ] Handling signals
- [ ] Handling profiling
- [ ] Version checks

Perhaps one or two more thing that will be uncovered in #877.   We may also decide that some of these responsibilities don't deserve their own types, but this story exists to capture the likely scenario that some will.

jonahx:
This is also closed by my PR:

https://github.com/cyberark/secretless-broker/pull/889

Issue #877, opened 2019-09-05
StartSecretless Is Implemented With Stubbed Collaborators


jonahx:
This will define, in code, exactly what `StartSecretless` does.  The Run() method be complete, even though the collaborators it's using won't be implemented for real yet.

The primary goal here is to make the high-level logic of Secretless completely explicit.

AC:
- [ ] Create stubs for collaborator objects
- [ ] Run method complete
- [ ] Unit tests or Run(), including error conditions, using doubles for collaborators

jonahx:
NOTE: This PR:

https://github.com/cyberark/secretless-broker/pull/889

address this ticket as well.

Issue #876, opened 2019-09-05
StartSecretless Interface exists


jonahx:
This is the interface for the top level method that runs secretless.  It will either replace or be called by what is currently `entrypoint`.  Much of the logic currently in `plugin.Manager` will be moved here, and `plugin.Manager` will go away. 

The interface itself will likely have only a single `Run()` method.  However, "designing the interface" here means listing precisely what collaborators will be given to it on construction, and what it's responsibilities are.

AC:
- [ ] Determine the collaborators it will need.
- [ ] Create the `NewStartSecretless` method that will accept those collaborators.
- [ ] Document what `Run()` will do.

izgeri:
Note: in this issue the logic around reloading Secretless plugins when a new configuration is detected may be removed, but the logic for detecting config changes (eg [fs watcher](https://github.com/cyberark/secretless-broker/blob/master/internal/app/secretless/configurationmanagers/configfile/fs_watcher.go) and [crd watcher](https://github.com/cyberark/secretless-broker/blob/master/internal/app/secretless/configurationmanagers/kubernetes/crd/crd_watcher.go)) should remain in the codebase so that we can restore the dynamic config reloading functionality in #881

PR #889, opened 2019-09-12
876 Refactor high-level structure of Secretless 

jonahx:
Connected to #876 
Connected to #877 
Connected to #879 

This does the following:

- Rewrites entrypoint as StartSecretless to include extracted pieces of
  plugin.Manager (perf profiling, plugin verification) that aren't part
  of the proper Secretless responsibilities.
- Simplifies profiling and signal handling code and extracts them into
  their own packages
- Defines (most of) the main high-level interfaces for the new architecture
- Defines (most of) the main high-level interfaces for the new architecture
- Creates the collaborators for Secretless proper
- Defines a stub for Secretless proper which accepts
- Creates implementation for extracted EventNotifier


cmd/secretless-broker/main.go
```diff
@@ -2,12 +2,48 @@ package main
...
+	"os/signal"
+	"syscall"
 
-	"github.com/cyberark/secretless-broker/internal/plugin"
 	"github.com/cyberark/secretless-broker/pkg/secretless"
+	v2 "github.com/cyberark/secretless-broker/pkg/secretless/config/v2"
+	"github.com/pkg/profile"
 )
 
 func main() {
```
sgnn7: This PR undoes partially the move of main broker code from cmd/ to pkg/. We should ensure that only flag parsing is done in main.go so that if needed, the broker can be run/started from a package that imports it.

cmd/secretless-broker/main.go
```diff
@@ -2,12 +2,48 @@ package main
...
+
+	cfg := v2.Config{
+		Debug: false,
+		Services: test,
+	}
+
+	secretless := NewSecretless()
+	perfProfile := NewPerformanceProfile()
+
+	StopOnKillSignal(secretless)
```
sgnn7: `StopOnKillSignal`: `kill` is only one of the signals that we watch for. maybe we should change the name to not be specific to signal type.

cmd/secretless-broker/main.go
```diff
@@ -41,5 +77,68 @@ func main() {
...
+		syscall.SIGINT,
+		syscall.SIGQUIT,
+		syscall.SIGTERM,
+	)
+	return signalChannel
+}
+
+type Stopper interface {
+	Stop()
+}
```
sgnn7: Stopper,  NewKillSignalChan,  and StopOnKillSignal should be in a separate package

cmd/secretless-broker/main.go
```diff
@@ -41,5 +77,68 @@ func main() {
...
+		<-killSignals
+		s.Stop()
+	}()
+}
+
+// type is "memory" or "cpu"
+type PerformanceProfile interface {
+	Start()
+	Stop()
+}
```
sgnn7: If both PerformanceProfile and Secretless implement the same methods interface, it might be good to find a common name and extract into a separate package that both of them inherit

internal/secretless.go
```diff
@@ -0,0 +1,39 @@
...
+func NewSecretless(cfg v2.Config, logger log.Logger, evtNotifier v1.EventNotifier) secretless.Secretless {
+	ret := _secretless{
+		config: cfg,
+		logger: logger,
+		eventNotifier: evtNotifier,
+	}
+
+	// TODO: create our unstarted ProxyServices here (sync with Kumbi)
+
+	return &ret
```
sgnn7: `ret` doesn't really tell us what type/object we're working with

cmd/secretless-broker/main.go
```diff
@@ -2,12 +2,48 @@ package main
...
+
+	// TODO: pass in logger, config, eventNotifier, what else?
+
+	var test []*v2.Service
+
+	// checksum logic
+
+	// allAvailablePlugins
+
+	// validate requested plugins
```
sgnn7: these three comments should be a single call to some package - main program doesn't need to care about how it gets the plugins but just that it has the list of them

cmd/secretless-broker/main.go
```diff
@@ -2,12 +2,48 @@ package main
...
+
+	cfg := v2.Config{
+		Debug: false,
+		Services: test,
+	}
+
+	secretless := NewSecretless()
```
sgnn7: This will need some params from CLI so we might need to keep an options struct that we pass into it

cmd/secretless-broker/main.go
jonahx: yeah good point i was thinking of "kill" as generic but you're right it's also a specific signal.  any suggestions?  `StopOnAnyHaltSignal` is verbose but perhaps clear?

cmd/secretless-broker/main.go
jonahx: I see what you mean.  I can go ahead and move all this stuff into `Secretless` per my original plan.

cmd/secretless-broker/main.go
jonahx: > that both of them inherit
Important nitpick: They don't inherit.  They just implement :)

But essentially what you're getting at is that they're both just "services" that can be started and stopped.  Perhaps the interface name should just be `Service`?  

You could call it a `StartStopper` in analogy with `ReadWriter`?  I kind of hate that but it's Go-ish...

cmd/secretless-broker/main.go
sgnn7: `StopOnProgramExit`? 🤷‍♂

cmd/secretless-broker/main.go
sgnn7: > But essentially what you're getting at is that they're both just "services" that can be started and stopped. Perhaps the interface name should just be Service

yup!

cmd/secretless-broker/main.go
```diff
@@ -3,16 +3,21 @@ package main
...
+	"github.com/cyberark/secretless-broker/pkg/secretless/entrypoint"
 )
 
 func main() {
+	params := CmdLineParams()
+	entrypoint.StartSecretless(params)
+}
+
+// CmdLineParams parses all cmd line options and returns the resulting CLIParams.
+func CmdLineParams() *entrypoint.CLIParams {
```
doodlesbykumbi: Yes please!

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,125 @@
...
+	"log"
+	"os"
+
+	"github.com/cyberark/secretless-broker/internal"
+	secretlessLog "github.com/cyberark/secretless-broker/internal/log"
+	"github.com/cyberark/secretless-broker/internal/plugin/v1/event_notifier"
+	"github.com/cyberark/secretless-broker/pkg/secretless"
+	"github.com/cyberark/secretless-broker/pkg/secretless/config"
+	v2 "github.com/cyberark/secretless-broker/pkg/secretless/config/v2"
+
```
doodlesbykumbi: random newline here

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,125 @@
...
+	cfg := ReadConfig(params.ConfigFile)
+	logger := secretlessLog.New(params.DebugEnabled)
+	evtNotifier := event_notifier.New(nil)
+	availPlugins := &internal.AvailPluginStub{}
+
+	// Prepare Secretless
+	secretless := internal.NewSecretless(cfg, availPlugins, logger, evtNotifier)
+	signal.StopOnExitSignal(secretless)
+
+	StartPerformanceProfiling(params.ProfilingMode)
```
doodlesbykumbi: The naming here is a bit misleading. This function conditionally starts performance profiling, I think it might be clearer to have an if statement.

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,125 @@
...
+// when an system halt is raised.
+func StartPerformanceProfiling(profileType string) {
+	// No profiling was requested
+	if profileType == "" {
+		return
+	}
+
+	// Validate requested type
+	if err := profile.ValidateType(profileType); err != nil {
+		log.Fatalln(err)
```
doodlesbykumbi: There's a few places where `log.Fatalln` is used to terminate the program "relatively deep in the stack". Any reason to not propagate errors upwards so that these methods don't have this catastrophic side-effect ?

pkg/secretless/entrypoint/start_secretless.go
jonahx: Ha.  I went back and forth on "the cleanness of all the logic in a func" vs "the naming being perfectly accurate".  I also considered something like `HandlePerformanceProfiling` which makes no false promise of starting -- what do you think of that?  Open to other suggestions as well.... I would like to keep the logic in the func if possible.

pkg/secretless/entrypoint/start_secretless.go
jonahx: Well, the idea is that `StartSecretless` is, in fact, the highest level, logically speaking.  There no one to handle errors up here.  `main` just parses cmd line flags and delegates to `StartSecretless`.

That said, we _probably_ want to make `StartSecretless` a proper function object that we can unit test at some point, and if we did that I would definitely do what you're suggesting.  I sort of feel like that should be a separate story though.

pkg/secretless/entrypoint/start_secretless.go
jonahx: I went ahead and changed to `HandlePerformanceProfiling` for now, but lmk if you don't like this solution.

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,124 @@
+package entrypoint
```
sgnn7: I don't think I've seen naming like this in other codebases nor in ours (`start_secretless` => `<verb>_<type>`) so I'm wondering what might be the reasoning behind such a large deviation from the norm? On face value of what I'm reading, the barrier for using a `.go` file named like that along with a few other public methods that do not seem to correlate too closely seems to be not enough for a deviation but I'd like to hear the other side of the argument on that.

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/cyberark/secretless-broker/internal/profile"
+	"github.com/cyberark/secretless-broker/internal/signal"
+	"github.com/cyberark/secretless-broker/pkg/secretless"
+	"github.com/cyberark/secretless-broker/pkg/secretless/config"
+	v2 "github.com/cyberark/secretless-broker/pkg/secretless/config/v2"
+)
+
+// CLIParams holds the command line flag information that Secretless was started
+// with.
+type CLIParams struct {
```
sgnn7: Nit: I guess this is my original addition fault but maybe this object should be more generic since it was designed to be used to start secretless from non-CLI usage.

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,124 @@
...
+	PluginChecksumsFile string
+	PluginDir           string
+	ProfilingMode       string
+	ShowVersion         bool
+}
+
+// StartSecretless method is the main entry point into the broker after the CLI
+// flags have been parsed
+func StartSecretless(params *CLIParams) {
+
```
sgnn7: nit: Unnecessary LF

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func ReadConfig(cfgFile string) v2.Config {
+	// TODO: Add back in CRD / generalized config option
+	cfg, err := config.LoadFromFile(cfgFile)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	return cfg
+}
```
sgnn7: nit: Not a blocker for this PR but this method should maybe be a part of internal/config since it's part of that functionality?

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,124 @@
...
+
+func ShowVersion(showVersionAndExit bool) {
```
sgnn7: nit: we're already in the ShowVersion method so that part of the param name can most likely be removed since it's somewhat redundant

sgnn7: nit: This method (since it's public) is lacking a comment that golint/govet might complain about

pkg/secretless/entrypoint/start_secretless.go
```diff
@@ -0,0 +1,124 @@
...
+	if checksumFile == "" {
+		log.Println("WARN: No PluginChecksumsFile provided - plugin tampering" +
+			" is possible!")
+		return
+	}
+
+	if _, err = plugin.VerifyPluginChecksums(pluginDir, checksumFile); err != nil {
+		log.Fatalln(err)
+	}
+}
```
sgnn7: nit: Not a blocker here for this PR but this whole block seems like it should be in internal/plugin rather than here

izgeri:
@jonahx should this still be a draft PR? can you add the context you gave in slack for what this PR does to the issue description?

jonahx:
> @jonahx should this still be a draft PR? can you add the context you gave in slack for what this PR does to the issue description?

@izgeri Done.

sgnn7:
@jonahx One more comment since I looked through the code one more time - seems like there's some missing tests for the added functionality too that we should add probably.

internal/secretless.go
jonahx: Raised in channel for resolution as code convention

jonahx: So let's do this, I'll go ahead and change it to something descriptive for now, and then I'll add the question to best practices doc which we'll eventually share in #about-golang, and we can get feedback on various decisions like this and perhaps vote on them, and eventually we'll have a nice best practices doc which us and the IL devs can all share.

pkg/secretless/entrypoint/start_secretless.go
jonahx: Ok so this one is a bit of a longer explanation.  First off, I should say that thing I'm currently calling `Secretless` should be renamed to `StartProxyServices` (for reasons I'll explain shortly).

This is based on a naming convention which I find to be one of the most helpful naming conventions I've ever encountered.  At a high-level, the convention is just:

- Name functions that return things after the noun they return.  Use "AdjectiveNoun" when needed for clarity.
- Name functions that _do_ things using a verb phrase describing what they _do_.  "Do" here means "do something with side effects".  Eg, "CreateUser" (adds user to a db) "StartServiceX" (starts listening on a network), "SaveConfigFile" (writes a file to disk), etc.

So where does this leave us with `StartSecretless`.  Well `StartSecretless` is a "function object".  That is, it's a type that takes in some side-effectful dependencies (injected via ctor), and is the "run" or "executed" (sometimes with additional runtime inputs, but not in this case).  The method that executes the function object is typically called `Run()` or `Call()` or in this case `Start()` (`Start()` is probably a mistake, actually).  Thus the naming convention for a function object is the same as for a side-effectful function.  

More on this technique (in the context of ruby, though that's not really relevant) [here](https://www.icelab.com.au/notes/functional-command-objects-in-ruby/).  I've used these conventions for the `authn` stuff in Conjur and have been happy with the results.

What might be even cleaner than how I have it now is for `StartSecretless` to return a type that has nothing but a `Call()` method, and for the `Call()` method to return a `Stopper` function.  Anyway, what do you think -- both about all this and the specific question I just asked?

pkg/secretless/entrypoint/start_secretless.go
jonahx: So my reasoning here was that `config.LoadFromFile(cfgFile)` is already in `config` (as you're saying) but this is a specific helper func to `StartSecretless` which has the property that it should just _die_ if there's any error -- and that is a use-case specific thing to `StartSecretless`.  Versus a general utility like `config.LoadFromFile`which you'd expect to return an error.

wdyt?

pkg/secretless/entrypoint/start_secretless.go
doodlesbykumbi: Yup, definitely coming from the unit testable angle. Happy for it to be a separate story.

PR #890, opened 2019-09-13
implement tcp proxy service

doodlesbykumbi:

An implementation of the TCPProxyService. There's some stuff left to do and test cases to add.
Using the Secretless `logger` as the primary means for error handling is a TODO


internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+// Zeroizes the values of the fetched credentials. We don't want to
+// rely on garbage collection for this (it might be slow and/or only free them) so
+// we manually clear
+func zeroizeCredentials(backendCredentials map[string][]byte) {
+	for _, rawCredential := range reflect.ValueOf(backendCredentials).MapKeys() {
+		credential := rawCredential.String()
+		for byteIndex := range backendCredentials[credential] {
+			backendCredentials[credential][byteIndex] = 0
+		}
+	}
```
sgnn7: This seems a bit of a roundabout way to do it I think. the map can return the key by itself without reflect and you don't need to turn that key into a string either.

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func duplexStream(source io.ReadWriter, destination io.ReadWriter) <- chan error {
+	errChan := make(chan error, 1)
```
sgnn7: hmm.. if we get more than one error buffered in the chan, this code will block. is that intended?

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+// NewTCPProxyService constructs a new instance of a TCP ProxyService. The
+// constructor takes a TCP Connector, CredentialResolver and Listener.
+// A TCP ProxyService is able to Connect with Credentials then subsequently stream
+// bytes between client and target service
+func NewTCPProxyService(
+	connector tcp.Connector,
+	resolveCredentials credentialsResolver,
+	listener net.Listener,
+	) (internal.ProxyService, error) {
+	errors := validation.Errors{}
```
sgnn7: Add newline above - it's hard to tell where the def ends and code starts

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+	return &tcpProxyService{
+		connector:          connector,
+		resolveCredentials: resolveCredentials,
+		listener:           listener,
+		done:               false,
+		error:              make(chan error),
+	}, nil
+}
+
+type credentialsResolver func() (map[string][]byte, error)
```
sgnn7: Don't we already have this type somewhere? Even if not, we should move it out since it will be used in more places

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (ps *tcpProxyService) handleConnection(clientConn net.Conn) error {
```
sgnn7: Not: Not a big deal and I know not everyone agrees with these but `ps` is really hard to identify as a tcpProxyService type. Any way to just change `ps`->`proxy`?

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+	var targetConn net.Conn
+	defer func() {
+		if clientConn != nil {
+			_ = clientConn.Close()
```
sgnn7: Nit: We could print these out to the logger as debug info when you have that added in

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return err
+	}
+
+	targetConn, err = ps.connector(clientConn, backendCredentials)
+	if err != nil {
+		return err
+	}
+
+	return <- duplexStream(clientConn, targetConn)
```
sgnn7: nifty syntax - I haven't seen this before!

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (ps *tcpProxyService) OpError() chan error {
```
sgnn7: `Op` in `OpError` is a bit unclear as to what it actually is - a different name might be helpful here.

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+				return
+			}
+			go func() {
+				err := ps.handleConnection(conn)
+				ps.error <- err
+			}()
+		}
+	}()
+
+	return nil
```
sgnn7: We might consider removing error return or return an error on socket creation.

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+			strings.Repeat("\x00", len([]byte("credValue"))),
+		)
+	})
+}
+
+func Test_tcpProxyService(t *testing.T) {
+	t.Run("proxy service streams from source to dest", func(t *testing.T) {
+		source, sourceBack := net.Pipe()
+		destination, destinationBack := net.Pipe()
+		listener :=	new(listenerMock)
```
sgnn7: Nit: spacing

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+			source,
+			nil)
+
+		// tcp service with 'destination" mock and listener whose first call to Accept return the
+		// "source" mock then blocks forever
+		ps, err := NewTCPProxyService(
+			func(clientConn net.Conn, secrets plugin.SecretsByID) (backendConn net.Conn, err error) {
+				return destination, nil
+			},
+			func() (i map[string][]byte, e error) {
```
sgnn7: could params be `_` - seems like the names aren't needed

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+				return nil, nil
+			},
+			listener)
+
+		err = ps.Start()
+		assert.NoError(t, err)
```
sgnn7: Don't we need that `if (err) fail` thing here too?

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		go sourceBack.Write([]byte("good heavens"))
```
sgnn7: Nit: this string should be a constant in this test rather than being duplicated twice. also, might be advisable to choose a fully non-controversial string val for it.

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		data := make([]byte, 123)
+		dataLen, err := destinationBack.Read(data)
+		assert.NoError(t, err)
+		assert.Equal(
```
sgnn7: Also needing `if(err)`?

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		go destinationBack.Write([]byte("meow"))
+		len, err = sourceBack.Read(data)
+		assert.NoError(t, err)
+		assert.Equal(
+			t,
+			string(data[:len]),
+			"meow",
+		)
+	})
```
sgnn7: use constants/vars for repeated strings

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+		assert.Equal(
+			t,
+			string(data[:len]),
+			"moo",
+		)
+
+		go destinationBack.Write([]byte("meow"))
+		len, err = sourceBack.Read(data)
+		assert.NoError(t, err)
+		assert.Equal(
```
sgnn7: needs `if (err)`?

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t.Run("streams between source and dest", func(t *testing.T) {
+		source, sourceBack := net.Pipe()
+		destination, destinationBack := net.Pipe()
+
+		go sourceBack.Write([]byte("moo"))
+		go duplexStream(source, destination)
+
+		data := make([]byte, 123)
+		len, err := destinationBack.Read(data)
+		assert.NoError(t, err)
```
sgnn7: needs `if (err)`?

internal/app/secretless/utils_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/stretchr/testify/mock"
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin"
+)
+
+type connectorMock struct {
+	mock.Mock
+}
+func (c *connectorMock) connect(clientConn net.Conn, secrets plugin.SecretsByID) (backendConn net.Conn, err error) {
```
sgnn7: LF above here

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+	validation "github.com/go-ozzo/ozzo-validation"
+
+	"github.com/cyberark/secretless-broker/internal"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
+)
+
+// Zeroizes the values of the fetched credentials. We don't want to
+// rely on garbage collection for this (it might be slow and/or only free them) so
+// we manually clear
+func zeroizeCredentials(backendCredentials map[string][]byte) {
```
izgeri: how does this work if it passes the argument in by value?

doodlesbykumbi: a map is a pointer, see https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it

izgeri: TIL. Thanks for the link @doodlesbykumbi!

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+			string(creds["credName"]),
+			strings.Repeat("\x00", len([]byte("credValue"))),
+		)
+	})
+}
+
+func Test_tcpProxyService(t *testing.T) {
+	t.Run("proxy service streams from source to dest", func(t *testing.T) {
+		source, sourceBack := net.Pipe()
+		destination, destinationBack := net.Pipe()
```
jonahx: These names weren't immediately obvious to me, and I would have liked a bit more handholding.  How about:
```go
// This allows us to control and view what happens in the client and backend connections
// that secretless is proxying.  Whatever we write into `clientConnSrc` can be read from
// `clientConn` by the ProxyService, and whatever the ProxyService writes into
// backendConn will be pipe into backendConnDest, so we can verify it.
clientConn, clientConnSrc := net.Pipe()
backendConn, backendConnDest := net.Pipe()
```

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+				return nil, nil
+			},
+			listener)
+
+		err = ps.Start()
+		assert.NoError(t, err)
+
+		go sourceBack.Write([]byte("good heavens"))
+
+		data := make([]byte, 123)
```
jonahx: `123` threw me because I thought it had some specific meaning.  I'd either use the exact `len` of what we're writing, or something like 999 which conveys "something way bigger than needed"

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+		assert.Error(t, err)
+		assert.Equal(
+			t,
+			string(creds["credName"]),
+			strings.Repeat("\x00", len([]byte("credValue"))),
+		)
+	})
+}
+
+func Test_tcpProxyService(t *testing.T) {
```
jonahx: Any reason this isn't `TestTCPProxyService`?

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+		assert.Equal(
+			t,
+			string(data[:dataLen]),
+			"good heavens",
+		)
+	})
+}
+
+func Test_duplexStream(t *testing.T) {
+	t.Run("streams between source and dest", func(t *testing.T) {
```
jonahx: This feels like it should be another test within `TestTCPProxyService`, and then you can delete `Test_duplexStream`

internal/app/secretless/tcp_proxy_service.go
doodlesbykumbi: Not sure I follow. This method exists because this implementation needs to adhere to the ProxyService interface

internal/app/secretless/tcp_proxy_service.go
doodlesbykumbi: Changed to OperationalError

internal/app/secretless/utils_test.go
```diff
@@ -0,0 +1,77 @@
...
+	args := c.Called()
+	// it then returns whatever we tell it to return
+	// in this case true to simulate an SMS Service Notification
+	// sent out
+	if args.Get(0) == nil {
+		return nil, args.Error(1)
+	}
```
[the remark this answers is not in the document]
doodlesbykumbi: Unfortunately these are needed otherwise there'll be a panic because nil is not whatever the type of the args is

internal/app/secretless/tcp_proxy_service_test.go
doodlesbykumbi: It seems to be a Go convention for the test case to be Test_lowercaseSomething when it's an internal thing vs TestUpcaseSomething when it's a public thing.

internal/app/secretless/tcp_proxy_service.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (ps *tcpProxyService) OpError() chan error {
+	return ps.error
+}
+
+// Start initiates the net.Listener to listen for incoming connections
+func (ps *tcpProxyService) Start() error {
+	go func() { // n go routines for n tcp proxyservices
+		// TODO: add mutex to done
```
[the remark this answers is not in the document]
doodlesbykumbi: The loop inside the go func is blocking since it calls Accept over and over.

internal/app/secretless/tcp_proxy_service_test.go
```diff
@@ -0,0 +1,126 @@
...
+			resolveCredentials: credentialResolver.resolveCredentials,
+			listener:           listener,
+		}
+		creds := map[string][]byte{
+			"credName": []byte("credValue"),
+		}
+		credentialResolver.On("resolveCredentials").Return(
+			creds,
+			fmt.Errorf("couldn't resolve credentials"))
+
```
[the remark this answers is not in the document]
doodlesbykumbi: creds isn't a temp var, it's the entity we're checking has been zeroized

internal/app/secretless/tcp_proxy_service.go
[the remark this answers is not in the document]
jonahx: Ah ok, that makes sense now.  But don't we still want to protect against calling `Start()` multiple times?

izgeri:
@doodlesbykumbi do you know why the k8s tests are failing in this branch? I don't think you made any changes that should affect this ...

internal/proxy_service/tcp/proxy_service.go
```diff
@@ -0,0 +1,144 @@
...
+	error               chan error
+}
+
+// NewProxyService constructs a new instance of a TCP ProxyService. The
+// constructor takes a TCP Connector, CredentialResolver and Listener.
+// A TCP ProxyService is able to Connect with Credentials then subsequently stream
+// bytes between client and target service
+func NewProxyService(
+	connector tcp.Connector,
+	resolveCredentials internal.CredentialsRetriever,
```
jonahx: Nit: switching the language here seems unnecessary.  I'd just called the local var `retrieveCredentials`

internal/proxy_service/tcp/proxy_service.go
```diff
@@ -0,0 +1,144 @@
...
+func (proxy *proxyService) handleConnection(clientConn net.Conn) error {
+	var targetConn net.Conn
+	defer func() {
+		if clientConn != nil {
+			_ = clientConn.Close()
+		}
+		if clientConn != nil {
+			_ = targetConn.Close()
+		}
+	}()
```
jonahx: question: this might be me misunderstanding something, but wouldn't this be called as soon as the function returns (which it would immediately, bc the streaming stuff happens in go routines) and then it would close the connections?

internal/proxy_service/tcp/proxy_service.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return err
+	}
+
+	return <-duplexStream(clientConn, targetConn)
+}
+
+func (proxy *proxyService) OperationalError() chan error {
+	return proxy.error
+}
```
jonahx: I don't see any references to this.... What's it being used for?

internal/proxy_service/tcp/proxy_service_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/cyberark/secretless-broker/internal/proxy_service/mock"
+)
+
+const testString1 = "good heavens"
+const testString2 = "moderate heavens"
```
jonahx: ❤️

internal/proxy_service/tcp/proxy_service_test.go
```diff
@@ -0,0 +1,174 @@
...
+		assert.NoError(t, err)
+		if err != nil {
+			return
+		}
+
+		err = ps.Start()
+		assert.Error(t, err)
+		if err == nil {
+			return
+		}
```
jonahx: This `if` appears superfluous.

internal/proxy_service/tcp/proxy_service_test.go
```diff
@@ -0,0 +1,174 @@
...
+	})
+
+	t.Run("proxy service streams from source to dest", func(t *testing.T) {
+		// prepare
+
+		// This allows us to control and view what happens in the client and
+		// backend connections that Secretless is proxying.  Whatever we write
+		// into `clientConnSrc` can be read from `clientConn` by the
+		// ProxyService, and whatever the ProxyService writes into backendConn will be pipe into backendConnDest, so we can verify it.
```
jonahx: Nit: Split line so it's not so long.

internal/proxy_service/tcp/proxy_service_test.go
```diff
@@ -0,0 +1,174 @@
...
+		clientConn, clientConnSrc := net.Pipe()
+		backendConn, backendConnDest := net.Pipe()
+
+		connector := mock.NewConnector()
+		connector.On("Connect").Return(
+			backendConn,
+			nil)
```
jonahx: Nit: combine into 1 line.

internal/proxy_service/tcp/proxy_service_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+		credentialRetriever := mock.NewCredentialRetriever()
+		credentialRetriever.On("RetrieveCredentials").Return(
+			nil,
+			nil)
+
```
jonahx: Nit: combine into 1 line

internal/proxy_service/tcp/proxy_service.go
doodlesbykumbi: Before bring in the logger my thoughts were to expose an error channel that the Secretless "Framework" could use to report errors from ProxyServices

jonahx: Don't have to change this now, but `Operational` doesn't seem to add any info over, say, `ErrorChannel()` or perhaps, `ConnectionErrors()`, to imply that these are errors that occur as a result of processing a new connection?  Open to better ideas....

internal/proxy_service/tcp/proxy_service.go
jonahx: Ah, ofc.

internal/proxy_service/tcp/proxy_service.go
doodlesbykumbi: This is gone now. Logger instead

internal/proxy_service/tcp/proxy_service_test.go
```diff
@@ -0,0 +1,348 @@
...
+		err = ps.Start()
+
+		// sanity assert
+		assert.NoError(t, err)
+		if err != nil {
+			return
+		}
+
+		// artificial sleep needed to wait for Errorf
+		time.Sleep(time.Millisecond)
```
jonahx: We don't need to do this now, but it's worth pointing out: One of the great things bout mocks is we totally control them.  So you could easily add a channel to the logger mock (this would change only the interface of the mock ctor, not the thing being mocked), and then use that to synchronize here without the indeterminacy of sleep.

doodlesbykumbi: Great, I was wrecking my brain about this one. Could you provide an example please, that'd help a whole bunch...

internal/proxy_service/tcp/proxy_service.go
```diff
@@ -0,0 +1,151 @@
...
+				proxy.logger.Warnf("failed on closing client connection: %s", err)
+			}
+		}
+		if targetConn != nil {
+			err := targetConn.Close()
+			if err != nil {
+				proxy.logger.Warnf("failed on closing target connection: %s", err)
+			}
+		}
+	}()
```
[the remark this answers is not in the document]
doodlesbykumbi: I like it

internal/proxy_service/mock/logger.go
```diff
@@ -0,0 +1,92 @@
+package mock
```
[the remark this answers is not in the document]
doodlesbykumbi: fair point

internal/proxy_service/tcp/proxy_service_test.go
```diff
@@ -0,0 +1,349 @@
...
+		// assert third packet
+		dataLen, err = backendConnDest.Read(data)
+		assert.Equal(
+			t,
+			string(data[:dataLen]),
+			testString3,
+		)
+	})
+
+	t.Run("proxy service streams packets between source and dest", func(t *testing.T) {
```
doodlesbykumbi: @jonahx here's the test between source and dest

Issue #878, opened 2019-09-05
AllAvailablePlugins type interface is implemented


jonahx:
- Write independently testable function `func ExternalPlugins(dir) AvailablePlugins`
- Similarly: `InternalPlugins() AvailablePlugins`
- And, for use by `Secretless`: `InternalPlugins(externalPluginDir) AvailablePlugins` which combines the two.

AC:
~- [ ] Verify and then create interface above~
~- [ ] Verify this can still work with gRPC~
- [ ] `AvailablePlugins` interface is implemented

izgeri:
this effort will create the one bit of code that knows
- where plugins are located
- how to load them
- have them all loaded up and indexed by name into the HTTPPlugin and TCPPlugin maps as the output

it uses info about what's statically available in the filesystem to generate an object that lists available connector plugins, but doesn't actually instantiate them

in another bit of code all of the proxyservices are actually instantiated based on secretless.yml

izgeri:
@sgnn7 just to note in #845 the data in this will be used with the config to determine which services should be active in secretless

sgnn7:
WIP branch is here: https://github.com/cyberark/secretless-broker/compare/883-allavailableplugins-interface?expand=1. Moving back to defined.

PR #892, opened 2019-09-16
Add AvailablePlugins module

sgnn7:
Adds plugin loading logic to our codebase

Connected to #878

### DIFF
--- a/internal/plugin/available_plugins.go
+++ b/internal/plugin/available_plugins.go
@@ -1,13 +1,103 @@
 package plugin
 
 import (
+	"github.com/cyberark/secretless-broker/pkg/secretless/log"
 	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/http"
 	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
 )
+
+// CompatiblePluginAPIVersion indicates what matching API version an external plugin
+// must have so that we are capable of loading it.
+var CompatiblePluginAPIVersion = "0.1.0"
 
 // AvailablePlugins is an interface that provides a list of all the available
 // plugins for each type that the broker supports.
 type AvailablePlugins interface {
 	HTTPPlugins() map[string]http.Plugin
 	TCPPlugins() map[string]tcp.Plugin
+}
+
+// Plugins represent a holding object for a bundle of plugins of different types.
+type Plugins struct {
+	HTTPPluginsByID map[string]http.Plugin
+	TCPPluginsByID  map[string]tcp.Plugin
+}
+
+// HTTPPlugins returns only the HTTP plugins in the Plugins struct.
+func (plugins *Plugins) HTTPPlugins() map[string]http.Plugin {
+	return plugins.HTTPPluginsByID
+}
+
+// TCPPlugins returns only the TCP plugins in the Plugins struct.
+func (plugins *Plugins) TCPPlugins() map[string]tcp.Plugin {
+	return plugins.TCPPluginsByID
+}
+
+// AllAvailablePlugins returns the full list of internal and external plugins
+// available to the broker.
+func AllAvailablePlugins(
+	pluginDir string,
+	checksumsFile string,
+	logger log.Logger,
+) (AvailablePlugins, error) {
+
+	return AllAvailablePluginsWithOptions(
+		pluginDir,
+		checksumsFile,
+		GetInternalPluginsFunc,
+		LoadPluginsFromDir,
+		logger,
+	)
+}
+
+// AllAvailablePluginsWithOptions returns the full list of internal and external plugins
+// available to the broker using explicitly-defined lookup functions.
+// TODO: Test this
+func AllAvailablePluginsWithOptions(
+	pluginDir string,
+	checksumsFile string,
+	internalLookupFunc InternalPluginLookupFunc,
+	externalLookupfunc ExternalPluginLookupFunc,
+	logger log.Logger,
+) (AvailablePlugins, error) {
+
+	internalPlugins, err := InternalPlugins(internalLookupFunc)
+	if err != nil {
+		return nil, err
+	}
+
+	externalPlugins, err := ExternalPlugins(
+		pluginDir,
+		externalLookupfunc,
+		logger,
+		checksumsFile,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	httpPlugins := internalPlugins.HTTPPlugins()
+	for name, httpPlugin := range externalPlugins.HTTPPlugins() {
+		if _, ok := httpPlugins[name]; ok {
+			logger.Warnf("Internal plugin '%s' is replaced by an externally-provided plugin",
+				name)
+		}
+
+		httpPlugins[name] = httpPlugin
+	}
+
+	tcpPlugins := internalPlugins.TCPPlugins()
+	for name, tcpPlugin := range externalPlugins.TCPPlugins() {
+		if _, ok := tcpPlugins[name]; ok {
+			logger.Warnf("Internal plugin '%s' is replaced by an externally-provided plugin",
+				name)
+		}
+
+		tcpPlugins[name] = tcpPlugin
+	}
+
+	return &Plugins{
+		HTTPPluginsByID: httpPlugins,
+		TCPPluginsByID:  tcpPlugins,
+	}, nil
 }
--- a/internal/plugin/external_plugins.go
+++ b/internal/plugin/external_plugins.go
@@ -1,0 +1,242 @@
+package plugin
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	go_plugin "plugin"
+	"strings"
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/log"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/http"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
+)
+
+// ExternalPluginLookupFunc returns all available external plugins.
+type ExternalPluginLookupFunc func(
+	pluginDir string,
+	checksumfile string,
+	logger log.Logger,
+) (map[string]*go_plugin.Plugin, error)
+
+// LoadPluginsFromDir loads all plugins from a specified directory and returns
+// Plugins struct with tcp and http connectors.
+func LoadPluginsFromDir(
+	pluginDir string,
+	checksumsFile string,
+	logger log.Logger,
+) (map[string]*go_plugin.Plugin, error) {
+
+	filePaths, err := checkedPlugins(pluginDir, checksumsFile, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	return loadPluginFiles(filePaths, logger)
+}
+
+func checkedPlugins(
+	pluginDir string,
+	checksumsFile string,
+	logger log.Logger,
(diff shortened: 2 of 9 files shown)


internal/plugin/available_plugins.go
```diff
@@ -11,3 +13,143 @@ type AvailablePlugins interface {
...
+
+	plugins := Plugins{
+		HTTPPluginMap: map[string]http.Plugin{},
+		TCPPluginMap:  map[string]tcp.Plugin{},
+	}
+
+	for _, rawPlugin := range rawPlugins {
+		pluginInfoSym, err := (*rawPlugin).Lookup("PluginInfo")
+		if err != nil {
+			//TODO: Decide what to do here
```
izgeri: at current if an external plugin errs we just skip loading it and print an error - can we do that for most of these too? for example: https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/manager.go#L426

internal/plugin/available_plugins.go
```diff
@@ -11,3 +13,143 @@ type AvailablePlugins interface {
...
+		if !ok {
+			//TODO: Decide what to do here
+		}
+
+		// TODO: Do some sort of check on api version
+		// TODO: What do we do if it fails?
+
+		pluginType := pluginInfo["type"]
+		if pluginType == "" {
+			//TODO: Decide what to do here
```
izgeri: "type is required" [skip plugin]

internal/plugin/available_plugins.go
```diff
@@ -11,3 +13,143 @@ type AvailablePlugins interface {
...
+		}
+
+		pluginId := pluginInfo["id"]
+		if pluginId == "" {
+			//TODO: Decide what to do here
+			//TODO: What if the plugin id doesn't fit our charset reqs?
```
izgeri: "id is missing" [skip plugin]
"plugin id does not meet [char requirements]" [skip plugin]

internal/plugin/available_plugins.go
```diff
@@ -11,3 +13,143 @@ type AvailablePlugins interface {
...
+			}
+
+			tcpPluginFunc, ok := pluginSym.(func() tcp.Plugin)
+			if !ok {
+				//TODO: Decide what to do here
+			}
+
+			plugins.TCPPluginMap[pluginId] = tcpPluginFunc()
+		default:
+			//TODO: Decide what to do here
```
izgeri: if it doesn't match http or tcp, that's also not supported, right? so we should write an error and skip loading it

internal/plugin/available_plugins.go
```diff
@@ -11,3 +18,235 @@ type AvailablePlugins interface {
...
+		// We override file listing if we did a verification to prevent additions
+		// to plugins between verification and loading the plugins.
+		var err error
+		if files, err = VerifyPluginChecksums(pluginDir, checksumsFile); err != nil {
+			logger.Errorln(err)
+			return nil, err
+		}
+	} else if len(files) > 0 {
+		logger.Warnln("Plugin hashes were not provided - tampering will not be detectable!")
+	}
```
jonahx: Here's a piece of cleanup that's not critical (because the above is still pretty clear) but I think is worth doing for extra crispness.  In the above both the nested `if` and the `if... else` break the "line of site" ideal.  In this case, the underlying issue imo is that we're mixing the "load the plugins" code (below this  my comment) and the "deal with checksums" code above it.  That is, we have a SRP violation which is expressing itself in the complexity of those `if`s.  

I'd recommend pulling out the "load the plugins" code into a `loadPluginFiles(files, logger)` method, which does nothing but file loading -- no checksum verification.

Then the code above gets nice and flat:

```go
func loadCheckedPlugins(
	pluginDir string,
	checksumsFile string,
	logger log.Logger,
) ([]*go_plugin.Plugin, error) {

	uncheckedFiles, err := ioutil.ReadDir(pluginDir)
	if err != nil {
		return nil, err
	}

	// no files found, return empty slice
	if len(uncheckedFiles) == 0 {
		return []*go_plugin.Plugin{}, nil
	}

	if checksumsFile == "" {
		logger.Warnln("Plugin hashes were not provided - tampering will not be detectable!")
		return loadPluginFiles(uncheckedFiles, logger), nil
	}

	if checkedFiles, err = VerifyPluginChecksums(pluginDir, checksumsFile); err != nil {
		logger.Errorln(err)
		return nil, err
	}

	return loadPluginFiles(checkedFiles, logger), nil
}

```

sgnn7: done

internal/plugin/available_plugins.go
```diff
@@ -1,13 +1,81 @@
...
+
+// TCPPlugins returns only the TCP plugins in the Plugins struct.
+func (plugins *Plugins) TCPPlugins() map[string]tcp.Plugin {
+	return plugins.TCPPluginMap
+}
+
+// AllAvailablePlugins returns the full list of internal and external plugins
+// available to the broker.
+// TODO: Test this
+func AllAvailablePlugins(pluginDir string,
```
doodlesbykumbi: nit: indent the args the same ? 
as you did with:
```
func ExternalPlugins(
	pluginDir string,
```

maybe we can agree on a style and be consistent with it

internal/plugin/external_plugins.go
```diff
@@ -0,0 +1,228 @@
...
+
+	plugins := Plugins{
+		HTTPPluginMap: map[string]http.Plugin{},
+		TCPPluginMap:  map[string]tcp.Plugin{},
+	}
+
+	for rawPluginName, rawPlugin := range rawPlugins {
+		logger.Infof("Loading plugin '%s'...", rawPluginName)
+
+		pluginType, pluginID, err := parsePluginMetadata(rawPlugin, rawPluginName, logger)
```
doodlesbykumbi: the `logger` is passed to parsePluginMetadata, getSymbolFromName etc. would it not be better for them to propagate errors upwards so that they can be logged here ? as is being done below

internal/plugin/external_plugins.go
```diff
@@ -0,0 +1,228 @@
...
+	symbolName string,
+	logger log.Logger,
+) (go_plugin.Symbol, error) {
+
+	symbol, err := (*rawPlugin).Lookup(symbolName)
+	if err != nil {
+		logger.Errorln(err)
+	}
+	return symbol, err
+}
```
jonahx: Similar to Kumbi's comment, I def think these helper method should just return errors, and the logging responsibility should always be maintained in the highest level funcs.

internal/plugin/external_plugins.go
```diff
@@ -0,0 +1,228 @@
...
+	checksumsFile string,
+	logger log.Logger,
+) (map[string]*go_plugin.Plugin, error) {
+
+	filenames, err := getCheckedPlugins(pluginDir, checksumsFile, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	return loadPluginFiles(pluginDir, filenames, logger)
```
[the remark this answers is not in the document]
sgnn7: Fixed but we're getting back the odd-looking `if`. Since VerifyPluginChecksums returns file objects it's either that or we have to make filenames array twice so I went with the former.

sgnn7:
@doodlesbykumbi / @jonahx / @izgeri fixed all comments afaict

internal/plugin/external_plugins.go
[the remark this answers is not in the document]
jonahx: _it is not the spoon that bends, it is only yourself..._

In this case you can either:

1. Change `VerifyPluginChecksums` to return full paths.  It's only being referenced in one place atm... my new code in entrypoint, but I don't even use the return value.  Downside here is you might have to update its tests.

2.  Create a quickie wrapper function, whose code you've actually already written.  I'd probably just make the function package private but it could even be local to this func.  Ie:
```go
	filePaths := func(path string, infos []os.FileInfo) {
		filePaths := []string{}
		for _, info := range infos {
			filePaths = append(filePaths, path.Join(path, info.Name()))
		}
	}
```
Then you can keep the old flat code:
```go
func getCheckedPlugins(
	pluginDir string,
	checksumsFile string,
	logger log.Logger,
) ([]*go_plugin.Plugin, error) {

	uncheckedFiles, err := ioutil.ReadDir(pluginDir)
	if err != nil {
		return nil, err
	}

	// no files found, return empty slice
	if len(uncheckedFiles) == 0 {
		return []*go_plugin.Plugin{}, nil
	}

	if checksumsFile == "" {
		logger.Warnln("Plugin hashes were not provided - tampering will not be detectable!")
		uncheckedPaths := filePaths(uncheckedFiles, pluginDir)
		return loadPluginFiles(uncheckedPaths, logger), nil
	}

	if checkedFiles, err = VerifyPluginChecksums(pluginDir, checksumsFile); err != nil {
		logger.Errorln(err)
		return nil, err
	}

	checkedPaths := filePaths(checkedFiles, pluginDir)
	return loadPluginFiles(checkedPaths, logger), nil
}
```

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -35,13 +34,17 @@ type SecretlessOptions struct {
...
 	showVersion(params.ShowVersion)
 
-	verifyPlugins(params.PluginDir, params.PluginChecksumsFile)
-
 	// Construct the deps of Service
 	cfg := readConfig(params.ConfigFile)
 	logger := secretlessLog.New(params.DebugEnabled)
 	evtNotifier := eventnotifier.New(nil)
-	availPlugins := &proxyservice.AvailPluginStub{}
+	availPlugins, err := plugin.AllAvailablePlugins(params.PluginDir,
```
doodlesbykumbi: nit: re consistent indentation choice...
there's still at least one instance of both:
```
plugin.AllAvailablePlugins(params.PluginDir,
	params.PluginChecksumsFile,...
```
and
```
plugin.AllAvailablePlugins(
	params.PluginDir,
	params.PluginChecksumsFile,...
```

internal/plugin/external_plugins.go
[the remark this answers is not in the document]
sgnn7: > Change VerifyPluginChecksums to return full paths

This isn't advisable. While we don't use the functionality provided by this module, returned `os.FileInfo` has a higher guarantee that the underlying file object did not change between checking the checksum and loading the file. Even the move to compose the paths earlier is already reducing the assurance that the file we verified is the same one we are loading so weakening this further would not be good.

> Create a quickie wrapper function

I'll do this due to time but I disagree on the approach tbh if it's just to avoid a single + small if/else block but we can discuss this in slack at a later point.

pkg/secretless/entrypoint/entrypoint.go
sgnn7: fixed for now - we can talk about the long-term approach on slack for this

internal/plugin/external_plugins.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func getFilePaths(pluginDir string, files []os.FileInfo) []string {
+	filePaths := []string{}
+	for _, file := range files {
+		filePaths = append(filePaths, path.Join(pluginDir, file.Name()))
+	}
+
+	return filePaths
+}
```
jonahx: Naming nit, though one I actually care about.  I want to get away from using `get`/`Get` for pure functions, such as this one.  I think adding a `get` or `fetch` prefix is ok, and sometime preferable, when we're interacting with the file system or network (eg, `getCheckedPlugins` above), but when we're just returning values with no possibility of error I'd much prefer to stick with "name the function after the thing it returns".

PR #897, opened 2019-09-18
limit k8s demo wait script

doodlesbykumbi:
limit k8s demo wait script to avoids forever builds


hughsaunders:
Related: conjurinc/ops#385

PR #898, opened 2019-09-18
implement an example tcp plugin

doodlesbykumbi:
1. implement an example tcp plugin


test/plugin/secretless_test_harness/main.go
```diff
@@ -0,0 +1,71 @@
...
+	tcpinternal "github.com/cyberark/secretless-broker/internal/proxyservice/tcp"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
+)
+
+// ListenPort is the port that the echo server is listening on
+const ListenPort = "6175"
+
+func main() {
+	log.Printf("listening on %s\n", ListenPort)
+	server, err := net.Listen("tcp", ":"+ListenPort)
```
izgeri: since this is our example plugin, it will probably good to explain all of the things that are happening here in main with comments

this isn't needed for e2e, but I think we'll want to tell a story in our example before we can call the example plugin "done"

test/plugin/secretless_test_harness/Makefile
```diff
@@ -0,0 +1,8 @@
+build_example:
+	go build -buildmode=plugin -o example.so ./example_plugin.go
+run_echo:
+	go run ../echo/main.go
+run_proxy:
+	go run main.go
+client:
+	nc 0.0.0.0 6175
```
sgnn7: We should keep away from makefiles - they tend to complicate things _really_ quickly

test/plugin/secretless_test_harness/README.md
```diff
@@ -0,0 +1,17 @@
+# secretless TCP plugin test harness
```
sgnn7: Capitalization

test/plugin/secretless_test_harness/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+1. Create your own plugin shared object file with the filename `example.so` in this folder. Otherwise, you can run `make build_example` to build the example.
+2. In a separate shell run `make run_echo` to run the TCP echo server at 127.0.0.1:6174.
+3. In a separate shell run `make run_proxy` to run the test harness with the TCP plugin loaded.
```
sgnn7: This should be secretless itself with some mix of internal funcs rather than a 1-off codebase

test/plugin/secretless_test_harness/main.go
sgnn7: Different comment but this should be `localhost`, not all interfaces probably

test/plugin/secretless_test_harness/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+	goPlugins "plugin"
+
+	secretlessLog "github.com/cyberark/secretless-broker/internal/log"
+	tcpinternal "github.com/cyberark/secretless-broker/internal/proxyservice/tcp"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
+)
+
+// ListenPort is the port that the echo server is listening on
+const ListenPort = "6175"
```
sgnn7: This should be probably randomized/passed-in to avoid parallel-build conflicts

doodlesbykumbi: It's intended for manual testing so the port definitely needs to be predictable.

sgnn7: What about having a default but capability to be over-ridden by CLI args for automation?

test/plugin/secretless_test_harness/main.go
doodlesbykumbi: both addressed

test/plugin/secretless_test_harness/Makefile
doodlesbykumbi: @sgnn7 i'm curious, how so ? in this case it's convenient to use them for showing the value prop of the harness.

sgnn7: It's convenient here but it's an additional tool that we probably don't need is the way I consider it. Makefiles also tend to get [unmaintainable](https://github.com/torvalds/linux/blob/master/Makefile) at a much quicker rate than pure Bash which we already use everywhere. Why not just have a few (or even just 1) bin script(s)?

If you're looking for a build system more complex than Bash, we should look into Meson instead since upstream tooling for many linux projects is switching away from Makefiles as they end up unmaintainable.

test/plugin/secretless_test_harness/main.go
sgnn7: Hmm.. want to ping me on slack or was this resolved on accident?

test/plugin/secretless_test_harness/main.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		log.Printf("error: %v\n", err)
+		os.Exit(1)
+	}
+
+	tcpPlugins := plugins.TCPPlugins()
+
+	// only use first TCP plugin
+	var pluginId string
+	for k := range tcpPlugins {
```
sgnn7: What's `k` in this context? I get that there's a bit of a preference for single-letter vars but this one I can't even tell what it's trying to convey.

sgnn7:
@doodlesbykumbi Seems like there's a few easy-to-resolve comments on CC too (I know some are a pita so you can ignore those)

test/plugin/secretless_test_harness/main.go
doodlesbykumbi: Oh yeah. I thought the comment would be helpful in explaining what's happening i.e. Extracting the first TCP plugin

internal/proxyservice/tcp/proxy_service.go
```diff
@@ -113,16 +118,36 @@ func (proxy *proxyService) handleConnection(clientConn net.Conn) error {
...
+	select {
+	case sourceErr := <-clientErrChan:
+		if sourceErr == nil {
+			proxy.logger.Infof("%v connection closed by client.\n", clientConn.LocalAddr())
+			return nil
+		}
+		return fmt.Errorf("%v connection failed on streaming from client connection: %s", clientConn.LocalAddr(), sourceErr)
+	case destErr := <- destErrChan:
+		if destErr == nil {
+			proxy.logger.Infof("%v connection closed by target.\n", clientConn.LocalAddr())
```
sgnn7: You use the `proxy.logger` a decent amount - maybe we could reduce code if you had `logger := proxy.logger` somewhere higher?

internal/signal/signal.go
```diff
@@ -43,8 +43,6 @@ func newHaltSignalChan() chan os.Signal {
 func StopOnExitSignal(s Stopper) {
 	killSignals := newHaltSignalChan()
 
-	go func() {
-		<-killSignals
-		s.Stop()
-	}()
+	<-killSignals
+	s.Stop()
```
sgnn7: Hmm... why this change? If I understood the old code correctly, we intentionally wanted an async stopping channel processing.

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,11 +50,17 @@ func StartSecretless(params *SecretlessOptions) {
...
 
-	secretless.Start()
+	err = secretless.Start()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	return func() {
+		signal.StopOnExitSignal(secretless)
+	}
```
sgnn7: This patterns seems overly complicated for what it's doing. Why couldn't we just wait forever here rather than in the caller? also, secretless should implement `Stop` already iirc.

test/plugin/example_plugin_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func ReadNLines(reader io.Reader, n int) ([]string, error) {
+	var lines []string
+	reReader := bufio.NewReader(reader)
+	for i := 0; i < n; i++ {
+		line, err := reReader.ReadString('\n')
+		lines = append(lines, line)
```
sgnn7: Maybe we don't care about this in tests but I think if you're appending a lot of lines you might want to go with `string.Builder`.

test/plugin/example_plugin_test.go
```diff
@@ -0,0 +1,75 @@
...
+	log.Printf("Trying to use host: %s", address)
+
+	Convey("can inject information from credentials and proxy connection", t, func() {
+
+		// initial write
+		connection, err := net.Dial("tcp", address)
+		So(err, ShouldBeNil)
+		defer connection.Close()
+		connection.SetWriteDeadline(time.Now().Add(3 * time.Second))
+		connection.SetReadDeadline(time.Now().Add(3 * time.Second))
```
sgnn7: This should probably be a constant up top

test/plugin/example_plugin_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+		_, err = connection.Write([]byte("hello\n"))
+		So(err, ShouldBeNil)
+
+		// initial read
+		lines, err := ReadNLines(connection, 2)
+		So(err, ShouldBeNil)
+		So(
```
sgnn7: I thought we weren't using Convey assertions anymore

test/plugin/secretless_test_harness/main.go
sgnn7: Hmm ... I guess I get that you're extracting the first plugin but `k` is neither short for `plugin` nor `first` so it's hard to extract its relevant meaning in this loop without reading the comment and the logic of how its used. Can we do `p` to make it obvious we're iterating over plugins?

test/plugin/secretless_test_harness/Makefile
jonahx: I think I agree with Srdjan though for a slightly different reason: it adds one more thing a dev working on our codebase is expected to know.  In the past, at this org, we have been _really cavalier_ about the cost of the these the kinds of decisions.  We are still not nearly principled enough about them.  A single dev should _never_ be able to introduce such a dependency without a visible, public discussion.

Imo if this can be rewritten easily enough in bash it should be, even if this is cleaner.  That is, "worse solution using team tech" > "elegant solution using new tech" (as a default -- obviously at some point the scales can tip).

Just as anecdata: I've written plenty of `Makefile`s in my time, but probably not in the last few years.  So I would be googling to understand this.  In this case, it wouldn't take long at all ofc, but that's not really the point.  There will be devs on this project who have _never_ used make.

test/plugin/example_plugin_test.go
jonahx: Why are we using `Convey` in new code?

test/plugin/example_plugin_test.go
```diff
@@ -0,0 +1,78 @@
...
+	return lines, nil
+}
+
+func TestTCPPlugin(t *testing.T) {
+	host := os.Getenv("SECRETLESS_HOST")
+	if host == "" {
+		host = "localhost"
+	}
+
+	address := host + ":6175"
```
sgnn7: https://golang.org/pkg/net/#JoinHostPort

test/plugin/example_plugin_test.go
doodlesbykumbi: We care about the lines individually so I’m not sure if Builder is appropriate for this case

doodlesbykumbi: We care about the lines individually so I’m not sure if Builder is appropriate for this case

test/plugin/example/example_plugin.go
```diff
@@ -0,0 +1,72 @@
...
+//
+// This example connector works as follows:
+// 1. Waits for the initial message from the client
+// 2. Connect to a target service whose address is the value of the secret identified by
+// the key "address"
+// 3. Inject credentials from a secret identified by the key "auth"
+// 4. Write the initial message from the client with some modification
+func connectorFunc(clientConn net.Conn, secrets plugin.SecretsByID) (net.Conn, error) {
+	r := bufio.NewReader(clientConn)
+	res, _, err := r.ReadLine()
```
jonahx: Because the distance from this declaration to its usage below is 14 lines, I would use the name `clientResponse` or something similar.  `r` is fine for reader, otoh, because the span of its usage is exactly 1 line away from the declaration.

test/plugin/example/example_plugin.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, err
+	}
+	_, err = conn.Write([]byte("initial message from client: " + string(res) + "\n"))
+	if err != nil {
+		return nil, err
+	}
+
+	return conn, nil
+}
```
jonahx: I like how simple and clear this function is ☝️

test/plugin/example/example_plugin.go
```diff
@@ -0,0 +1,72 @@
...
+
+// PluginInfo is required as part of the Secretless plugin spec. It provides
+// important metadata about the plugin.
+func PluginInfo() map[string]string {
+	return map[string]string{
+		"pluginAPIVersion": "0.1.0",
+		"type":             "connector.tcp",
+		"id":               "example-tcp-connector",
+		"description": "it's just an example",
```
jonahx: 🤣

test/plugin/example_plugin_test.go
jonahx: I believe what you want here is [`bufio.Scanner`](https://godoc.org/bufio#Scanner)

test/plugin/example_plugin_test.go
jonahx: > This should probably be a constant up top

should it?  not sure there's any benefit to the abstraction here and it might actually harm readability.

test/plugin/secretless_test_harness/Makefile
doodlesbykumbi: I agree in the general sense. `make` was definitely useful to get e2e quick. Maybe for primetime we'd want a more well-though out tool or just scripts. Any-who, this part is now taken out of the PR so that's a conversation for another time.

test/plugin/example_plugin_test.go
doodlesbykumbi: Used `bufio.Scanner`

test/plugin/example_plugin_test.go
doodlesbykumbi: It's useful to break this test into parts. I wonder what y'all think. In this way the connection sort of becomes a constant (if that's what was being talked about). If it was the deadlines then those have been refactored into something even better.

Also the connector interface definitely needs the deadline setting capabilities of net.Conn vs io.ReaderWriter. Quite a bit relies on being able to set deadlines

internal/proxyservice/tcp/proxy_service.go
```diff
@@ -112,17 +118,37 @@ func (proxy *proxyService) handleConnection(clientConn net.Conn) error {
...
+			return nil
+		}
+		return fmt.Errorf("%v connection failed on streaming from client connection: %s", clientConn.LocalAddr(), sourceErr)
+	case destErr := <- destErrChan:
+		if destErr == nil {
+			logger.Infof("%v connection closed by target.\n", clientConn.LocalAddr())
+			return nil
+		}
+		return fmt.Errorf("%v connection failed on streaming from target connection: %s", clientConn.LocalAddr(), destErr)
+	}
```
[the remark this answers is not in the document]
doodlesbykumbi: done

internal/log/mock/logger.go
```diff
@@ -21,6 +21,43 @@ func (l *LoggerMock) Errorf(format string, args ...interface{}) {
...
+func (l *LoggerMock) Error(...interface{}) {}
+// Errorln mocks the method of the same name on the log.Logger interface
+func (l *LoggerMock) Errorln(...interface{}) {}
+// Panic mocks the method of the same name on the log.Logger interface
+func (l *LoggerMock) Panic(...interface{}) {}
+// Panicf mocks the method of the same name on the log.Logger interface
+func (l *LoggerMock) Panicf(string, ...interface{}) {}
+// Panicln mocks the method of the same name on the log.Logger interface
+func (l *LoggerMock) Panicln(...interface{}) {}
+
```
jonahx: Can't you just embed the interface to avoid all this boilerplate.  Than any test using the mock would be over-riding specific methods itself.

doodlesbykumbi: I don't know in advance which methods are being used in a test. If they're not implemented and they get use then that results in a panic. Might as-well have stub methods

jonahx: Hmmm... I'm not so sure.  I think it might be an advantage to force tests to be explicit.  Then again, I'm not sure it's wrong, either, so let's leave it but just keep an eye an that.

test/plugin/example_plugin_test.go
```diff
@@ -0,0 +1,105 @@
...
+	var lines []string
+	scanner := bufio.NewScanner(conn)
+
+	i := 0
+	_ = conn.SetReadDeadline(time.Now().Add(testIODeadline))
+	for i < n && scanner.Scan() {
+		i++
+		lines = append(lines, scanner.Text())
+
+		_ = conn.SetReadDeadline(time.Now().Add(testIODeadline))
```
sgnn7: nit: I know that we're explicitly ignoring the error but I think it's really superfluous for `SetReadDeadline`/`SetWriteDeadline` in this file

PR #900, opened 2019-09-20
853 proxy services

jonahx:
WIP

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -138,6 +138,37 @@ func (logger *Logger) log(severityLevel severity, args ...interface{}) {
 	logger.logln(severityLevel, args...)
 }
 
+// TODO: This duplication is quite hideous, and should be cleaned up by
+//   delegating everything to stdlib logger in a more straightforward way.
+func (logger *Logger) panicf(severityLevel severity, format string, args ...interface{}) {
+	if !logger.shouldPrint(severityLevel) {
+		return
+	}
+
+	if logger.prefix != "" {
+		format = "%s: " + format
+		args = prependString(logger.prefix, args...)
+	}
+
+	logger.BackingLogger.Panicf(format, args...)
+}
+
+func (logger *Logger) panicln(severityLevel severity, args ...interface{}) {
+	if !logger.shouldPrint(severityLevel) {
+		return
+	}
+
+	if logger.prefix != "" {
+		args = prependString(logger.prefix+":", args...)
+	}
+
+	logger.BackingLogger.Panicln(args...)
+}
+
+func (logger *Logger) panic(severityLevel severity, args ...interface{}) {
+	logger.panicln(severityLevel, args...)
+}
+
 // ---------------------------
 // Specific API implementation
 
@@ -163,7 +194,7 @@ func (logger *Logger) Errorf(format string, args ...interface{}) {
 
 // Panicf prints to stdout a formatted panic-level logging message
 func (logger *Logger) Panicf(format string, args ...interface{}) {
-	logger.logf(PanicSeverity, format, args...)
+	logger.panicf(PanicSeverity, format, args...)
 }
 
 // Debugln prints to stdout a debug-level logging message
@@ -188,7 +219,7 @@ func (logger *Logger) Errorln(args ...interface{}) {
 
 // Panicln prints to stdout a panic-level logging message
 func (logger *Logger) Panicln(args ...interface{}) {
-	logger.logln(PanicSeverity, args...)
+	logger.panicln(PanicSeverity, args...)
 }
 
 // Debug prints to stdout a debug-level logging message. Alias of
@@ -218,5 +249,5 @@ func (logger *Logger) Error(args ...interface{}) {
 // Panic prints to stdout a panic-level logging message. Alias of
 // Panicln method.
 func (logger *Logger) Panic(args ...interface{}) {
-	logger.log(PanicSeverity, args...)
+	logger.panic(PanicSeverity, args...)
 }
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -68,21 +68,15 @@ func NewWithOptions(outputBuffer io.Writer, prefix string, isDebug bool) log_api
 }
 
 func (logger *Logger) shouldPrint(severityLevel severity) bool {
-	if !logger.IsDebug && (severityLevel == InfoSeverity || severityLevel == DebugSeverity) {
-		return false
-	}
+	debugOnlySeverity := severityLevel == DebugSeverity ||
+		severityLevel == InfoSeverity
 
-	return true
+	return !debugOnlySeverity || logger.IsDebug
 }
 
 func prependString(prependString string, args ...interface{}) []interface{} {
-	newArgs := make([]interface{}, len(args)+1)
-	newArgs[0] = prependString
-	for idx, val := range args {
-		newArgs[idx+1] = val
-	}
-
-	return newArgs
+	prependSlice := []interface{}{prependString}
+	return append(prependSlice, args...)
 }
 
 // DebugEnabled returns if the debug logging should be displayed for a particular
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -91,6 +91,14 @@ func (logger *Logger) DebugEnabled() bool {
 	return logger.IsDebug
 }
 
+func (logger *Logger) CopyWith(prefix string, isDebug bool) log_api.Logger {
+	return NewWithOptions(
+		logger.BackingLogger.Writer(),
+		prefix,
+		isDebug,
+	)
+}
+
 // Prefix returns the prefix that will be prepended to all output messages
 func (logger *Logger) Prefix() string {
 	return logger.prefix
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -91,6 +91,8 @@ func (logger *Logger) DebugEnabled() bool {
 	return logger.IsDebug
 }
 
+// CopyWith creates a copy of the logger with the prefix and debug values
+// overridden by the arguments.
 func (logger *Logger) CopyWith(prefix string, isDebug bool) log_api.Logger {
 	return NewWithOptions(
 		logger.BackingLogger.Writer(),
--- a/internal/log/log_test.go
+++ b/internal/log/log_test.go
@@ -65,7 +65,6 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
 		"Infof": lt.logger.Infof,
 		"Warnf": lt.logger.Warnf,
 		"Errorf": lt.logger.Errorf,
-		"Panicf": lt.logger.Panicf,
 	} {
 		lt.ResetBuffer()
 		t.Run(
@@ -87,8 +86,6 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
 		"Warnln":  lt.logger.Warnln,
 		"Error":   lt.logger.Error,
 		"Errorln": lt.logger.Errorln,
-		"Panic":   lt.logger.Panic,
-		"Panicln": lt.logger.Panicln,
 	} {
 		lt.ResetBuffer()
 		t.Run(
@@ -96,6 +93,49 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
 			func(t *testing.T) {
 				method(testCaseArgs...)
 				assert.Regexp(t, lt.expectedOutput(methodName), lt.CurrentOutput())
+			},
+		)
+	}
+
+	// Formatted panic: panicf
+	lt.ResetBuffer()
+	t.Run(
+		lt.descriptionForTest("panicf"),
+		func(t *testing.T) {
+			defer func() {
+				paniced := false
+				if r := recover(); r != nil {
+					paniced = true
+				}
+				if !paniced {
+					assert.FailNow(t, "Should have paniced but didn't")
+				}
+				assert.Regexp(t, lt.expectedOutput("panicf"), lt.CurrentOutput())
+			}()
+			lt.logger.Panicf(testCaseFormatStr, testCaseArgs...)
+		},
+	)
+
+	// Unformatted panic methods
+	for methodName, method := range map[string]logMethod{
+		"Panic":   lt.logger.Panic,
+		"Panicln": lt.logger.Panicln,
+	} {
+		lt.ResetBuffer()
+		t.Run(
+			lt.descriptionForTest(methodName),
+			func(t *testing.T) {
+				defer func() {
+					paniced := false
+					if r := recover(); r != nil {
+						paniced = true
+					}
+					if !paniced {
+						assert.FailNow(t, "Should have paniced but didn't")
+					}
+					assert.Regexp(t, lt.expectedOutput(methodName), lt.CurrentOutput())
+				}()
+				method(testCaseArgs...)
 			},
 		)
 	}
--- a/internal/log/log_test.go
+++ b/internal/log/log_test.go
@@ -103,12 +103,12 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
 		lt.descriptionForTest("panicf"),
 		func(t *testing.T) {
 			defer func() {
-				paniced := false
+				panicked := false
 				if r := recover(); r != nil {
-					paniced = true
+					panicked = true
 				}
-				if !paniced {
-					assert.FailNow(t, "Should have paniced but didn't")
+				if !panicked {
+					assert.FailNow(t, "Should have panicked but didn't")
 				}
 				assert.Regexp(t, lt.expectedOutput("panicf"), lt.CurrentOutput())
 			}()
@@ -126,12 +126,12 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
 			lt.descriptionForTest(methodName),
 			func(t *testing.T) {
 				defer func() {
-					paniced := false
+					panicked := false
 					if r := recover(); r != nil {
-						paniced = true
+						panicked = true
 					}
-					if !paniced {
-						assert.FailNow(t, "Should have paniced but didn't")
+					if !panicked {
+						assert.FailNow(t, "Should have panicked but didn't")
 					}
 					assert.Regexp(t, lt.expectedOutput(methodName), lt.CurrentOutput())
 				}()
--- a/internal/proxyservice/proxy_service.go
+++ b/internal/proxyservice/proxy_service.go
@@ -24,26 +24,6 @@ type proxyServices struct {
 	runningServices []secretless.Service
 }
 
-type connectorResources struct {
-	logger logapi.Logger
-	config []byte
-}
-
-func (cr *connectorResources) Logger() logapi.Logger {
-	return cr.logger
-}
-
-func (cr *connectorResources) Config() []byte {
-	return cr.config
-}
-
-func newConnectorResources(l logapi.Logger, cfg []byte) connector.Resources {
-	return &connectorResources{
-		logger: l,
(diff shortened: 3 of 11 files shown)


internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,147 @@
...
 }
 
-// TODO: Rename to Call or Run and return a Stopper instead of having Stop()
-// Start starts all proxy services
-func (s *proxyServices) Start() {
-	// TODO: Implement
+type connectorResources struct {
+	logger logapi.Logger
+	config []byte
+}
```
sgnn7: Code duplication. Why not just make [this](https://github.com/cyberark/secretless-broker/blob/master/pkg/secretless/plugin/connector/resources.go#L19-L30) a public struct instead?

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,147 @@
...
-	// 1. Rewrap the Logger with service name prefix
-	// 2. Create the ConnectorResources object
-	// etc...
+// Start starts all proxy services
+func (s *proxyServices) Start() error {
+	for _, svc := range s.servicesToStart() {
+		err := svc.Start()
+		if err != nil {
+			s.logger.Errorf("could not start proxy service: %s", err)
+			continue
```
sgnn7: I think errors in starting services are fatal - we probably don't want half-working broker running.

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,147 @@
...
 }
 
 // Stop stops all proxy services
-func (s *proxyServices) Stop() {
-	// Stop all the Service
+func (s *proxyServices) Stop() error {
+	for _, svc := range s.runningServices {
+		err := svc.Stop()
+		if err != nil {
+			s.logger.Errorf("could not stop proxy service: %s", err)
```
sgnn7: We now return `error` but we don't propagate it from the inner block. Stop should probably return one or the other. Maybe the return type should be `[]error`?

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,147 @@
...
+	tcpPlugins := s.availPlugins.TCPPlugins()
+	// httpPlugins := s.availPlugins.HTTPPlugins()
+
+	for _, svc := range s.config.Services {
+		requestedPlugin := svc.Connector //TODO: this rename is a name smell
+
+		// first check the available TCP Plugins
+		tcpPlugin, found := tcpPlugins[requestedPlugin]
+		if found {
+			if tcpSvc := s.createTCPService(svc, tcpPlugin); tcpSvc != nil {
```
sgnn7: This probably should return an `error` instead of us relying on the `nil` as the error condtion

sgnn7: Error here should probably be a fatal error

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,147 @@
...
+			continue
+		}
+
+		// TODO: next check available HTTP Plugins
+		// httpPlugin, found := httpPlugins[requestedPlugin]
+
+		// TODO: Deal with SSH in a hardcoded way
+
+		// Default case: not found
+		s.logger.Errorf("plugin '%s' not available.", requestedPlugin)
```
sgnn7: This should probably also be a fatal error

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,147 @@
...
+	svc *v2.Service,
+	plugin tcp.Plugin,
+) secretless.Service {
+
+	//TODO: Add validation somewhere about overlapping listenOns
+	listener, err := net.Listen("tcp", strings.TrimLeft(svc.ListenOn, "tcp://"))
+	if err != nil {
+		// TODO: Should we do more than this?
+		s.logger.Errorf("could not create listener on: %s", svc.ListenOn)
+		return nil
```
sgnn7: This condition is most likely a terminal one too. As mentioned before, we should return this error as well.

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,147 @@
...
+
+	newSvc, err := tcpproxy.NewProxyService(
+		connector_,
+		listener,
+		svcLogger,
+		credsRetriever,
+	)
+
+	if err != nil {
+		// TODO: Should we do more than this?
```
sgnn7: Yes, this is also a terminal error I think

internal/proxyservice/proxy_service.go
```diff
@@ -53,3 +157,16 @@ func NewProxyServices(
...
+	providerFactories := make(map[string]func(v1.ProviderOptions) (v1.Provider, error))
+
+	for providerID, providerFactory := range internal.InternalProviders {
+		providerFactories[providerID] = providerFactory
+	}
+
+	resolver := plugin.NewResolver(providerFactories, nil, nil)
+
+	return resolver.Resolve(secrets)
+}
```
sgnn7: Not a blocker but this recreates the whole thing for every plugin. We should probably try to figure out how to optimize this by caching it.

internal/signal/signal.go
```diff
@@ -42,9 +42,6 @@ func newHaltSignalChan() chan os.Signal {
 //   the signal package.  Not ideal.
 func StopOnExitSignal(s Stopper) {
 	killSignals := newHaltSignalChan()
-
-	go func() {
-		<-killSignals
-		s.Stop()
-	}()
+	<-killSignals
+	s.Stop()
```
sgnn7: I'm still unclear why we no longer do async handling of this logic

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -90,6 +94,8 @@ func handlePerformanceProfiling(profileType string) {
 
 	// Start profiling
 	perfProfile := profile.New(profileType)
-	signal.StopOnExitSignal(perfProfile)
+
+	// Wrap the exit signal handler in go routine so it won't block
+	go func() { signal.StopOnExitSignal(perfProfile) }()
```
sgnn7: nit: I know the codeblock is tiny but this might look better split into multiple lines

internal/log/log.go
```diff
@@ -91,6 +91,14 @@ func (logger *Logger) DebugEnabled() bool {
 	return logger.IsDebug
 }
 
+func (logger *Logger) CopyWith(prefix string, isDebug bool) log_api.Logger {
```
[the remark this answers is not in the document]
jonahx: Yes, but the problem is it forces you to use `defaultOutputBuffer`, which is stdout.  Therefore, if we used that instead of `CopyWith` here, it would screw up the forthcoming unit tests, where we'll want to make assertions and a mocked log writer.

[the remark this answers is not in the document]
sgnn7: Hmm I see what you mean though I'm a bit curious why not go with something like `NewForServiceWithOptions` to match the other patterns we have here

internal/proxyservice/proxy_service.go
jonahx: So yeah, there's some thinking to be done here... if we return a custom error type then it won't be implementing `secretless.Service` anymore, but maybe that doesn't matter anyway.  I feel like the whole idea of the "generic service" with `Start` / `Stop` defined by `secretless.Service` may be not ideal, but it's sort of a can of worms I don't want to open atm (though I think that design decision should be revisited).

For the time being, how about I collect a list of the services that fail to stop, and return a single error whose message contains all of them?

internal/proxyservice/proxy_service.go
jonahx: yeah, i'll just go ahead and make this fatal.

internal/proxyservice/proxy_service.go
jonahx: Since `servicesToStart()` will take an error as being fatal, i'm leaving the panic-ing up to him, and changing this lower-level method to just return the error.  That way we won't duplicate the logic.

internal/proxyservice/proxy_service.go
jonahx: Addressed by solution to previous comment

internal/proxyservice/proxy_service.go
jonahx: Agreed.   The whole `GetSecrets` solution here is just temporary to get e2e.  It should be getting injected anyway.  But really as you said we should just revisit it from first principles.  Will add TODO though.

internal/signal/signal.go
jonahx: Quoting from the clean history of my upcoming push :)

```
   Make StopOnExitSignal synchronous
    
    This will allow clients to decide if they want it to run async or not.
    Adding async is as simple as wrapping the call in a go routine.  Whereas
    it's useful to be able to block if we want to do, as we do in the main
    entrypoint method where Secretless needs to block after starting up all
    its services.
```

I'm open to doing this differently though, if you have a better solution.  Seems like something else we can think about and improve later if need be.  Personally I think the current solution is ok, but again am open to hearing reasons why it's not.

internal/log/log.go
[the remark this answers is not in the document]
jonahx: That would be a totally viable alternative.  We'd have to make the backing writer accessible.  

I still prefer the semantics of the above, though, since it imposes less burden on the client code.  That is, in this particular case, it's one less thing the client code has to destructure and put back into the ctor.  Beyond that (and this is not too important here but is worth thinking aobut in general), the copy semantics make the clients totally protected from future changes.  What we're saying, by using the copy method, is: "Change this and that, and keep everything else the same," and that everything else could be future things which don't even exist yet.  If you think about it, these semantics are exactly what you want in this case.

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,141 @@
...
-	// 2. Create the ConnectorResources object
-	// etc...
+func (s *proxyServices) Start() error {
+	for _, svc := range s.servicesToStart() {
+		err := svc.Start()
+		// Showstopper: Failure to start a service is fatal.
+		if err != nil {
+			// TODO: Upgrade our logger so we can use Fatalf here
+			s.logger.Panicf("could not start proxy service: %s", err)
+			continue
```
sgnn7: If we panic out, `continue` is unreachable

internal/proxyservice/proxy_service.go
sgnn7: TODO missing

internal/signal/signal.go
sgnn7: So after thinking about this a while, I think the problems here if we move to a sync method are:
- there is most likely no longer a need for this method at all since the caller can just call `Stop()` right after
- there's two places that we use this in: one needs to wait (secretless) and one doesn't (entrypoint). we're catering to a one-off consumer with the sync logic. we can easily just add a "block indefinitely" block right after calling an async version in Secretless.
- if we do choose to leave it sync, it's no longer matching its name `StopOnExitSignal` - it would probably be something like `BlockUntilHalt`

internal/log/log.go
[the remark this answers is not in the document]
sgnn7: Eh I'm not 100% on this being needed but not a blocker for now.

sgnn7:
@jonahx PS: make sure to add a story for enumerating whatever will need testing still - I see some new files covered that have no test issues associated.

internal/signal/signal.go
jonahx: I went ahead and reverted `StopOnExitSignal` to being async... I think you're right.  It makes the client code cleaner and why not have it.

To deal with the blocking issue, I made `NewExitSignalChan` public, so we can use it like this after starting Secretless:

```go
	// Block until we receive an exit signal
	<- signal.NewExitSignalChan()
```

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,140 @@
...
+		}
+
+		// TODO: next check available HTTP Plugins
+		// httpPlugin, found := httpPlugins[requestedPlugin]
+
+		// TODO: Deal with SSH in a hardcoded way
+
+		// Default case: not found
+		// TODO: Add Fatalf to our logger and use that
+		s.logger.Panicf("plugin '%s' not available.", requestedPlugin)
```
doodlesbykumbi: in that case there should be a `continue` for the found case

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -48,13 +48,19 @@ func StartSecretless(params *SecretlessOptions) {
...
+	allServices := proxyservice.NewProxyServices(cfg, availPlugins, logger, evtNotifier)
+	signal.StopOnExitSignal(allServices)
+
+	err = allServices.Start()
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	// Block until we receive an exit signal
+	<- signal.NewExitSignalChan()
```
doodlesbykumbi: There's a race condition here. The service cleanup will never happen because once line 63 is hit the program will terminate. We'll need some way to coordinate multiple consumers of these exit signals.

internal/proxyservice/proxy_service.go
jonahx: The panic is effectively fatal (see the TODO above) because it is never `recover`ed.  It actually had a `continue` before and I removed it in response to Srdjan's "this is unreachable" comment lol.

doodlesbykumbi: The issue is that the block at line 73 is executed because the tcp plugin is found but on line 80 we don't `continue` instead we go back to the parent block and print `"plugin '%s' not available."` even though the plugin is available.

jonahx: Ah, sorry I read your initial comment too fast.  Thanks!

pkg/secretless/entrypoint/entrypoint.go
jonahx: Rewrote the signal package entirely, should be fixed now.

internal/log/log_test.go
```diff
@@ -87,15 +86,56 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
...
+		)
+	}
+
+	// Formatted panic: panicf
+	lt.ResetBuffer()
+	t.Run(
+		lt.descriptionForTest("panicf"),
+		func(t *testing.T) {
+			defer func() {
+				paniced := false
```
sgnn7: s/paniced/panicked

internal/log/log_test.go
```diff
@@ -87,15 +86,56 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
...
+	for methodName, method := range map[string]logMethod{
 		"Panic":   lt.logger.Panic,
 		"Panicln": lt.logger.Panicln,
 	} {
 		lt.ResetBuffer()
 		t.Run(
 			lt.descriptionForTest(methodName),
 			func(t *testing.T) {
+				defer func() {
+					paniced := false
```
sgnn7: s/paniced/panicked

internal/proxyservice/proxy_service.go
```diff
@@ -1,43 +1,140 @@
...
 type proxyServices struct {
-	config        v2.Config
-	logger        log.Logger
-	eventNotifier v1.EventNotifier
-	availPlugins  plugin.AvailablePlugins
+	config          v2.Config
+	logger          logapi.Logger
+	eventNotifier   v1.EventNotifier
+	availPlugins    plugin.AvailablePlugins
+	runningServices []secretless.Service
```
sgnn7: nit: this list should probably be alphabetized

internal/log/log.go
```diff
@@ -128,6 +132,37 @@ func (logger *Logger) log(severityLevel severity, args ...interface{}) {
 	logger.logln(severityLevel, args...)
 }
 
+// TODO: This duplication is quite hideous, and should be cleaned up by
+//   delegating everything to stdlib logger in a more straightforward way.
```
sgnn7: Sidenote but we could also just pull in a structured logger too to fix this

internal/signal/signal.go
```diff
@@ -21,30 +18,56 @@ var exitSignals = []os.Signal{
...
-type Stopper interface {
-	Stop()
+// Handler is a simply a function intended to be called in response to a signal.
+type Handler func()
+
+// Exit reifies the idea of "an exit signal" so that such an exit is smart: It
+// allows you to add handlers that will be called when it occurs (AddHandler),
+// and to block until it does occur (Await).
+type Exit interface {
+	AddHandler(Handler)
```
sgnn7: This name isn't descriptive enough - it will be invoked as `signal.AddHandler()` and doesn't describe well enough what signal and on what event it will be triggered.

internal/log/log.go
jonahx: Which part?  The fundamental problem as I see it now (I mentioned this after we merged the log stuff) is that we tested it wrong.  We're testing it as if it's a black box / greenfield project.  Whereas in fact  _our_ logger is just a thin wrapper over another (already reliable) logger, to which we delegate all important work.  Our tests should reflect this by simply asserting that a mock was called at the appropriate times, and if we write them in this way they'll be much simpler.

The only piece using a different underlying logger might help with is that it might put prefix is the place we want it by default, so we wouldn't have to rewrite the prefixing logic as we've done.  

That said, I'm not against if we have a good reason.  My position was just that we shouldn't bring in deps because they have features _we might want_ in the future.  Bringing in a dep to solve a specific, current need is fine.

Btw, we could also just take the stdlib's logger default way of printing prefixes.  I personally would be fine with that.

Anyway would like to hear more of your reasoning...

internal/signal/signal.go
jonahx: The current code looks like this:

```go
exitSignal := signal.NewExit()
exitSignal.AddHandler(...)
```
I think this looks good and is in line with Go conventions.  Ie, use the package name as a descriptive modifier so your package types and funcs don't stutter, and then correctly name your local variables to capture the context.

sgnn7:
@jonahx Just noticed 22 commits too - the history will need a cleanup a bit or we'll have to do a squash-merge

internal/signal/signal.go
```diff
@@ -21,30 +18,56 @@ var exitSignals = []os.Signal{
...
+	Await()
```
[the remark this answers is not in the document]
jonahx: Even though I think `Await` reads more naturally in a vacuum, I am persuaded by this consistency argument.

internal/signal/signal.go
```diff
@@ -21,30 +18,56 @@ var exitSignals = []os.Signal{
...
+// Await does two things: 1. It kicks off the "listening" process, so that
+// Handlers will be notified of an exit. 2.  Blocks until an exit occurs.
+func (p *exit) Await() {
 	go func() {
-		<-killSignals
-		s.Stop()
+		<-p.exitSignalChannel
+
+		for _, sub := range p.handlers {
+			sub()
```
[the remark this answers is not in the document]
jonahx: Totally.  That was a carryover name from the old code.

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -90,6 +106,12 @@ func handlePerformanceProfiling(profileType string) {
 
 	// Start profiling
 	perfProfile := profile.New(profileType)
-	signal.StopOnExitSignal(perfProfile)
-	perfProfile.Start()
+	exitSignals.AddHandler(func() {
+		_ = perfProfile.Stop()
```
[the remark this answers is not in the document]
jonahx: Don't need, but you get IDE warnings without out and I think there's value in saying, "I, as the programmer, did not forget there was an error.  I know an error is returned, and I am purposefully ignoring it."

PR #905, opened 2019-09-25
Update tests to do a better job of cleaning up after exit

izgeri:
We've noticed two problems with the tests:
- the tests running in GKE don't always delete the namespaces that are created on exit
- the integration tests are experiencing collisions where one test removes a container running for another test

This PR updates so that all GKE namespaces will be removed on test exit, and all tests will run in their own `COMPOSE_PROJECT_NAME` so they won't have mid-air collisions

ETA: I also updated the `conjur` integration tests because they were failing on the `latest` branch - it turns out that this had already been fixed by someone (yay!) which I noticed when I cherry-picked my changes from `test-branch` to here, but I left my changes bc I think they further simplify things

You can see these changes passing on top of the `latest` branch here: https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--secretless-broker/detail/test-branch/2/pipeline/36

The `test-branch` branch can be deleted once this is merged.

### DIFF
--- a/k8s-ci/test
+++ b/k8s-ci/test
@@ -20,12 +20,11 @@ docker inspect secretless-broker:latest > /dev/null 2>&1 || {
 }
 
 # Set environment variables
-export UNIQUE_TEST_ID
-UNIQUE_TEST_ID="k8s-test-$(uuidgen | tr "[:upper:]" "[:lower:]" | head -c 12 | tr -d -)"
+export TEST_ID="$(uuidgen | tr "[:upper:]" "[:lower:]" | head -c 12 | tr -d -)"
 export SECRETLESS_IMAGE_NAME="${DOCKER_REGISTRY_PATH}/secretless-broker"
-export SECRETLESS_IMAGE_TAG="${UNIQUE_TEST_ID}"
-
-export TEST_NAMESPACE="secretless-${UNIQUE_TEST_ID}"
+export SECRETLESS_IMAGE_TAG="${TEST_ID}"
+export SECRETLESS_IMAGE="${SECRETLESS_IMAGE_NAME}:${SECRETLESS_IMAGE_TAG}";
+export TEST_NAMESPACE="test-secretless-${TEST_ID}"
 
 # Clean up when script completes
 function finish {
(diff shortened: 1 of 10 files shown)


izgeri:
Note that the GKE tests ran and did not leave behind any trailing namespaces:
```
$ kns
conjur-5-1fc976dd-a-test
default
kube-public
kube-system
```
(the `conjur-5` is from another repo - we don't create a ns of that form)

izgeri:
Note that I'm not sure if these tests require other changes - the ssh handler test dies on the error `ssh: Could not resolve hostname secretless: Name does not resolve`, which _might_ mean the hostname needs to include the COMPOSE_PROJECT_NAME somehow (my rust is showing here) - but I can't play around with it easily bc of the broken master branch.

izgeri:
In `test-branch` I cherry-picked these commits onto the `latest` branch to see if they would pass

k8s-ci/test
```diff
@@ -34,9 +34,11 @@ function finish {
   # Delete registry images that were used
   deleteRegistryImage "${SECRETLESS_IMAGE_NAME}" "${SECRETLESS_IMAGE_TAG}"
 
-  # Delete test namespace
+  # Delete test namespace for CRD tests and app/backend namespaces for demo test
```
doodlesbykumbi: This test script is intended to be a generic way to run k8s tests. I think there is stuff left over as part of the conversion from the script only serving the CRD test case.

1. The changeset in this file does cleanup the app/backend namespaces for the demo test. It's doing so by giving the responsibility to the generic script. Ideally this should happen inside the k8s-demo test script. The logic for it does exist inside the k8s-demo test. It's just borked for some reason - https://github.com/cyberark/secretless-broker/blob/master/demos/k8s-demo/stop. 

2. I think we should remove all mentions of CRD in this script.
`SECRETLESS_CRD_SUFFIX` should be passed down to the CRD test script as `TEST_ID`, then we can change that script to receive it as such.

doodlesbykumbi: Here are the changes I had in mind. https://github.com/cyberark/secretless-broker/compare/test-branch...doodlesbykumbi-patch-1?expand=1

izgeri:
Ok @doodlesbykumbi I think I've finally gotten this to a good place 😅 

to see the changes passing when run against `latest` you can look at the [test branch job](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/test-branch/)

the [test branch diff](https://github.com/cyberark/secretless-broker/compare/test-branch) is effectively the same set of changes as this branch. I've incorporated your changes, too.

k8s-ci/test
```diff
@@ -49,19 +48,17 @@ function runTest() {
...
 kubens '${TEST_NAMESPACE}';
 
-# run CRD tests
-cd $test_path;
-export SECRETLESS_IMAGE_NAME='${SECRETLESS_IMAGE_NAME}';
-export SECRETLESS_IMAGE_TAG='${SECRETLESS_IMAGE_TAG}';
+export TEST_ID='${TEST_ID}'
+
+# run tests
+cd k8s-ci/k8s_crds;
```
doodlesbykumbi: @izgeri not sure what happened here but i think this should be dynamic

```suggestion
cd "$test_path";
```

izgeri:
OKAY I was able to run the k8s-demo and CRD tests using the `summon` commands from the Jenkinsfile in this branch, so those will work as expected.

To be clear "expected" means the CRD tests pass but the demo pass fails on
```
Waiting for application to boot up
(This may take more than 1 minute)...
If you don't see a command prompt, try pressing enter.
.........
Application failed to boot up after 10 pings at 2 second intervals.
pod "alpine-curl" deleted
```

I have verified the namespaces were cleaned up afterward for both tests.

The Conjur tests passing can be validated somewhat by the green build in the `test-branch` branch, but won't be fully validated until the Conjur HTTP handler is e2e.

PR #908, opened 2019-09-26
906 http proxy service

jonahx:
History is clean: Use "Commits" tab to review.

Also note: None of the CC errors are new code.  They were all introduced in the final comment when I moved stuff around to fix a circular dep problem.  And almost all of them will be fixed when we delete the old code.

### DIFF
--- a/internal/plugin/manager/manager.go
+++ b/internal/plugin/manager/manager.go
@@ -17,7 +17,7 @@ import (
 	"time"
 
 	plugin2 "github.com/cyberark/secretless-broker/internal/plugin"
-	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/so"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/sharedobj"
 	"github.com/pkg/profile"
 	"gopkg.in/yaml.v2"
 
@@ -410,7 +410,7 @@ func (manager *Manager) LoadLibraryPlugins(path string, checksumsFile string) er
 		// We override file listing if we did a verification to prevent additions
 		// to plugins between verification and loading the plugins.
 		var err error
-		if files, err = so.VerifyPluginChecksums(path, checksumsFile); err != nil {
+		if files, err = sharedobj.VerifyPluginChecksums(path, checksumsFile); err != nil {
 			log.Fatalln(err)
 		}
 	} else if len(files) > 0 {
--- a/internal/plugin/manager/manager.go
+++ b/internal/plugin/manager/manager.go
@@ -1,4 +1,4 @@
-package plugin
+package manager
 
 import (
 	"encoding/json"
@@ -16,6 +16,8 @@ import (
 	"syscall"
 	"time"
 
+	plugin2 "github.com/cyberark/secretless-broker/internal/plugin"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/so"
 	"github.com/pkg/profile"
 	"gopkg.in/yaml.v2"
 
@@ -338,7 +340,7 @@ func (manager *Manager) Run(configManagerID string, configManagerSpec string) er
 	log.Println("Initialization of plugins done.")
 
 	log.Println("Initializing the proxy...")
-	resolver := NewResolver(manager.ProviderFactories, manager, nil)
+	resolver := plugin2.NewResolver(manager.ProviderFactories, manager, nil)
 
 	manager.Proxy = secretless.Proxy{
 		Config:          configuration,
@@ -408,7 +410,7 @@ func (manager *Manager) LoadLibraryPlugins(path string, checksumsFile string) er
 		// We override file listing if we did a verification to prevent additions
 		// to plugins between verification and loading the plugins.
 		var err error
-		if files, err = VerifyPluginChecksums(path, checksumsFile); err != nil {
+		if files, err = so.VerifyPluginChecksums(path, checksumsFile); err != nil {
 			log.Fatalln(err)
 		}
 	} else if len(files) > 0 {
--- a/internal/proxyservice/proxy_service.go
+++ b/internal/proxyservice/proxy_service.go
@@ -15,7 +15,8 @@ To put this in context, the complete high-level flow is:
 	“subservices” associated with it.
 
 	4. Each of those subservices needs two things: a connector (which knows how
-	authenticate requests) and a way to get the current credentials at runtime.
+	to authenticate requests) and a way to get the current credentials at
+	runtime.
 
 	5. Now note the signature of the connector itself just looks like this:
 
@@ -187,6 +188,7 @@ func (s *proxyServices) createHTTPService(
 	proxyName := httpSvcCfg.Name()
 	svcLogger := s.loggerFor(proxyName)
 
+	// TODO: NewHTTPProxyFunc needs to be injected
 	newSvc, err := httpproxy.NewProxyService(subservices, listener, svcLogger)
 	if err != nil {
 		s.logger.Errorf("could not create http proxy service '%s'", proxyName)
@@ -214,6 +216,7 @@ func (s *proxyServices) createTCPService(
 	svcConnector := pluginInst.NewConnector(connResources)
 	credsRetriever := s.credsRetriever(config.Credentials)
 
+	// TODO: NewTCPProxyFunc needs to be injected
 	newSvc, err := tcpproxy.NewProxyService(
 		svcConnector,
 		listener,
--- a/internal/proxyservice/proxy_service.go
+++ b/internal/proxyservice/proxy_service.go
@@ -1,56 +1,3 @@
-/*
-Package proxyservice takes a Secretless configuration and available plugins
-and constructs the requires ProxyServices that Secretless will run.
-
-To put this in context, the complete high-level flow is:
-
-	1. Parse the `secretless.yml` into individual “service” configs,
-	corresponding to each service entry in the yml.
-
-	2. Identify http services that share a `listenOn`, because we know those
-	will all be part of a single “http proxy service” that uses traffic routing
-	within it to delegate to the subservice connectors.
-
-	3. So now we have have the http service's `listenOn` and all the
-	“subservices” associated with it.
-
-	4. Each of those subservices needs two things: a connector (which knows how
-	to authenticate requests) and a way to get the current credentials at
-	runtime.
-
-	5. Now note the signature of the connector itself just looks like this:
-
-		type Connector func(request *http.Request, secrets plugin.SecretsByID) error
-
-	6. So it is the responsibility of the proxy service to actually fetch the
-	credentials.  So what does the proxy service need for each of those
-	subservices? Precisely this:
-
-		type HTTPSubService struct {
-		  connector http.Connector,
-		  retrieveCredentials internal.CredentialsRetriever,
-		}
-
-	7. Putting all the together, here’s what we need to construct a new http
-	proxy service:
-
-		func NewProxyService(
-			subservices []HTTPSubService,
-			sharedListener net.Listener,
-			logger log.Logger,
-		) (internal.Service, error) {
-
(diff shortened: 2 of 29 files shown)


internal/proxyservice/proxy_service.go
```diff
@@ -57,78 +59,246 @@ func (s *proxyServices) Stop() error {
...
+			// we won't be unit testing this.  Instead, we'll make the change
+			// in the TODO_ above and this code will be deleted then.
+			log.Fatalln("Nil configuration is not allowed!")
+		}
+		rawConfigs = append(rawConfigs, *cfg)
+	}
+
+	// Separate HTTP and TCP configs
+	var httpConfigs []v2.Service
+	var	tcpConfigs []v2.Service
```
sgnn7: nit: Spacing

internal/proxyservice/proxy_service.go
```diff
@@ -57,78 +59,246 @@ func (s *proxyServices) Stop() error {
...
+
+	// TODO: Add proper validation here of the type.  This should moved into
+	//   IsHTTPPlugin, whose API will likely change to returning a type or an
+	//   error
+	for _, cfg := range rawConfigs {
+		if plugin.IsHTTPPlugin(availPlugins, cfg.Connector) {
+			httpConfigs = append(httpConfigs, cfg)
```
sgnn7: It seems that we are leaking implementation details up to the level where they might not need to be visible - I think the knowledge of how to squash and parse http service configs probably needs to be in some `.... .http ...` package since the `proxyservice` should ideally not need to know anything other than how to start/stop services

internal/proxyservice/proxy_service.go
```diff
@@ -57,78 +59,246 @@ func (s *proxyServices) Stop() error {
...
+
+	// Now create proper HTTPServiceConfig objects from our map
+	var httpServiceConfigs []HTTPServiceConfig
+	for listenOn, configs := range httpByListenOn {
+		httpServiceConfig := HTTPServiceConfig{
+			SharedListenOn:    listenOn,
+			SubserviceConfigs: configs,
+		}
+		httpServiceConfigs = append(httpServiceConfigs, httpServiceConfig)
+	}
```
sgnn7: This is somewhat of a hefty method to parse through in terms of logic - any way to split this up into some smaller functions?

internal/proxyservice/proxy_service.go
```diff
@@ -57,78 +59,246 @@ func (s *proxyServices) Stop() error {
...
+	proxyName := fmt.Sprintf("HTTP Proxy on %s", httpSvcConfig.SharedListenOn)
+	svcLogger := s.logger.CopyWith(proxyName, s.logger.DebugEnabled())
+
+	newSvc, err := httpproxy.NewProxyService(subservices, listener, svcLogger)
+	if err != nil {
+		s.logger.Errorf("could not create http proxy service '%s'", proxyName)
+		return nil, err
+	}
+	return newSvc, nil
+}
```
sgnn7: I'm guessing that this method's implementation (or http-specific parts) might be a better fit for its own module/package to split concerns

internal/proxyservice/proxy_service.go
jonahx: This is exactly right.  You'll note the comment at the top of the method :)

> // TODO: Eventually it should get moved to the config package, and this
> //   transformation should occur at the border.

internal/proxyservice/proxy_service.go
jonahx: Definitely.  I'll plan to do this Monday.

internal/proxyservice/proxy_service.go
jonahx: I'll think about this some more, but do keep in mind in that:
```go
	newSvc, err := httpproxy.NewProxyService(subservices, listener, svcLogger)
```
which is the punchline of this method is already split out, and that's where the http specific logic will this.  This method you're commenting on here is only responsible for putting the arguments together that get passed down to the real proxy service.  So it's fairly cohesive as is.  That said, it could probably be improved further.

internal/proxyservice/proxy_service.go
```diff
@@ -57,89 +111,145 @@ func (s *proxyServices) Stop() error {
...
-			servicesToStart = append(servicesToStart, tcpSvc)
-			continue
+
+	// TCP Plugins
+	for _, cfg := range configsByType.TCP {
+		// Validation will have already happened
+		tcpSvc, err := s.createTCPService(cfg, tcpPlugins[cfg.Connector])
+		if err != nil {
+			// TODO: Add Fatalf to our logger and use that
+			s.logger.Panicf("unable to create TCP service '%s'", cfg.Name)
```
doodlesbykumbi: we should append the error to the panic message so the user has additional information as to the cause

pkg/secretless/config/v2/config.go
```diff
@@ -47,3 +51,115 @@ func NewConfig(v2YAML []byte) (*Config, error) {
...
+) ConfigsByType {
+
+	// Get the nil checks out of the way.
+	var rawConfigs []Service
+	for _, cfg := range uncheckedConfigs {
+		if cfg == nil {
+			// Hard-coding log here is okay since nils should never occur and
+			// we won't be unit testing this.  Instead, we'll make the change
+			// in the TODO_ above and this code will be deleted then.
+			log.Fatalln("Nil configuration is not allowed!")
```
doodlesbykumbi: This is a convenience function called quite deep in the stack. I don't think it should have this catastrophic side effect. I think that should bubble up to a place where it's more obvious what the issue.

internal/proxyservice/proxy_service.go
```diff
@@ -57,89 +111,145 @@ func (s *proxyServices) Stop() error {
...
+func (s *proxyServices) connectorResources(svc v2.Service) connector.Resources {
+	svcLogger := s.loggerFor(svc.Name)
+	return connector.NewResources(svc.ConnectorConfig, svcLogger)
+}
+
+func (s *proxyServices) loggerFor(name string) logapi.Logger {
+	return s.logger.CopyWith(name, s.logger.DebugEnabled())
+}
+
+func (s *proxyServices) credsRetriever(
```
doodlesbykumbi: proxyServices seems super busy. there's a lot of responsibility here. nothing comes to mind immediately for how to break it up but just wanted to bring that up. i'll be thinking about it ...

internal/plugin/manager/manager.go
```diff
@@ -16,6 +16,8 @@ import (
 	"syscall"
 	"time"
 
+	plugin2 "github.com/cyberark/secretless-broker/internal/plugin"
```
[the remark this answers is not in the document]
jonahx: This file will be removed.  This was just a side-effect of some IDE "Move to..." refactoring.  If it's a problem in the code we're keeping let's address that separately.

internal/plugin/manager/manager.go
```diff
@@ -408,7 +410,7 @@ func (manager *Manager) LoadLibraryPlugins(path string, checksumsFile string) er
 		// We override file listing if we did a verification to prevent additions
 		// to plugins between verification and loading the plugins.
 		var err error
-		if files, err = VerifyPluginChecksums(path, checksumsFile); err != nil {
+		if files, err = so.VerifyPluginChecksums(path, checksumsFile); err != nil {
```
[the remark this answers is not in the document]
jonahx: Yeah I feel you.  Go does recommend short names though... how about `sharedobj` as compromise?

internal/proxyservice/proxy_service.go
```diff
@@ -1,3 +1,55 @@
+/*
+Package proxyservice takes a Secretless configuration and available plugins
+and constructs the requires ProxyServices that Secretless will run.
+
+To put this in context, the complete high-level flow is:
+
+	1. Parse the `secretless.yml` into individual “service” configs,
+	corresponding to each service entry in the yml.
+
+	2. Identify http services that share a `listenOn`, because we know those
```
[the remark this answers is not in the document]
jonahx: This is something we've discussed before iirc.  My feeling is that it gets handled with first match.  We could log a warning when there are multiple matches.  And then we'd make sure to document that in the public docs as well.

internal/proxyservice/proxy_service.go
```diff
@@ -57,89 +111,145 @@ func (s *proxyServices) Stop() error {
...
-	credsCopy := svc.Credentials
-	credsRetriever := func() (map[string][]byte, error) {
-		return GetSecrets(credsCopy)
+	// TODO: Add validation somewhere about overlapping listenOns
+	// TODO: v2.NetworkAddress is a value type.  It needs to be moved to its
+	//   own package with no deps (stdlib deps are ok).
+	netAddr := v2.NetworkAddress(config.ListenOn)
+	listener, err := net.Listen(netAddr.Network(), netAddr.Address())
+	if err != nil {
+		s.logger.Errorf("could not create listener on: %s", config.ListenOn)
```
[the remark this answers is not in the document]
jonahx: i don't understand this one.   can you clarify?

internal/proxyservice/proxy_service.go
jonahx: let's discuss this afternoon

pkg/secretless/config/v2/config.go
jonahx: Ah, so remember per some of my other TODOs, all this stuff will eventually happen at the border.  So it would be fail fast code, exactly where want it, once we make that change.  I didn't want to make that change yet, however, as it's... not hard exactly, but a can of worms in the same way you updating all the application code to use v2 instead of v1 was.  

TLDR: I agree with you assessment, and plan to solve it, but I think solving it later is the right play.

pkg/secretless/config/v2/service.go
```diff
@@ -126,42 +160,48 @@ func NewService(svcName string, svcYAML *serviceYAML) (*Service, error) {
...
+	return a.split()[0]
+}
+
+// Address returns the "address" part of a network address, eg, "127.0.0.1".
+func (a *NetworkAddress) Address() string {
+	return a.split()[0]
+}
+
+func (a *NetworkAddress) split() []string {
+	return strings.Split(string(*a), "://")
```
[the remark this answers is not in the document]
jonahx: Yeah we could do that.  I'd be fine with it.   It seems like the convention is to use pointer receivers for methods -- that's the only reason I did it -- but there's no reason we have to.  Making it a value causes some extra copying but it's trivial so it doesn't really matter.

pkg/secretless/plugin/so/available_plugins_test.go
```diff
@@ -34,7 +35,7 @@ func getMockPlugins() AvailablePlugins {
 
 func TestPlugins(t *testing.T) {
 	t.Run("HTTPPlugins", func(t *testing.T) {
-		httpPlugins := getMockPlugins().HTTPPlugins()
+		httpPlugins := HTTPPlugins()
```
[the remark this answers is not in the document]
jonahx: Let me look into this later, this wasn't code I wrote.  This just got folded into the PR because of ripple changes from the "Move to..." refactoring the IDE did.

pkg/secretless/plugin/so/internal_plugins_test.go
```diff
@@ -1,16 +1,17 @@
-package plugin
+package so
 
 import (
 	"errors"
 	"testing"
 
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin"
```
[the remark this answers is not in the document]
jonahx: #$%! IDE... I thought it was supposed to hanle this automatically.  Anyway, fixed :)

internal/proxyservice/proxy_service.go
jonahx: closed per discussion on phone

pkg/secretless/config/v2/service.go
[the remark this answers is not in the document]
jonahx: Done.

Issue #899, opened 2019-09-18
Add tests for available_plugins and interal_plugins


sgnn7:
Current issue (#883) does not track the testing needed for this particular feature

AC:
- [ ] available_plugins.go is unit tested
- [x] internal_plugins.go is unit tested
- ~[ ] external_plugins.go is unit tested~ Moved to https://github.com/cyberark/secretless-broker/issues/934

Edit:
~Current PR (#910) adds tests for `available_plugins`. `internal_plugins` tests were included as part of the previously-merged story so the only thing that should remain is tests for `external_plugins`. Per @jonahx we might be able to do this without including a mocking framework like `mockery` so that part of the issue will be left to be finished up next.~

PR #911, opened 2019-09-26
899 Improved available_plugins module and added full unit tests for it

sgnn7:
Addition of tests to this module required a bit more decoupling of
modules and found a bug where the plugin maps were volatile from
consuming classes so this sets of hanges fixes that and adds tests to
cover majority of the functionality.

Connected to #899 but *does not finish it* (`external_plugins` still needs tests)

`go test -v ./internal/plugin/...`


jonahx:
These are my general notes after looking at what we're trying to do here, and thinking about how unit testing it.  This is just outline so may require some in person discussion to clarify.

## Internal Plugins

The current UTs for the internal plugins should be removed imo.

Since we are simply returning a struct, there is a nothing to test.  The
`InternalPluginLookupFunc`, in particular, serves no purpose here. If one were
to insist on testing something here, it would simply be a test asserting that
the plugins we expect are there are there.  And it would need to updated in
tandem with any additions or deletions to the internal plugins we support. 

## External Plugins

For the func `ExternalPlugins`, there are tests to do.  However, they are all
simply tests that our validations are working.  That is, assertions on the
errors  returned for the various error conditions.

Before listing those, some general notes first:

- Give it the raw plugins, rather than the func to look them up.
- When there are true errors, we should return them and leave it to the
  framework to log them.  Which means the only thing we should be using the
  logger for in here is issuing warnings. I would consider passing only a
  warning function, rather than the full logger, to make this clear.  That
  will slightly ease testing the expected warnings as well.
- Note that `go_plugin.Plugin` does not provide you with an interface, so to
  mock this you'll need to create one.  It has only the single method `Lookup`,
  so it will be easy.

So what do we need to test here:

- Returns appropriate errors for the validations done in `parsePluginMetadata`
- Check that it returns any the expected errors
- Check that it issues all the expected warnings
- Correctly separates plugins by type 


For `LoadPluginsFromDir`, there's not much to test.  I would probably not bother.
If you _were_ going to test this though, you'd need to inject the `os` functions 
somehow (not hard, but slightly longer discussion).  Then you'd be testing that:

- `"Plugin directory '%s' not found. Ignoring external plugins..."` warning is issued 
when appropriate.
- That the plugins are verified.  Again, requires injection of
  `VerifyPluginChecksums`.  Also not hard but details are another discussion.

That's it.  There's nothing else to test here.  The actual logic in
`VerifyPluginChecksums` is tested in those UTs.

sgnn7:
@jonahx some thoughts on your comments:

> it would simply be a test asserting that
the plugins we expect are there are there. And it would need to updated in
tandem with any additions or deletions to the internal plugins we support.

Sure but that's exactly the idea behind a UT here: we would be verifying that some aspects of internal plugins list is correct (length, type, maybe both?) so that problems here are identified by the tests (even though it's duplication and seems superfluous to some extent).

> the only thing we should be using the logger for in here is issuing warnings

Yes and no to an extent? I feel like we should strive to log whatever is valuable and the debug filter should snip out the non warnings if that flag is false otherwise that debugging info won't exist at all. To me it kind of seems like the point of the logger and you dial up/down the level to show more or less of that info.

> For the func ExternalPlugins, there are tests to do

So this is the problem - I know what I should test but _how_ to test this seemed overly gnarly to me without a mocking framework. Any way you can do a test or two on this so that I can see what you had in mind?

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -16,26 +19,111 @@ type mockTCPPlugin struct{ tcp.Plugin }
...
+
+	for name, httpPlugin := range expectedPlugins {
+		assert.Contains(t, allPlugins.HTTPPlugins(), name)
+		if _, ok := allPlugins.HTTPPlugins()[name]; !ok {
+			t.Fail()
+		}
+
+		if comparePointers {
+			assert.Equal(t, httpPlugin, allPlugins.HTTPPlugins()[name])
+		}
```
jonahx: I forget the name for this, but it's something like "Control Flow Via Parameter", and it's something you want to avoid.  It looks like in this case there really are two different tests -- one that the key is there, on that the value is the same -- and so you should just make separate assert helpers.

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -16,26 +19,111 @@ type mockTCPPlugin struct{ tcp.Plugin }
...
+		assert.Contains(t, allPlugins.TCPPlugins(), name)
+		if _, ok := allPlugins.TCPPlugins()[name]; !ok {
+			t.Fail()
+		}
+
+		if comparePointers {
+			assert.Equal(t, tcpPlugin, allPlugins.TCPPlugins()[name])
+		}
+	}
+}
```
jonahx: Same comment here...

Also, f'ing Go and its lack of generics.

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -16,26 +19,111 @@ type mockTCPPlugin struct{ tcp.Plugin }
...
+
+func newLogger() log_api.Logger {
+	return log.New(true)
+}
+
```
jonahx: Def not a suggestion for now as it's not related to this code, but this makes me think we should have used the functional options patterns with a logger, so we could write:
```go
log.New(log.WithDebug, etc...)
```
The "true" it totally meaningless unless I'm viewing with my IDE.

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -16,26 +19,111 @@ type mockTCPPlugin struct{ tcp.Plugin }
...
 
-func getMockPlugins() plugin.AvailablePlugins {
+func mockPlugins() plugin.AvailablePlugins {
 	return &Plugins{
 		HTTPPluginsByID: mockHTTPPlugins,
 		TCPPluginsByID:  mockTCPPlugins,
 	}
 }
 
+func mockInternalPlugins() plugin.AvailablePlugins {
```
jonahx: Naming nit: 

Probably even better is to create a mock subpackage, which will clean up this test file and let you rely on the package modifier for naming.... mock.InternalPlugins() etc.

Kumbi's been using this pattern in some places.

pkg/secretless/plugin/sharedobj/available_plugins_test.go
sgnn7: I agree but personally I would like use to move to a structured logging levels which would make that happen by the nature of it: 
```
log.New(etc...)
log.SetLevel(log.DebugSeverity)
```
 (maybe the invocation would be different though)

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -6,263 +6,161 @@ import (
...
 		assert.Nil(t, err)
 		if err != nil {
 			t.Fail()
 		}
 
-		assert.Equal(t, 5, len(allPlugins.HTTPPlugins()))
-		assert.Equal(t, 6, len(allPlugins.TCPPlugins()))
+		// Ensure the override occurred
+		assert.Equal(t, extHTTPPlugins["intHTTP2"], allPlugins.HTTPPlugins()["intHTTP2"])
+		assert.Equal(t, extTCPPlugins["intTCP1"], allPlugins.TCPPlugins()["intTCP1"])
```
sgnn7: Not a blocker: This test is a much weaker assertion - we don't know for a fact that the rest of the plugins are unchanged

Blocker: `assert.Equal` compares values, not pointers so this test isn't testing the right thing afaict since both the original and replacement structs have the same empty value.

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -1,60 +1,166 @@
...
-
-// TODO: Test AllAvailablePluginsWithOptions
-
-// These need to sit outside of the func since we are comparing them in the
-// tests
-var mockHTTPPlugins = map[string]http.Plugin{
-	"one": mockHTTPPlugin{},
-	"two": mockHTTPPlugin{},
+func TestCompatiblePluginVersion(t *testing.T) {
+	assert.Equal(t, "0.1.0", CompatiblePluginAPIVersion)
```
jonahx: It's unclear to me what the value of this test is.  If you forgot to update this on a version change, the test would continue to pass.  It would only be valuable if you happened to remember to update the test but forgot to update the implementation.  It seems like rather than a test we want an item in a checklist somewhere that would remind us to update the actual impl value when appropriate.

pkg/secretless/plugin/sharedobj/available_plugins.go
```diff
@@ -91,7 +83,16 @@ func AllAvailablePluginsWithOptions(
...
+	tcpPlugins := map[string]tcp.Plugin{}
+
+	for name, tcpPlugin := range internalPlugins.TCPPlugins() {
+		if _, ok := tcpPlugins[name]; ok {
+			logger.Warnf("Internal plugin '%s' is replaced by an externally-provided plugin",
+				name)
+		}
+
+		tcpPlugins[name] = tcpPlugin
+	}
```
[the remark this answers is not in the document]
sgnn7: Created: https://github.com/cyberark/secretless-broker/issues/979

pkg/secretless/plugin/sharedobj/available_plugins.go
```diff
@@ -105,3 +97,27 @@ func AllAvailablePluginsWithOptions(
 		TCPPluginsByID:  tcpPlugins,
 	}, nil
 }
+
+// NewEmptyPlugins plugins creates a new instance of Plugins with both maps
+// initialized but empty.
+func NewEmptyPlugins() Plugins {
```
sgnn7: Why not just `NewPlugins` - It seems like one can assume that they are empty (and not `nil` like they would be if you instantiate `Plugins` struct directly.

pkg/secretless/plugin/sharedobj/available_plugins_test.go
sgnn7: Maybe the better test would be a regex that asserts that we are in semVer format?

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -1,60 +1,177 @@
...
+
+		assert.EqualValues(t, mock.AllHTTPPlugins(), allPlugins.HTTPPlugins())
+		assert.EqualValues(t, mock.AllTCPPlugins(), allPlugins.TCPPlugins())
+	})
 
-		assert.NotNil(t, httpPlugins)
-		if httpPlugins == nil {
-			t.Fail()
+	t.Run("External plugins override same-named internal plugins", func(t *testing.T) {
+
```
sgnn7: No need for this LF

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -1,60 +1,177 @@
...
+		// Define our expected results
+		expectedHTTP := mock.AllHTTPPlugins()
+		expectedHTTP["intHTTP2"] = httpOverride
+		expectedTCP := mock.AllTCPPlugins()
+		expectedTCP["intTCP1"] = tcpOverride
+
+		// Ensure both the override and everything else worked
+		assert.Equal(t, expectedHTTP, allPlugins.HTTPPlugins(), "HTTP override failed")
+		assert.Equal(t, expectedTCP, allPlugins.TCPPlugins(), "TCP override failed")
+
```
sgnn7: Extra line not needed

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -1,60 +1,177 @@
...
+		// Define our expected results
+		expectedHTTP := mock.AllHTTPPlugins()
+		expectedHTTP["intHTTP2"] = httpOverride
+		expectedTCP := mock.AllTCPPlugins()
+		expectedTCP["intTCP1"] = tcpOverride
+
+		// Ensure both the override and everything else worked
+		assert.Equal(t, expectedHTTP, allPlugins.HTTPPlugins(), "HTTP override failed")
+		assert.Equal(t, expectedTCP, allPlugins.TCPPlugins(), "TCP override failed")
```
sgnn7: Just a nit but this is where a sanity check would be useful (`assert.NotEqual` before you reassign the plugins here (L121/L123))

Issue #894, opened 2019-09-17
Logger includes severity level in messages outputted


sgnn7:
```
...
2019/09/17 13:02:17 File 'test.sh' ignored as a plugin - missing appropriate extension
2019/09/17 13:02:17 plugin: symbol PluginInfo not found in plugin plugin/unnamed-42c72f2d67625dd02f4b78e493d5dd498fad6fe8
2019/09/17 13:02:17 Listing http plugins...
2019/09/17 13:02:17 Listing tcp plugins...
```

first is a warning, second is an error, third/forth are info messages. they look the same on the console.

AC:
- All logger messages contain prominent severity level when printed

An example output of that code with AC completed (but feel free to change it):
```
2019/09/17 13:02:17 WARN: File 'test.sh' ignored as a plugin - missing appropriate extension
2019/09/17 13:02:17 ERRO: plugin: symbol PluginInfo not found in plugin plugin/unnamed-42c72f2d67625dd02f4b78e493d5dd498fad6fe8
2019/09/17 13:02:17 INFO: Listing http plugins...
2019/09/17 13:02:17 INFO: Listing tcp plugins...
```

PR #912, opened 2019-09-27
894 Logging level output

sgnn7:
Adds severity levels to the logger output

Connected to #894 

Diff

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -1,6 +1,7 @@
 package log
 
 import (
+	"fmt"
 	"io"
 	stdlib_log "log"
 	"os"
@@ -36,14 +37,14 @@ const (
 	PanicSeverity
 )
 
-// severityLevels is a list of all available severity levels. This is a shorthand
-// to work around the fact that Golang cannot `range` over an enum.
-var severityLevels = []severity{
-	DebugSeverity,
-	InfoSeverity,
-	WarnSeverity,
-	ErrorSeverity,
-	PanicSeverity,
+// severityLevels is a mapping of all available severity levels to their printed
+// values.
+var severityLevels = map[severity]string{
+	DebugSeverity: "DEBUG",
+	InfoSeverity:  "INFO",
+	WarnSeverity:  "WARN",
+	ErrorSeverity: "ERROR",
+	PanicSeverity: "PANIC",
 }
 
 // New method instantiates a new logger that we can write things to.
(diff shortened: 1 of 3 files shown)


internal/log/log.go
```diff
@@ -36,14 +37,14 @@ const (
...
-	ErrorSeverity,
-	PanicSeverity,
+// severityLevels is a mapping of all available severity levels to their printed
+// values.
+var severityLevels = map[severity]string{
+	DebugSeverity: "DEBUG",
+	InfoSeverity:  "INFO",
+	WarnSeverity:  "WARN",
+	ErrorSeverity: "ERROR",
+	PanicSeverity: "PANIC",
```
jonahx: Minorish thing: Not directly related to this PR, but I noticed with doing the panic fix that the `Severity` suffix on the constant names didn't add anything and felt like it was actually harming readability.  I say we just drop it.

internal/log/log.go
```diff
@@ -113,6 +118,10 @@ func (logger *Logger) logf(severityLevel severity, format string, args ...interf
 		args = prependString(logger.prefix, args...)
 	}
 
+	format = "%s " + format
+
+	args = prependString(severityPrefix(severityLevel), args...)
+
```
jonahx: Feel free to disregard because this is a borderline/subjective case, but fwiw I think the local var name `sev` would go down easier as you parse this code, and would be ok given how short the func is.

internal/log/log_test.go
```diff
@@ -57,21 +57,29 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
...
 			func(t *testing.T) {
 				method(testCaseFormatStr, testCaseArgs...)
-				assert.Regexp(t, lt.expectedOutput(methodName), lt.CurrentOutput())
+				severityLevelStr := strings.ToUpper(methodName[:len(methodName)-1])
+				assert.Regexp(
+					t,
+					lt.expectedOutput(
+						methodName,
+						severityLevelStr,
+					),
```
jonahx: Collapse the 4 lines above into a single line since they'll fit comfortably.

internal/log/log_test.go
```diff
@@ -92,7 +100,18 @@ func (lt *LogTest) RunAllTests(t *testing.T) {
...
+				severityLevelStr := strings.ToUpper(baseMethodName)
+
+				assert.Regexp(
+					t,
+					lt.expectedOutput(
+						methodName,
+						severityLevelStr,
+					),
+					lt.CurrentOutput(),
+				)
```
jonahx: This feels like a little funky to me, but these tests are already kind of out of control after the panic fix, so not a blocker for this PR.  We should def do a proper fixup eventually though 😬

sgnn7: fair enough

PR #923, opened 2019-10-01
848 http proxy service exists

jonahx:
WIP

### DIFF
--- a/internal/proxyservice/http/proxy_service.go
+++ b/internal/proxyservice/http/proxy_service.go
@@ -1,8 +1,15 @@
 package http
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io"
 	"net"
+	gohttp "net/http"
+	"regexp"
+
+	validation "github.com/go-ozzo/ozzo-validation"
 
 	"github.com/cyberark/secretless-broker/internal"
 	"github.com/cyberark/secretless-broker/pkg/secretless/log"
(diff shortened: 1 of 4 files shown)


internal/proxyservice/http/proxy_service.go
```diff
@@ -12,18 +19,269 @@ import (
...
+	}
+
+	// Send request to target service
+
+	proxy.handleRequest(w, r)
+}
+
+// validateProxyServerRules ensures that the request being made is a valid
+// request for a proxy server.
+func (proxy *proxyService) validateProxyServerRules(
```
doodlesbykumbi: the naming seems a bit dishonest here. i might be mis-assigning meaning to the term "validation". my expectation, which might be wrong, is for validation to tell me if the input is valid or not, without side-effects.

internal/proxyservice/http/proxy_service.go
```diff
@@ -12,18 +19,269 @@ import (
...
+
+// handleRequest sends the request to the target service and writes the response
+// back to the client.
+func (proxy *proxyService) handleRequest(
+	w gohttp.ResponseWriter, r *gohttp.Request,
+) {
+	logger := proxy.logger
+
+	// TODO: Must be reset for... some reason?
+	r.RequestURI = ""
```
doodlesbykumbi: I think this is because we're getting our request from the client. I suspect that `RequestURI` is populated from the client...
The Go docs say
> // It is an error to set this field in an HTTP client request.

internal/proxyservice/http/proxy_service.go
```diff
@@ -12,18 +19,269 @@ import (
...
+func (proxy *proxyService) Start() error {
+	logger := proxy.logger
+
+	logger.Infof("Starting service")
+
+	if proxy.done {
+		return fmt.Errorf("cannot call Start on stopped ProxyService")
+	}
+
+	// TODO: why do we start a new goroutine here?
```
doodlesbykumbi: The ProxyService `Start` method must be `non-blocking`. `http.Server` is blocking

internal/proxyservice/proxy_service.go
```diff
@@ -123,9 +123,17 @@ func (s *proxyServices) createHTTPService(
...
+		// Get the http traffic patterns to match from the connector config.
+		httpCfg, err := v2.NewHTTPConfig(subCfg.ConnectorConfig)
+		if err != nil {
+			return nil, err
+		}
+
 		subservices = append(subservices, httpproxy.Subservice{
-			Connector:           curConnector,
-			RetrieveCredentials: credsRetriever,
+			ConnectorID:              subCfg.Connector, // TODO: Rename connectorID
```
doodlesbykumbi: As far as I can tell ConnectorID is used in a single place for logging. Wondering if it's overkill to have it as a field on the subservice.

internal/proxyservice/http/proxy_service.go
```diff
@@ -12,18 +19,269 @@ import (
...
+
+	// Multiple matches: Warn!
+	if len(matchingSubs) > 1 {
+		msg := "Multiple subservices matched request '%s': %v\n"
+		proxy.logger.Warnf(msg, r.URL.String(), matchingSubs)
+	}
+
+	// Select first (or only) match
+	subservice := matchingSubs[0]
+	msg := "Using connector '%s' for request %s"
```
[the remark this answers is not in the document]
jonahx: 2 lines instead of 4.  Just something I was experimenting with.  Wdyt?

internal/proxyservice/http/proxy_service.go
jonahx: Yeah I think that is a mis-assignment, at least in my mind.

Following the "verb phrase for side effect" rule `ValidateXX` implies that it will do a validation and (in other langs) throw errors or (in Go) return an error or (in this case) write to a log.  If you were writing a method that just returned a boolean it's name would be of the form `IsValidXXX`.  You can see I followed this convention here: 

https://github.com/cyberark/secretless-broker/blob/master/internal/profile/profile.go#L37 

Ofc this isn't set in stone but the distinction is useful and the naming convention is in line with the more general one so I like it.

internal/proxyservice/proxy_service.go
jonahx: Yeah there's two perspectives here.  One is the narrower one which is, "Eh, let's just remove it _and_ the log statement and keep it simple."  The larger one, in my mind, which I was hinting at with this:

> // TODO: Looks like ConnectorID and the function it represents shouldn't be separated

is that this "little" problem with the log is actually a smell that our design is subtly wrong, and the the concept of a Connector, as we think about it, isn't just a raw function, but at minimum a function _and_ and an id/name.  This hasn't bit us except here, and so practically we can probably ignore it, at least for now, but I do believe it's a small design error.

internal/proxyservice/http/proxy_service.go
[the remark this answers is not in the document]
doodlesbykumbi: I'm actually not sure. Part of me is screaming vehemently like "We must adhere to the convention". Another part of me thinks the separate message makes it easier to parse

internal/proxyservice/proxy_service.go
doodlesbykumbi: I agree. I suppose we can reevaluate this as more information comes to light.

PR #926, opened 2019-10-08
add postgres service connector

doodlesbykumbi:


internal/proxyservice/tcp/pg/backend.go
```diff
@@ -2,51 +2,46 @@ package pg
...
-	"github.com/cyberark/secretless-broker/internal/handlers/pg/protocol"
 	"github.com/cyberark/secretless-broker/internal/handlers/ssl"
-	"github.com/cyberark/secretless-broker/internal/util"
+	"github.com/cyberark/secretless-broker/internal/proxyservice/tcp/pg/protocol"
 )
 
 // ConfigureBackend resolves the backend connection settings and credentials and sets the
 // connectionDetails field.
-func (h *Handler) ConfigureBackend() error {
+func (s *SingleUseConnector) ConfigureBackend(credentials map[string][]byte) error {
```
jonahx: (make sure to read the security comment below, as it's related to this one...)

I know this is a carryover from the old code, but this name isn't really accurate or descriptive.  And since we're rewriting, we should improve the old mistakes. 

I have two thoughts:

1. `BackendConfig` -> `ConnectionDetails` (name we use in mysql, which is good)
2. _drumroll please...._ 

DELETE this method entirely.  If you think about what it's doing, it's quite silly.  Instead, we should move the relevant content that's in this method into `NewBackendConfig` / `NewConnectionDetails`.  Then update this part of `Connect`: 
```go
	if err := s.ConfigureBackend(credentialValuesByID); err != nil {
		s.abort(err)
		return nil, err
	}
```
to something like:
```go
connDetails, err := NewConnectionDetails(credentialValuesByID); err != nil {
 		s.abort(err)
		return nil, err
}
s.ConnectionDetails = connDetails
```

internal/proxyservice/tcp/pg/backend.go
```diff
@@ -104,23 +99,23 @@ func (h *Handler) ConnectToBackend() (err error) {
 		}
 	}
 
-	startupMessage := protocol.CreateStartupMessage(h.BackendConfig.Username, h.ClientOptions.Database, h.BackendConfig.Options)
+	startupMessage := protocol.CreateStartupMessage(s.backendConfig.Username, s.clientOptions.Database, s.backendConfig.Options)
```
jonahx: split into lines (same with other ones below)

internal/proxyservice/tcp/pg/backend.go
```diff
@@ -2,51 +2,46 @@ package pg
...
+	if connection, err = net.Dial("tcp", s.backendConfig.Address()); err != nil {
 		return
 	}
 
-	debug := util.OptionalDebug(h.GetConfig().Debug)
-	debug("Sending startup message")
+	s.logger.Debugln("Sending startup message")
 
-	tlsConf, err := ssl.NewDbSSLMode(h.BackendConfig.SSLOptions, true)
+	tlsConf, err := ssl.NewDbSSLMode(s.backendConfig.SSLOptions, true)
```
jonahx: Old code I know, and not a blocker, but if there's a clean way to pull out that `if tlsConf.UseTLS {` block below (I didn't give it too much thought) it would be nice to.

internal/proxyservice/tcp/pg/backend_config.go
```diff
@@ -16,7 +15,7 @@ var sslOptions = []string{
 	"sslcert",
 }
 
-// BackendConfig stores the connection info to the real backend database.
+// backendConfig stores the connection info to the real backend database.
```
jonahx: See my comment above on this struct.

internal/proxyservice/tcp/pg/handler.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/cyberark/secretless-broker/internal/proxyservice/tcp/pg/protocol"
+	"github.com/cyberark/secretless-broker/pkg/secretless/log"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+)
+
+// clientOptions stores the option that were specified by the connection client.
+// The User and Database are required. Other options are stored in a map.
+type ClientOptions struct {
+	// TODO: remove this when custom authorization is removed.
```
jonahx: This is probably existing code, but I didn't understand the TODOs.  Can you clarify if you do?

internal/proxyservice/tcp/pg/handler.go
```diff
@@ -0,0 +1,80 @@
...
+func (s *SingleUseConnector) abort(err error) {
+	if s.clientConn != nil {
+		pgError := protocol.Error{
+			Severity: protocol.ErrorSeverityFatal,
+			Code:     protocol.ErrorCodeInternalError,
+			Message:  err.Error(),
+		}
+		s.clientConn.Write(pgError.GetPacket())
+	}
+}
```
jonahx: 👍

internal/proxyservice/tcp/pg/plugin.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
+)
+
+// NewConnector returns a tcp.Connector which returns an authenticated connection
+// to a target service for each incoming client connection.
+// It is a required method on the tcp.Plugin interface.
+// The single argument passed in is of type connector.Resources. It contains
+// connector-specific config and a logger.
```
jonahx: Nit: formatting.  Hotkey that bad boy. Also I think lines 13/14 aren't needed, ie, no need to re-iterate the interface def of `connector.Resources` in the docs for `NewConnector`

internal/proxyservice/tcp/pg/startup.go
```diff
@@ -43,17 +40,15 @@ func (h *Handler) Startup() (err error) {
 		return
 	}
 
-	if h.GetConfig().Debug {
-		log.Printf("h.Client version : %v, (SSL mode: %v)", version, version == protocol.SSLRequestCode)
-	}
+	s.logger.Debugln("s.Client version : %v, (SSL mode: %v)", version, version == protocol.SSLRequestCode)
```
jonahx: split lines

internal/proxyservice/tcp/pg/backend.go
```diff
@@ -2,51 +2,46 @@ package pg
...
 // connectionDetails field.
-func (h *Handler) ConfigureBackend() error {
+func (s *SingleUseConnector) ConfigureBackend(credentials map[string][]byte) error {
 	var err error
-	var options map[string][]byte
 
-	if options, err = h.Resolver.Resolve(h.GetConfig().Credentials); err != nil {
-		return err
-	}
+	keys := reflect.ValueOf(credentials).MapKeys()
```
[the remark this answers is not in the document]
doodlesbykumbi: Taken straight from old code...

I understand not using reflection to write our own meta this that and the other, because it's relatively complex and requires a grasp of reflection (which is a bar we're not willing to erect for our developers). There might be a stronger and separate reason, if so please share.

In this case, it is a single line and doesn't seem to require explanation. I'd be inclined to leave it be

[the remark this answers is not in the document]
jonahx: > Taken straight from old code...Mainly just cause every best practice guide I've come across says to avoid it (eg, https://github.com/codeship/go-best-practices/tree/master/reflection).

_Would you jump off a building if a best practice guide told you to, Jonah?_

Yes.  Yes I would.

But I do actually have some reasons:
1. Since the standard is never to use reflection unless needed, it makes anyone reading the code pause and ask why you did use reflection here.  Is there some reason we're not use using a loop?  Is the code doing more than it seems?
2. Broken windows phenomenon: It gives the stamp approval to this, and then other devs may start using it.  That is, idioms spread inside a code base.
3. I think it makes us look like we don't know

internal/proxyservice/tcp/pg/startup.go
```diff
@@ -26,14 +25,12 @@ func newClientOptions(options map[string]string) (co *ClientOptions, err error)
 	return
 }
 
-// Startup performs the startup handshake with the client and parses the ClientOptions.
-func (h *Handler) Startup() (err error) {
-	if h.GetConfig().Debug {
-		log.Printf("Handling connection %v", h.GetClientConnection())
-	}
+// Startup performs the startup handshake with the client and parses the clientOptions.
+func (s *SingleUseConnector) Startup() (err error) {
```
jonahx: Shouldn't be using naked returns here: https://github.com/golang/go/wiki/CodeReviewComments#naked-returns

internal/proxyservice/tcp/pg/backend.go
```diff
@@ -2,51 +2,46 @@ package pg
...
-	if h.GetConfig().Debug {
-		keys := reflect.ValueOf(options).MapKeys()
-		log.Printf("%s backend connection parameters: %s", h.GetConfig().Name, keys)
+	if len(credentials["address"]) > 0 {
+		s.logger.Warnf("'address' has been deprecated for PG connector. " +
+			"Please use 'host' and 'port' instead.'")
 	}
 
-	backendConfig, err := NewBackendConfig(options)
+	backendConfig, err := NewBackendConfig(credentials)
```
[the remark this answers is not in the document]
doodlesbykumbi: This is a bigger story/conversation than this PR. I'd say let's defer this one to later.

internal/proxyservice/tcp/pg/handler.go
doodlesbykumbi: ClientOptions isn't actually needed. Got rid of it

internal/proxyservice/tcp/pg/connector.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		s.clientConn.Write(pgError.GetPacket())
+	}
+}
+
+// Connect implements the tcp.Connector interface
+//
+// It is the main method of the SingleUseConnector. It:
+//   1. Constructs connection details from the provided credentials map.
+//   1. Dials the backend using credentials.
```
jonahx: two 1. and no 2.

PR #927, opened 2019-10-10
add ssh service proxy

doodlesbykumbi:


internal/proxyservice/proxy_service.go
```diff
@@ -87,6 +88,7 @@ func (s *proxyServices) servicesToStart() (servicesToStart []internal.Service) {
 		httpSvc, err := s.createHTTPService(httpSvcConfig, httpPlugins)
 		if err != nil {
 			// TODO: Add Fatalf to our logger and use that
+			s.logger.Error(err)
```
sgnn7: Seems superfluous if you can attach the error to the panic message below it

internal/proxyservice/proxy_service.go
```diff
@@ -156,6 +169,38 @@ func (s *proxyServices) createHTTPService(
 	return newSvc, nil
 }
 
+func (s *proxyServices) createSSHService(
```
sgnn7: I think we might be stuffing way too much into this file in an almost copy/paste manner making me think this can get collapsed in some way. Can this be split out or genericized to avoid needing extra code here for every type of connector?

internal/proxyservice/ssh/listener.go
```diff
@@ -0,0 +1,81 @@
...
+	config_v2 "github.com/cyberark/secretless-broker/pkg/secretless/config/v2"
+)
+
+// Listener accepts SSH connections and MITMs them using a ServiceConnector.
+//
+// NOTE: This MITM approach to SSH is experimental. The ssh-agent approach is
+// better validated and probably better all-around.
+
+// serviceHasCredentials validates that a service has all necessary credentials.
+type serviceHasCredentials struct {
```
sgnn7: I think this name might need a better one - it sounds like it's a boolean but it's not

internal/proxyservice/ssh/proxy_service.go
```diff
@@ -0,0 +1,203 @@
+package ssh
```
sgnn7: why not `ssh.go` for filename? I guess the question is mostly due to disambiguating this file and `proxyservice/proxy_service.go`

internal/proxyservice/ssh/proxy_service.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	err := conn.Close()
+	if err != nil {
+		logger.Warnf("Failed on closing %s connection: %s", connDescription, err)
+	}
+}
+
+func (proxy *proxyService) handleConnections(channels <-chan ssh.NewChannel) error {
+	var connector = ServiceConnector{
+
```
sgnn7: stray LF

internal/proxyservice/ssh/proxy_service.go
```diff
@@ -0,0 +1,203 @@
...
+		channels: channels,
+		logger:   proxy.logger,
+	}
+	backendCredentials, err := proxy.retrieveCredentials()
```
sgnn7: tiny nit - maybe an LF right before this line might help with readability

internal/proxyservice/ssh/proxy_service.go
```diff
@@ -0,0 +1,203 @@
...
+	logger := proxy.logger
+
+	logger.Infof("Starting service")
+
+	if proxy.done {
+		return fmt.Errorf("cannot call Start on stopped ProxyService")
+	}
+
+	go func() {
+
```
sgnn7: stray LF

pkg/secretless/config/v2/config.go
```diff
@@ -78,7 +78,20 @@ func NewConfigsByType(
 		rawConfigs = append(rawConfigs, *cfg)
 	}
 
-	httpConfigs, tcpConfigs := separatedHTTPAndTCPConfigs(rawConfigs, availPlugins)
+	var httpConfigs , tcpConfigs, sshConfigs []Service
```
sgnn7: stray whitespace

pkg/secretless/config/v2/config.go
```diff
@@ -78,7 +78,20 @@ func NewConfigsByType(
...
+
+	for _, cfg := range rawConfigs {
+		if sharedobj.IsHTTPPlugin(availPlugins, cfg.Connector) {
+			httpConfigs = append(httpConfigs, cfg)
+			continue
+		} else if cfg.Connector == "ssh" {
+			sshConfigs = append(sshConfigs, cfg)
+			continue
+		}
```
sgnn7: unrelated/nit/notablocker: we might want this giant select generalized if we add one more somehow

pkg/secretless/config/v2/config.go
```diff
@@ -94,6 +107,7 @@ func NewConfigsByType(
 	return ConfigsByType{
 		TCP:  tcpConfigs,
 		HTTP: httpServiceConfigs,
+		SSH: sshConfigs,
```
sgnn7: members sort needed

pkg/secretless/config/v2/config.go
```diff
@@ -116,6 +130,7 @@ func (cfg *HTTPServiceConfig) Name() string {
 // will be created for each entry in http, and one for each entry in tcp.
 type ConfigsByType struct {
 	TCP  []Service
+	SSH  []Service
```
sgnn7: member sort needed

internal/proxyservice/ssh/listener.go
doodlesbykumbi: wasn't needed anymore

internal/proxyservice/ssh/proxy_service.go
doodlesbykumbi: i see your point. however, this is consistent with the other proxy services (tcp and http). i'd say that's a refactor for later

internal/proxyservice/proxy_service.go
doodlesbykumbi: We definitely should do that but I'd say let's not do it as part of this particular PR. Unless if there's a good abstraction in mind right now

sgnn7: Fair but I would expect to see at least a bit of work on this for your next PR if you're touching the same code :)

internal/proxyservice/ssh/proxy_service.go
sgnn7: 😿  but makes sense

Issue #916, opened 2019-09-27
Reimplement/reattach configfile configuration file loading logic


sgnn7:
Given that we don't use a configmanager pattern anymore, our configuration file loading is practically non-existent and not backard compatible. Reapply the same logic to loading of config files as before (hopefully reusing [configfile](https://github.com/cyberark/secretless-broker/tree/latest/internal/app/secretless/configurationmanagers/configfile) package)

AC:
- [ ] Broker config is attempted to be loaded from standard paths automatically (`secretless.yml` -> `$HOME/.secretless.yml` -> `/etc/secretless.yml`

PR #935, opened 2019-10-15
Reenable full configfile manager functionality

sgnn7:
New code used a bespoke (and basic) configfile loading procedure so with
these commits, we are reattaching the old configfile manager to it.
There's no new API yet, and we still need to get the CRDs configuration manager
attached but this should be a stepping stone for that.

Changes:
- Bespoke configuration loader removed
- `configfile` manager re-enabled
- Indirect result of this also enables USR1 signal reload-config trigger
- `configfile` manager modified to support this feature
- Proxy loop modified to remove existing socket files before reloading (to prevent failures from lack of cleanups)
- Main proxy loop modified to support a config channel configuration

Connected to #916 

Diff

### DIFF
--- a/internal/configurationmanagers/configfile/manager.go
+++ b/internal/configurationmanagers/configfile/manager.go
@@ -171,8 +171,8 @@ func ConfigurationManagerFactory(options plugin_v1.ConfigurationManagerOptions) 
 	}
 }
 
-// NewManager returns a file-based ConfigurationManager channel object
-func NewManager(configfile string, fsWatchEnabled bool) (<-chan config_v2.Config, error) {
+// NewConfigChannel returns a file-based ConfigurationManager channel object
+func NewConfigChannel(configfile string, fsWatchEnabled bool) (<-chan config_v2.Config, error) {
 	configChangedChan := make(chan config_v2.Config)
 
 	manager := &configurationManager{}
--- a/internal/configurationmanagers/configfile/manager.go
+++ b/internal/configurationmanagers/configfile/manager.go
@@ -29,6 +29,17 @@ type configurationManager struct {
 	Name              string
 }
 
+// Local adapter struct that changes the ConfigurationChangedHandler events to
+// chan config_v2.Config messages
+type changeHandler struct {
+	ConfigChangeChan chan config_v2.Config
+}
+
+func (ch *changeHandler) ConfigurationChanged(_ string, config config_v2.Config) error {
+	ch.ConfigChangeChan <- config
+	return nil
+}
+
 func (configManager *configurationManager) getConfigFilePreferenceOrder() ([]string, error) {
 	configFileOrder := []string{
 		"./" + ConfigFileName,
@@ -158,4 +169,25 @@ func ConfigurationManagerFactory(options plugin_v1.ConfigurationManagerOptions) 
 	return &configurationManager{
 		Name: options.Name,
 	}
+}
+
+// NewManager returns a file-based ConfigurationManager channel object
+func NewManager(configfile string, fsWatchEnabled bool) (<-chan config_v2.Config, error) {
+	configChangedChan := make(chan config_v2.Config)
+
+	manager := &configurationManager{}
+
+	cfgChangeHandler := changeHandler{
+		ConfigChangeChan: configChangedChan,
+	}
+
+	// Managers expect parameters to be passed in as URL params
+	cfgSpec := fmt.Sprintf("%s?watch=%t", configfile, fsWatchEnabled)
+
+	err := manager.Initialize(&cfgChangeHandler, cfgSpec)
+	if err != nil {
+		return nil, err
+	}
+
+	return cfgChangeHandler.ConfigChangeChan, nil
 }
--- a/internal/proxyservice/proxy_service.go
+++ b/internal/proxyservice/proxy_service.go
@@ -3,6 +3,8 @@ package proxyservice
 import (
 	"fmt"
 	"net"
+	"net/url"
+	"os"
 	"strings"
 
 	"github.com/cyberark/secretless-broker/internal"
@@ -24,6 +26,7 @@ import (
 type proxyServices struct {
 	availPlugins    plugin2.AvailablePlugins
 	config          v2.Config
+	configsByType   v2.ConfigsByType
 	eventNotifier   v1.EventNotifier
 	logger          logapi.Logger
 	runningServices []internal.Service
@@ -31,6 +34,14 @@ type proxyServices struct {
 
 // Start starts all proxy services
 func (s *proxyServices) Start() error {
+	for _, service := range s.config.Services {
+		err := s.ensureSocketIsDeleted(service.ListenOn)
+		if err != nil {
+			// TODO: Add Fatalf to our logger and use that
+			s.logger.Panic(err)
+		}
+	}
+
 	for _, svc := range s.servicesToStart() {
 		err := svc.Start()
 		if err != nil {
@@ -45,6 +56,8 @@ func (s *proxyServices) Start() error {
 // Stop stops all proxy services
 func (s *proxyServices) Stop() error {
 	var stopFailures []string
+
+	s.logger.Infoln("Stopping services...")
 	for _, svc := range s.runningServices {
 		err := svc.Stop()
 		if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,49 +53,77 @@ func StartSecretless(params *SecretlessOptions) {
...
+	util.SetAppIsLive(false)
 
-	handlePerformanceProfiling(params.ProfilingMode, exitListener)
+	var allServices internal.Service
+
+	configChangedFunc := func(cfg v2.Config) {
+		util.SetAppIsLive(false)
+		// Start Services
+		allServices = proxyservice.NewProxyServices(cfg, availPlugins, logger, evtNotifier)
+		err = allServices.Start()
```
jonahx: Don't we need to stop the old services too?

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,49 +53,77 @@ func StartSecretless(params *SecretlessOptions) {
...
+		for {
+			logger.Info("Waiting for configuration...")
+			cfg := <-configChangedChan
 
-	// Health check: Live
-	util.SetAppIsLive(true)
+			logger.Debug("Got new configuration")
+			configChangedFunc(cfg)
+		}
+	}()
```
jonahx: Looks like this goroutine is not stopped cleanly when we exit.  Not a blocker for me but should at least add a TODO.

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,49 +53,77 @@ func StartSecretless(params *SecretlessOptions) {
...
 
-	// Optional Performance Profiling
+	// Health check: Initialized
+	util.SetAppInitializedFlag()
+	util.SetAppIsLive(false)
 
-	handlePerformanceProfiling(params.ProfilingMode, exitListener)
+	var allServices internal.Service
+
+	configChangedFunc := func(cfg v2.Config) {
```
jonahx: I think a good naming rule is to reserve the `Func` suffix for naming function _types_. 

If you think about it, it doesn't really make sense that you'd change your naming scheme just because you're defining a function as a local var vs doing `func blah(...` in the package.  

Thus we can stick with the normal practice of finding a descriptive verb phrase... `reloadConfig` seems appropriate here.  

Then down below it will read:
```go
logger.Debug("Got new configuration")
reloadConfig(cfg)
```
which is more natural and descriptive than:
```go
```go
logger.Debug("Got new configuration")
configChangedFunc(cfg)
```

internal/proxyservice/proxy_service.go
```diff
@@ -67,12 +71,19 @@ func (s *proxyServices) servicesToStart() (servicesToStart []internal.Service) {
...
 
 	// TCP Plugins
 	for _, cfg := range configsByType.TCP {
+		// If we're a unix socket listener, delete the old file first
+		err := s.deleteSocketIfExists(cfg.ListenOn)
+		if err != nil {
+			// TODO: Add Fatalf to our logger and use that
+			s.logger.Panic(err)
+		}
+
```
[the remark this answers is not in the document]
sgnn7: > How come we delete the unix sockets but not tcp ones?

you can delete a unix socket file but a tcp socket doesn't really exist in the same way. the fs-socket remove just unlinks it (it still exists) and you create a new one that the new fs view will present to new listeners. with tcp sockets, you _have_ to wait for it to be released (afaik at least).

> Should we warn?

yes


> What's the correct behavior if tcp ones exist?

you have to wait or kill the parent process - there's really no other way that I know of (but I haven't looked into this deeply)

pkg/secretless/entrypoint/entrypoint.go
sgnn7: heh - I moved the var outside of the config loop but forgot to actually add the reason why I did that. good catch!

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,49 +53,77 @@ func StartSecretless(params *SecretlessOptions) {
...
+	}
 
-	allServices := proxyservice.NewProxyServices(cfg, availPlugins, logger, evtNotifier)
 	exitListener.AddHandler(func() {
-		fmt.Println("wait for all services signal")
+		fmt.Println("Received a stop signal")
+		if allServices == nil {
+			return
+		}
+
```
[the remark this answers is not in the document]
sgnn7: Hmm - I'll rework this a bit.

internal/configurationmanagers/configfile/manager.go
```diff
@@ -159,3 +170,24 @@ func ConfigurationManagerFactory(options plugin_v1.ConfigurationManagerOptions)
...
+	manager := &configurationManager{}
+
+	cfgChangeHandler := changeHandler{
+		ConfigChangeChan: configChangedChan,
+	}
+
+	// Managers expect parameters to be passed in as URL params
+	cfgSpec := fmt.Sprintf("%s?watch=%t", configfile, fsWatchEnabled)
+
+	err := manager.Initialize(&cfgChangeHandler, cfgSpec)
```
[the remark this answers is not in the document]
sgnn7: Deferred

pkg/secretless/entrypoint/entrypoint.go
sgnn7: TODO added

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,49 +53,88 @@ func StartSecretless(params *SecretlessOptions) {
...
 	// Health check: Initialized
 	util.SetAppInitializedFlag()
 	util.SetAppIsLive(false)
 
-	err = allServices.Start()
-	if err != nil {
-		log.Fatalln(err)
+	var allServices internal.Service
+
+	configReload := func(cfg v2.Config) {
```
jonahx: `configReload` -> `reloadConfig`

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,49 +53,88 @@ func StartSecretless(params *SecretlessOptions) {
...
-		}
-	})
+	configChangedChan, err := newConfigChangeChan(
+		params.ConfigFile,
+		params.ConfigManagerSpec,
+		params.FsWatchEnabled,
+	)
+	if err != nil {
+		log.Fatalln(err)
+	}
```
jonahx: Put the health check stuff 👇 above this channel block ☝️ .  

Otherwise it's like we're starting on the topic of config reloading, randomly mentioning health checks, then going back to talking about config reloading. 

Also add section comments (comments with blank line above and below) for each of the config pieces:

```
// Create channel new configs

// Create config reload function 
//   --> this can become a single line with the function pulled out

// etc
```

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -50,49 +53,88 @@ func StartSecretless(params *SecretlessOptions) {
...
+	exitListener.AddHandler(func() {
+		fmt.Println("Received a stop signal")
+
+		if allServices != nil {
+			err := allServices.Stop()
+			if err != nil {
+				// Log but but allow cleanup of other subscribers to continue.
+				log.Println(err)
+			}
+		}
```
jonahx: Line of sight again: I'd prefer repeating the `os.Exit`, ie
```go
if allServices == nil {
  os.Exit()
}
err := allServices.Stop()
if err != nil {
  // Log but but allow cleanup of other subscribers to continue.
  log.Println(err)
}
os.Exit()

File: pkg/secretless/entrypoint/entrypoint.go [hunk 10]
```diff
@@ -50,49 +53,88 @@ func StartSecretless(params *SecretlessOptions) {
...
+
+		// TODO: Ideally we would soft-close all goroutines rather than rely on the
+		//       heavy-handed os.Exit to exit the broker when we want to.
+		os.Exit(0)
+	})
+
+	go func() {
```
jonahx: Add comment section above this `go func` like 
```go
// Listen for and restart on config changes
```

pkg/secretless/entrypoint/entrypoint.go
jonahx: Optional: Per slack convo, I think this whole function ☝️ would be a good candidate to pull out into it's own function.

You'd only have to pass in `allServices`.  Speaking of slack convos and the "all is a name smell" (yes i know I named it!), wdyt of `allServices` -> `services`?

sgnn7: This seems less readable to me but i'll change it up

sgnn7: I like the change but it will have to wait given the time pressure :(

PR #945, opened 2019-10-21
Revise integration tests to use existing images

izgeri:
At current, a handful of integration tests assume you'll build the needed
image at test time instead of using an existing image built using bin/build

This commit revises it so that all integration tests refer to images built
on top of an existing image (eg secretless-broker or secretless-dev) or uses
an existing image directly.


izgeri:
Closing, since #946 appears to have fixed the flapping test failures for now.

PR #964, opened 2019-10-24
Add mssql connector

jonahx:
This adds an end-to-end, working POC of the MSSQL connector, including a basic integration test.

Future work to make it more robust:

  - Add more integration tests
    - The current test suite includes:
      - Application can connect to MSSQL through Secretless without creds in the request
      - Application cannot connect directly to MSSQL without creds in the request
  - Add UTs
  - Verify the PRELOGIN response is created correctly
  - Test the OK message with different TDS versions and Server versions
  - Build the OK message the way we do it PG
  - Check if we can extract Login Ack response details
  - Run a security review of the 3rd party (https://github.com/denisenkom/go-mssqldb)

All the code changes requirements have TODOs in the code.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/cyberark/conjur-api-go v0.5.2
 	github.com/cyberark/conjur-authn-k8s-client v0.13.0
 	github.com/cyberark/summon v0.7.0
+	github.com/denisenkom/go-mssqldb v0.0.0-20191001013358-cfbb681360f0
 	github.com/fsnotify/fsnotify v1.4.7
 	github.com/ghodss/yaml v1.0.0 // indirect
 	github.com/go-ozzo/ozzo-validation v0.0.0-20170913164239-85dcd8368eba
@@ -34,7 +35,7 @@ require (
 	github.com/smartystreets/goconvey v0.0.0-20190330032615-68dc04aab96a
 	github.com/spf13/pflag v1.0.2 // indirect
 	github.com/stretchr/testify v1.3.0
-	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
+	golang.org/x/crypto v0.0.0-20190325154230-a5d413f7728c
 	gopkg.in/DATA-DOG/go-sqlmock.v1 v1.3.0 // indirect
 	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
 	gopkg.in/fsnotify.v1 v1.4.7 // indirect
@@ -48,3 +49,5 @@ require (
 	k8s.io/client-go v0.0.0-20180806134042-1f13a808da65
 	k8s.io/kube-openapi v0.0.0-20180731170545-e3762e86a74c // indirect
 )
+
+replace github.com/denisenkom/go-mssqldb => github.com/cyberark/go-mssqldb v0.0.0-20191030142036-b5a965a47dd3
--- a/internal/plugin/connectors/tcp/mssql/authentication.go
+++ b/internal/plugin/connectors/tcp/mssql/authentication.go
@@ -1,0 +1,165 @@
+package mssql
+
+import (
+	"fmt"
+
+	mssql "github.com/denisenkom/go-mssqldb"
+)
+
+func (connector *SingleUseConnector) performPreLoginHandshake() error {
+	// using the default packet size of 4096
+	// (see https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/configure-the-network-packet-size-server-configuration-option)
+	clientBuffer := mssql.NewTdsBuffer(4096, connector.clientConn)
+	preloginRequest, err := mssql.ReadPreloginWithPacketType(clientBuffer, mssql.PackPrelogin)
+	if err != nil {
+		return fmt.Errorf("failed to read prelogin request: %s", err)
+	}
+
+	preloginResponse := preparePreloginResponse(preloginRequest)
+
+	err = mssql.WritePreloginWithPacketType(clientBuffer, preloginResponse, mssql.PackReply)
+	if err != nil {
+		return fmt.Errorf("failed to write prelogin response: %s", err)
+	}
+
+	// we actually don't need the client's handshake response.
+	// we just need for them to not be blocked
+	err = clientBuffer.ReadNextPacket()
+	if err != nil {
+		return fmt.Errorf("failed to read client login message: %s", err)
+	}
+
+	return nil
+}
+
+func preparePreloginResponse(preLoginRequest map[uint8][]byte) map[uint8][]byte {
+	// PRELOGIN message docs - https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/60f56408-0188-4cd5-8b90-25c6f2423868
+
+	// The prelogin response message is identical to the request with minor changes
+	preloginResponse := preLoginRequest
+
+	// According to the mssql docs, The client can use the VERSION returned from
+	// the server to determine which features SHOULD be enabled or disabled.
+	// TODO: Extract version from server instead of hard-coded one.
+	// we use now the version of the sql server in the test
+	preloginResponse[mssql.PreloginVERSION] = []byte{0x0e, 0x00, 0x0c, 0xa6, 0x00, 0x00}
+
+	// Remove Client SSL Capability from Server Handshake Packet
+	// to force client to connect to Secretless without SSL
+	preloginResponse[mssql.PreloginENCRYPTION] = []byte{mssql.EncryptNotSup}
+
+	// According to the docs, this value should be empty when being sent from
+	// the server to the client.
+	preloginResponse[mssql.PreloginTHREADID] = []byte{}
+
+	// TODO: verify the rest of the message stays the same
+
+	return preloginResponse
+}
+
+// CreateAuthenticationOKMessage creates an MSSQL message which indicates
+// successful authentication.
+func (connector *SingleUseConnector) CreateAuthenticationOKMessage() []byte {
+	// The packet format - https://www.freetds.org/tds.html#packet
+	// The login ack response - https://www.freetds.org/tds.html#t173
+
+	// TODO: This message should stay static but it's better to build it the same way
+	// we do in the pg connector -
+	// https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/tcp/pg/protocol/auth.go#L150
+
+	// TODO: Also check if we can (and need to) use the actual TDS version & server
+	//  details using go-mssqldb.
+	//  If so, we need to merge this PR - https://github.com/cyberark/go-mssqldb/pull/3
+
+	// Create a hard-coded OK response message
+	message := []byte{
+		// header
+		// [ REPLY packet type, last packet indicator, packet size (2 bytes),
+		// channel (2 bytes, can be zeroed), packet number, window (can be zeroed) ]
+		0x04, 0x01, 0x00, 0x4E, 0x00, 0x00, 0x01, 0x00,
+
+		// LoginAck Token
+		// Login ack indicator
+		0xad,
+		// Packet length - 54 bytes from the next one until the Done Token (not including)
+		0x36, 0x00,
+		// ack success - TODO: test with TDS 5.0, because there the success ack is 0x05
+		0x01,
+		// TDS version - TODO: verify that it can be hard coded regardless to
+		//  the actual TDS version of the server
+		0x74, 0x00, 0x00, 0x04,
+		// server name length - 22 chars
+		0x16,
+		// server name - 'Microsoft SQL Server'
+		0x4d, 0x00, 0x69, 0x00, 0x63, 0x00, 0x72, 0x00, 0x6f, 0x00, 0x73, 0x00,
+		0x6f, 0x00, 0x66, 0x00, 0x74, 0x00, 0x20, 0x00, 0x53, 0x00, 0x51, 0x00,
+		0x4c, 0x00, 0x20, 0x00, 0x53, 0x00, 0x65, 0x00, 0x72, 0x00, 0x76, 0x00,
+		0x65, 0x00, 0x72, 0x00, 0x00, 0x00, 0x00, 0x00,
+		// server version - TODO: verify that it can be hard coded regardless
+		//  to the actual server version
+		0x0e, 0x00, 0x0c, 0xa6,
+
+		// Done Token - indicates the end of the packet
+		0xfd, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	}
+
+	return message
+}
+
+// CreateGenericErrorMessage creates an MSSQL error message
+func (connector *SingleUseConnector) CreateGenericErrorMessage() []byte {
+	// The packet format - https://www.freetds.org/tds.html#packet
+	// The Error token - https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/9805e9fa-1f8b-4cf8-8f78-8d2602228635
+
+	// TODO: Create the error object with go-mssqldb. In order to use it, we should
+	// add there a function to convert the object to a byte array
+
+	// Create a hard-coded error response message
+	message := []byte{
+		// header
+		// [ REPLY packet type, last packet indicator, packet size (2 bytes),
+		// channel (2 bytes, can be zeroed), packet number, window (can be zeroed) ]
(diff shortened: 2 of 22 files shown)


go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/cyberark/conjur-api-go v0.5.2
 	github.com/cyberark/conjur-authn-k8s-client v0.13.0
 	github.com/cyberark/summon v0.7.0
+	github.com/denisenkom/go-mssqldb v0.0.0-00010101000000-000000000000
```
sgnn7: This dependency needs LICENSES check/update.

CC: @izgeri

go.mod
```diff
@@ -48,3 +49,5 @@ require (
 	k8s.io/client-go v0.0.0-20180806134042-1f13a808da65
 	k8s.io/kube-openapi v0.0.0-20180731170545-e3762e86a74c // indirect
 )
+
+replace github.com/denisenkom/go-mssqldb => github.com/orenbm/go-mssqldb v0.0.0-20191023140747-b0d280c79777
```
sgnn7: This needs to be cyberark/conjur org hosted repo

internal/plugin/connectors/tcp/mssql/authentication_handshake.go
```diff
@@ -0,0 +1,132 @@
...
+	clientBuf := mssql.NewTdsBuffer(4096, clientConn)
+	msg, err := ReadClientPrelogin(clientBuf)
+	if err != nil {
+		return fmt.Errorf("bad prelogin: %s", err)
+	}
+
+	// based on wireshark, the prelogin response from the server is almost identical
+	// to the login message from the client, with minor changes as below:
+	msg[mssql.PreloginTHREADID] = []byte{}
+	msg[mssql.PreloginVERSION] = []byte{0x0e, 0x00, 0x0c, 0xa6, 0x00, 0x00}
```
sgnn7: It would be good to comment on what this binary string is composed of in terms of fields or its string value

internal/plugin/connectors/tcp/mssql/authentication_handshake.go
```diff
@@ -0,0 +1,132 @@
...
+	// we just need for them to not be blocked
+	err = clientBuf.ReadNextPacket()
+	if err != nil {
+		return fmt.Errorf("bad client login read: %s", err)
+	}
+
+	// created an unencrypted connection to mssql, and just copied the raw bytes
+	// this has loginack, and done tokens.
+	_, err = clientConn.Write([]byte{
+		0x04, 0x01, 0x00, 0x64, 0x00, 0x37, 0x01, 0x00, 0xad, 0x36, 0x00, 0x01, 0x74, 0x00, 0x00, 0x04, 0x16, 0x4d, 0x00, 0x69, 0x00, 0x63, 0x00, 0x72, 0x00, 0x6f, 0x00, 0x73, 0x00, 0x6f, 0x00, 0x66, 0x00, 0x74, 0x00, 0x20, 0x00, 0x53, 0x00, 0x51, 0x00, 0x4c, 0x00, 0x20, 0x00, 0x53, 0x00, 0x65, 0x00, 0x72, 0x00, 0x76, 0x00, 0x65, 0x00, 0x72, 0x00, 0x00, 0x00, 0x00, 0x00, 0x0e, 0x00, 0x0c, 0xa6, 0xe3, 0x13, 0x00, 0x04, 0x04, 0x34, 0x00, 0x30, 0x00, 0x39, 0x00, 0x36, 0x00, 0x04, 0x34, 0x00, 0x30, 0x00, 0x39, 0x00, 0x36, 0x00, 0xfd, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```
sgnn7: this for sure needs a breakdown in what's being written in terms of serialized fields and a link to docs

internal/plugin/connectors/tcp/mssql/authentication_handshake.go
```diff
@@ -0,0 +1,132 @@
...
+		rec_type := struct_buf[offset]
+		if rec_type == mssql.PreloginTERMINATOR {
+			break
+		}
+
+		rec_offset := binary.BigEndian.Uint16(struct_buf[offset+1:])
+		rec_len := binary.BigEndian.Uint16(struct_buf[offset+3:])
+		value := struct_buf[rec_offset : rec_offset+rec_len]
+		results[rec_type] = value
+		offset += 5
```
sgnn7: This block needs comments on parts

internal/plugin/connectors/tcp/mssql/authentication_handshake.go
```diff
@@ -0,0 +1,132 @@
...
+		v := fields[k]
+		written, err := w.Write(v)
+		if err != nil {
+			return err
+		}
+		if written != len(v) {
+			return errors.New("Write method didn't write the whole value")
+		}
+	}
+	return w.FinishPacket()
```
sgnn7: This block also needs comments on what's being done

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -0,0 +1,61 @@
...
+	//
+	// NOTE: Secretless has some unfortunate naming collisions with the
+	// go-mssqldb driver package.  The driver package has its own concept of a
+	// "connector", and its connectors also have a "Connect" method.
+	driverConnector, err := mssql.NewConnector(
+		fmt.Sprintf(
+			"sqlserver://%s:%s@%s",
+			credentialValuesByID["user"],
+			credentialValuesByID["password"],
+			credentialValuesByID["address"],
```
sgnn7: Do these need escaping? If so, we either need a fix for it or a new issue opened

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, err
+	}
+
+	// Verify the driverConn is an mssql driverConn object and get its underlying transport
+	mssqlConn := driverConn.(*mssql.Conn)
+	backEndConn := mssqlConn.NetConn()
+
+	connector.logger.Infof("backend driverConn ready")
+
+	// TODO: add a comment on this - why do we need it?
```
sgnn7: +1

internal/plugin/connectors/tcp/mssql/plugin.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
+)
+
+// NewConnector returns a tcp.Connector which returns an authenticated connection
+// to a target service for each incoming client connection.
+// It is a required method on the tcp.Plugin interface.
+// The single argument passed in is of type connector.Resources. It contains
+// connector-specific config and a logger.
```
sgnn7: Weird line wrap on the comment sentences

go.mod
izgeri: I'm glad to see it here, because for prior connectors we haven't had it be this obvious.

it's good to get notified of it here in the PR, but we don't need to make the fix in practice until we're ready for the next tag. I expect that will be within 2 weeks, to include whatever fixes we discover are needed as part of perf testing and the initial MSSQL connector.

go.mod
izgeri: this is probably true, if we want to use it in the actual broker now. let me look into how we can get this going.

internal/plugin/connectors/tcp/mssql/authentication_handshake.go
brikelly: Given the amount of magic constant values here, it would be good to add a comment to indicate that this is going to be eventually changed to be done at runtime rather than just hardcoded within.

go.mod
izgeri: I moved the content from @orenbm's fork into https://github.com/cyberark/go-mssqldb after reviewing the upstream repo and noting that there has been no activity by the maintainer this month. strong preference to contribute our changes upstream to the project (along with whatever improvements we can help make after leveraging the library ourselves), but for now if we have to use a fork let's use our official one.

orenbm:
@sgnn7 @brikelly we didn't add too many comments explaining what we are doing as this code will change in a future PR (as stated in the PR description).

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -0,0 +1,111 @@
...
+
+	err = connector.performPreLoginHandshake()
+	if err != nil {
+		connector.logger.Errorf("Failed to handle client prelogin request: %s", err)
+		return nil, err
+	}
+
+	// Now that the backend connection is authenticated, send the client
+	// a successful authentication response.
+	if _, err = clientConn.Write(connector.CreateAuthenticationOKMessage()); err != nil {
```
doodlesbykumbi: There seems to be some logic missing for sending a not-OK message to the client. I'm not sure how this connector is currently propagating errors to the client, if at all.

orenbm: oh i see that now. good catch!

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Verify the driverConn is an mssql driverConn object and get its underlying transport
+	mssqlConn := driverConn.(*mssql.Conn)
+	connector.backendConn = mssqlConn.NetConn()
+
+	return nil
+}
+
+func validateCredentialsExist(credentialValuesByID connector.CredentialValuesByID) error {
+	requiredCredentials := []string{"user", "password", "address"}
```
izgeri: can we switch this to host/port? we recently moved pg to host/port, and it would be mega-nice to maintain this standard

for consistency, the config would be `host` / `port` / `username` / `password`, and the `port` could be optional and default to the DB's typical default value

internal/plugin/connectors/tcp/mssql/connector.go
orenbm: added code to send the client a generic error. we should also send the actual error but i didn't get to that

internal/plugin/connectors/tcp/mssql/connector.go
orenbm: Done, thanks for the input!

internal/plugin/connectors/tcp/mssql/authentication.go
```diff
@@ -0,0 +1,156 @@
...
+	preloginResponse := preLoginRequest
+
+	// According to the mssql docs, The client can use the VERSION returned from
+	// the server to determine which features SHOULD be enabled or disabled. The
+	// client SHOULD do this only if it can determine whether a feature is
+	// supported by the current version of the database.
+	//
+	// In our case, the client will not use the VERSION so we don't need the actual
+	// version of the server
+	preloginResponse[mssql.PreloginVERSION] = []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00}
```
doodlesbykumbi: As you mentioned this version is necessary fr the client to determine which features are available on the server. We need to dutifully transmit whatever comes from the server.

When I run `./start` in the test folder and wait for mssql to come up, then run:
```
sqlcmd -S "127.0.0.1,2223" -U x -P "x"
```

It results in:
```
Sqlcmd: Error: Microsoft ODBC Driver 17 for SQL Server : ODBC Driver 17 for SQL Server does not support connections to SQL Server 2000 or earlier versions..
Sqlcmd: Error: Microsoft ODBC Driver 17 for SQL Server : Client unable to establish connection.
```

This is 100% a result of PreloginVERSION. Connecting directly works fine.

orenbm: very nice catch

test/mssql_connector/wait_for_mssql
```diff
@@ -0,0 +1,66 @@
...
+        count=$((count + 1))
+        if [ "${count}" -lt "${retries}" ]; then
+            # There are still retries left, calculate delay and notify user.
+            backoff=$((2 * count))
+            if [[ "${backoff}" -gt "${MAX_BACKOFF}" ]]; then
+                backoff=${MAX_BACKOFF}
+            fi;
+
+            # Add a random amount to the delay to prevent competing processes
+            # from re-colliding.
```
jonahx: This comment isn't clear to me, whatever the reason is that we're doing this.... since the delay is random, it could only make what we're trying to avoid slightly less likely.  It can't prevent it.

However, that is a kind of a moot point for me because I have a larger issue:

Imo using exponential backoff here is unnecessary complexity for little to no benefit.  I think a simpler strategy with less maintenance burden would just be:

1. Decide on a max time we give up after.  Say, 30s.
2. Run the `wait_for_mssql` command every second or half-second until it succeeds or we reach the max time.

internal/plugin/connectors/tcp/mssql/authentication.go
```diff
@@ -0,0 +1,161 @@
...
+	preloginResponse := preLoginRequest
+
+	// According to the mssql docs, The client can use the VERSION returned from
+	// the server to determine which features SHOULD be enabled or disabled. The
+	// client SHOULD do this only if it can determine whether a feature is
+	// supported by the current version of the database.
+	//
+	// In our case, the client will not use the VERSION so we don't need the actual
+	// version of the server
+	preloginResponse[mssql.PreloginVERSION] = []byte{0x0e, 0x00, 0x0c, 0xa6, 0x00, 0x00}
```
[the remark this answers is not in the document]
orenbm: good call, this comment is no longer correct. we found out that the version is crucial. i'l change the comment and add a TODO

internal/plugin/connectors/tcp/mssql/authentication.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Remove Client SSL Capability from Server Handshake Packet
+	// to force client to connect to Secretless without SSL
+	preloginResponse[mssql.PreloginENCRYPTION] = []byte{mssql.EncryptNotSup}
+
+	// This value should be empty when being sent from the server to the client.
```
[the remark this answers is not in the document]
orenbm: they just state that in the docs. I don't want to further investigate this now

internal/plugin/connectors/tcp/mssql/authentication.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// TODO: Create the error object with go-mssqldb. In order to use it, we should
+	// add there a function to convert the object to a byte array
+
+	// Create a hard-coded error response message
+	message := []byte{
+		// header
+		// [ REPLY packet type, last packet indicator, packet size (2 bytes),
+		// channel (2 bytes, can be zeroed), packet number, window (can be zeroed) ]
+		0x04, 0x01, 0x00, 0x7a, 0x00, 0x33, 0x01, 0x00,
```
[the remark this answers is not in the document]
orenbm: it is not repeated, the headers are different in both messages. anyway, the beta code will not have hard-coded bytes but will extract actual data from the server so this code ill change anyway.

orenbm:
@jonahx i have no idea what you are talking about here: https://github.com/cyberark/secretless-broker/pull/964#discussion_r342642659

the `Connect` method sends an error to the client (`connector.sendErrorToClient()`) every time we catch one

test/mssql_connector/wait_for_mssql
```diff
@@ -0,0 +1,66 @@
...
+# Delay is 2 * n + random up to 30s, then 30s + random after that.
+# For large numbers of retries the max delay is effectively the retry
+# in minutes.
+# Based on:
+# https://gist.github.com/sj26/88e1c6584397bb7c13bd11108a579746
+# but now quite heavily modified.
+function retry {
+    # Maxiumum amount of fixed delay between attempts
+    # a random value will still be added.
+    local -r MAX_BACKOFF=30
```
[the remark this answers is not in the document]
orenbm: this function was copy-pasted from Hugh's `bash-lib`. didn't go into details too much.

test/mssql_connector/wait_for_mssql
orenbm: this function was copy-pasted from Hugh's `bash-lib`. didn't go into details too much.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -0,0 +1,111 @@
...
+	logger      log.Logger
+}
+
+// Connect implements the tcp.Connector func signature
+//
+// It is the main method of the SingleUseConnector. It:
+//   1. Constructs connection details from the provided credentials map.
+//   2. Dials the backend using credentials.
+//   3. Runs through the connection phase steps to authenticate.
+//   4. Pipes all future bytes unaltered between client and server.
```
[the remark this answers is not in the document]
jonahx: @orenbm Looks like this one slipped through the cracks?

PR #970, opened 2019-10-25
useful logs around failure to start services

doodlesbykumbi:


internal/proxyservice/proxy_service.go
```diff
@@ -127,12 +153,18 @@ func (s *proxyServices) servicesToStart() (servicesToStart []internal.Service) {
...
+				err)
+			continue
 		}
 		servicesToStart = append(servicesToStart, sshAgentSvc)
 	}
 
-	// TODO: Deal with SSH in a hardcoded way
+	if err := errors.Filter(); err != nil {
+
+		s.logger.Panicf("unable to start services using collection of plugins (%v): %s", strings.Join(allPlugins, ","), err)
```
sgnn7: Line too long. Also there's a stray LF above here.

pkg/secretless/plugin/sharedobj/external_plugins.go
```diff
@@ -136,43 +136,44 @@ func ExternalPlugins(
 	for rawPluginName, rawPlugin := range rawPlugins {
 		logger.Infof("Loading plugin '%s'...", rawPluginName)
 
+		pluginLogger := logger.CopyWith(rawPluginName, logger.DebugEnabled())
```
jonahx: So this is a pretty serious code smell with our design.  Can you give me a little more context about the reason for this?  The whole point of injecting a logger is so that the code using it only has to call logging methods.  It should have _no knowledge_ of how the logger is configured. 

To this another way: These are error happening inside the `ExternalPlugins` func.  But we're making them _look_ like errors from inside of something else. 

I suspect that there's a deeper design issue here that you're hacking around to solve, but I need a bit more context to tell what it is.

doodlesbykumbi: @jonahx I'm just using this other logger to prepend the errors with the plugin name

jonahx: @doodlesbykumbi So this is not how prefixes are intended to be used.  As I said, this makes it appear that the code is _coming from within_ the particular plugins.  In this case, it looks like you want to use simple error wrapping.

There are [some new facillities for doing this](https://blog.golang.org/go1.13-errors):

> In Go 1.13, the fmt.Errorf function supports a new %w verb. When this verb is present, the error returned by fmt.Errorf will have an Unwrap method returning the argument of %w, which must be an error. In all other ways, %w is identical to %v.
> ```go 
> if err != nil {
>     // Return an error which unwraps to err.
>     return fmt.Errorf("decompress %v: %w", name, err)
> }
> ```

internal/proxyservice/proxy_service.go
```diff
@@ -34,8 +36,51 @@ type proxyServices struct {
...
+		allPluginsMap[p] = true
+	}
+
+	errors := validation.Errors{}
+	for _, service := range s.config.Services {
+		connectorExists := allPluginsMap[service.Connector]
+		if !connectorExists {
+			errors[service.Name] = fmt.Errorf(
+				`missing service connector "%s"`,
+				service.Connector)
```
jonahx: nit: I noticed you've been sometimes putting closing parens on same line, and sometimes on their own line (eg, line 51 above).  I think we should be consistent about this one way or the other.  I don't care too much which way, but it seems like we've been doing own line mostly in the code base, and it seems more consistent with Go style generally (eg, closing brace on own line).  Whereas the above feels more python-y.

internal/proxyservice/proxy_service.go
```diff
@@ -34,8 +36,51 @@ type proxyServices struct {
...
+
+	s.logger.Infof(
+		"Validating services against available plugins: %s",
+		strings.Join(allPlugins, ","),
+	)
+
+	allPluginsMap := map[string]bool{}
+	for _, p := range allPlugins  {
+		allPluginsMap[p] = true
+	}
```
jonahx: I had to do a minor double-take to figure out that `allPluginsMap` existed solely so that we don't have to have a nested loop within the loop below to check for existence.  This is a good reason to have it, but a small comment here would help readers know what it's for without having to read the code below.  Something like:

```go
// Convert available plugins to a map, so that we can check for their
// existence in the loop below using a map lookup rather than a nested
// loop, which would be required otherwise.

File: pkg/secretless/plugin/sharedobj/external_plugins.go [hunk 5]
```diff
@@ -136,43 +136,57 @@ func ExternalPlugins(
 	for rawPluginName, rawPlugin := range rawPlugins {
 		logger.Infof("Loading plugin '%s'...", rawPluginName)
 
+		reportLoadPluginError := func(err error) {
```
jonahx: Naming nit: `reportLoadPluginError` -> `logPluginError`.  

No need to introduce a new verb `report` when we already have `log` and that's what we use everywhere.

PR #974, opened 2019-10-30
improve pg error propagation

doodlesbykumbi:
Instead of giving the client a useless error of the form:
```
psql: FATAL:  Expected 82 message type, got 69
```
This PR makes it so that pg errors are faithfully propagated to the client:

```
psql: FATAL:  password authentication failed for user "test"
```

### DIFF
--- a/internal/plugin/connectors/tcp/pg/connector.go
+++ b/internal/plugin/connectors/tcp/pg/connector.go
@@ -25,14 +25,22 @@ type SingleUseConnector struct {
 }
 
 func (s *SingleUseConnector) abort(err error) {
-	if s.clientConn != nil {
-		pgError := protocol.Error{
+	if s.clientConn == nil {
+		return
+	}
+
+	pgError, ok := err.(*protocol.Error)
+	// Not an expected protocol error, so we wrap it in a protocol error, so
+	// the client can see what went wrong.
+	if !ok {
+		pgError = &protocol.Error{
 			Severity: protocol.ErrorSeverityFatal,
 			Code:     protocol.ErrorCodeInternalError,
 			Message:  err.Error(),
 		}
-		s.clientConn.Write(pgError.GetPacket())
 	}
+
+	s.clientConn.Write(pgError.GetPacket())
 }
 
 // Connect implements the tcp.Connector func signature.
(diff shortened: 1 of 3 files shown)


internal/plugin/connectors/tcp/pg/protocol/error.go
```diff
@@ -56,6 +59,63 @@ func (e *Error) Error() string {
...
+
+	for {
+		fieldType, readErr = b.ReadByte()
+		if readErr != nil {
+			if readErr == io.EOF {
+				break
+			}
+
+			return fmt.Errorf(unparseableErr)
+		}
```
jonahx: Line of sight:
```go
// We're done reading
if readErr == io.EOF {
    break
}
// Unexpected error
if readErr != nil {
    return fmt.Errorf(unparseableErr)
}

File: internal/plugin/connectors/tcp/pg/protocol/error.go [hunk 2]
```diff
@@ -56,6 +59,63 @@ func (e *Error) Error() string {
...
+
+		fieldContents, readErr = b.ReadString()
+		if readErr != nil {
+			if readErr == io.EOF {
+				break
+			}
+
+			return fmt.Errorf(unparseableErr)
+		}
```
jonahx: Same here.

internal/plugin/connectors/tcp/pg/connector.go
```diff
@@ -26,11 +26,15 @@ type SingleUseConnector struct {
...
-			Message:  err.Error(),
+		pgError, ok := err.(*protocol.Error)
+		if !ok {
+			pgError = &protocol.Error{
+				Severity: protocol.ErrorSeverityFatal,
+				Code:     protocol.ErrorCodeInternalError,
+				Message:  err.Error(),
+			}
 		}
+
```
jonahx: Line of sight:
```go
func (s *SingleUseConnector) abort(err error) {
	if s.clientConn == nil {
		return
	}

	pgError, ok := err.(*protocol.Error)
	// Not an expected protocol error, so we wrap it in a protocol error, so
	// the client can see what went wrong.
	if !ok {
		pgError = &protocol.Error{
			Severity: protocol.ErrorSeverityFatal,
			Code:     protocol.ErrorCodeInternalError,
			Message:  err.Error(),
		}
	}

	s.clientConn.Write(pgError.GetPacket())
}
```

Issue #901, opened 2019-09-23
Unit Tests exist for Secretless profile package


jonahx:
## Description

The `profile` package is a thin wrapper over `github.com/pkg/profile`, and all the interesting work is delegated to that.   Our tests should reflect this: The right strategy here is to create:

```go
type StartProfilingFunc func Start(options ...func(*Profile)) interface { Stop() }

func NewWithOptions(fn StartProfilingFunc, profileType string) secretless.Service {
  // actual creation
}

func New(profileType string) secretless.Service {
  // just delegates to NewWithOptions, passing `profile.Start` 
  // (from pkg) as the `StartProfilingFunc`
}
```

AC:
- [ ] Constructors and types per above
- [ ] Tests using mocks to ensure `StartProfilingFunc` is called when expected
- [ ] Tests for the validation logic via ctor
- [ ] Direct tests of validation logic via `ValidateType` function

PR #981, opened 2019-11-01
901 profile tests

sgnn7:
Refactors the profile class to have simpler usage of the API and adds tests for the class.

Connected to #901 

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--secretless-broker/job/901-profile-tests/)

### DIFF
--- a/internal/profile/profile.go
+++ b/internal/profile/profile.go
@@ -2,40 +2,42 @@ package profile
 
 import (
 	"fmt"
-	"strings"
 
-	"github.com/cyberark/secretless-broker/internal"
 	"github.com/pkg/profile"
 )
 
-func (pp *perfProfile) Start() error {
-	switch pp.profileType {
-	case "cpu":
-		pp.stopper = profile.Start(profile.NoShutdownHook)
-	case "memory":
-		pp.stopper = profile.Start(profile.MemProfile, profile.NoShutdownHook)
-	default:
-		// will be impossible when New is used as ctor
-		return fmt.Errorf("attempt to start profiling with invalid profileType")
-	}
-	return nil
+// ProfilingService is the interface to this class' functionality.
+type ProfilingService interface {
+	Start()
+	Stop()
 }
 
-func (pp *perfProfile) Stop() error {
+// StartProfilingFunc is a function that will be invoked when profile start
+// is requested. This method signature is just a type wrapper around the
+// `Start()` signature from `github.com/pkg/profile`.
+type StartProfilingFunc func(...func(*profile.Profile)) interface{ Stop() }
+
+// Start is used to begin the profiling
+func (pp *perfProfile) Start() {
+	pp.stopper = pp.startProfiling(pp.profileOptions...)
+}
+
+// Start is used to end the profiling
+func (pp *perfProfile) Stop() {
 	pp.stopper.Stop()
-	return nil
 }
 
 type perfProfile struct {
-	profileType string
-	stopper interface { Stop() }
+	profileOptions []func(*profile.Profile)
+	startProfiling StartProfilingFunc
+	stopper        interface{ Stop() }
 }
 
-// ValidTypes are the valid types of profiling you can perform.
-var ValidTypes = []string{"cpu", "memory"}
+// validTypes are the valid types of profiling you can perform.
+var validTypes = []string{"cpu", "memory"}
 
 func isValidType(profileType string) bool {
-	for _, curType := range ValidTypes {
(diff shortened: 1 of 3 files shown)


internal/profile/profile_test.go
```diff
@@ -0,0 +1,79 @@
...
+		startEventChan := make(chan struct{}, 10)
+		stopEventChan := make(chan struct{}, 10)
+
+		mockProfiler := func(profiles ...func(*gh_profile.Profile)) interface{ Stop() } {
+			startEventChan <- struct{}{}
+
+			return &stopper{
+				eventChan: stopEventChan,
+			}
+		}
```
jonahx: There's a (imo slightly cleaner) way to do this that avoids the `stopper` struct (which is A. not needed and B. a little misleading in that the `-er` naming is typically used for interfaces, whereas this is an impl struct).

In any case, we already have 
```go
type StartProfilingFunc func(...func(*profile.Profile)) interface{ Stop() }
```
and we can complete the symmetry here with:
```go
type StopProfilingFunc func()
```
and then use the "function type calling itself" trick to make it conform to `interface{ Stop() }`:
```go
func (stop StopProfilingFunc) Stop() {
    stop()
}
```
Now we may elimnate the current `stopper` struct, and instead rely on a closure to get hold of our `stopEventChan`, and replace lines 52-54 with:
```go
stop := func() {
    stopEventChan <- struct{}{}
}
return StopProfilingFunc(stop)
```
POC of the concept: https://play.golang.org/p/Zfw3bJvnUP_Q

internal/profile/profile_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+		assert.Equal(t, 1, len(startEventChan))
+		assert.Equal(t, 0, len(stopEventChan))
+
+		profile.Stop()
+
+		assert.Equal(t, 1, len(startEventChan))
+		assert.Equal(t, 1, len(stopEventChan))
+	})
+}
```
jonahx: Per our conversation earlier today:

> I definitely think “only use channels when you have to” is good advice

I think we don't need channels here.  All we're testing is that the `Start()` func we're delegating to does in fact get called (same with `Stop()`).   And this is definitely the right way to do this test. 

But instead of closing over `startEventChan` and `stopEventChan` to capture that info, you can close over simple boolean flags `startWasCalled` and `stopWasCalled` and then assert on them directly.

internal/profile/profile.go
```diff
@@ -2,64 +2,87 @@ package profile
...
+// Start is used to end the profiling
 func (pp *perfProfile) Stop() error {
 	pp.stopper.Stop()
 	return nil
 }
 
 type perfProfile struct {
-	profileType string
-	stopper interface { Stop() }
+	profiles           []func(*profile.Profile)
```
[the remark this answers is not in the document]
sgnn7: Neither of those two alternatives describe adequately that they are all specifically of `func(*profile.Profile)` type so a generic `options` doesn't really fit here and `functionalOptions` is also too generic of a name when we have such a consistent vararg and we hint at what's being held there. How about `profilingFuncs`, `profilers`, or something with `profile` in the name?

Also:

internal/profile/profile_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+		for _, profileType := range []string{"foo", "bar"} {
+			profile, err := profile.New(profileType)
+
+			assert.Nil(t, profile)
+
+			// Sanity check
+			assert.NotNil(t, err)
+			if err == nil {
+				t.Fail()
```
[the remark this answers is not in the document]
sgnn7: This is to fail fast without running asserts below on first error but a `continue` should do the same trick so I'll change it to that.

internal/profile/profile.go
```diff
@@ -2,64 +2,87 @@ package profile
...
 		if curType == profileType {
 			return true
 		}
 	}
 	return false
 }
 
-// ValidateType returns an error unless its argument is a valid profile type.
-func ValidateType(profileType string) error {
+func validateType(profileType string) error {
```
jonahx: This shoudn't be working, as we use this function in entrypoint.

sgnn7: The change here makes that validation part of instantiating the `profile` service: https://github.com/cyberark/secretless-broker/pull/981/files#diff-06fef48cb3a3ba8ce687ec606fb1f771R82. There's no need for entrypoint to know what are the exact valid types since this class can do it on it's own and return an error otherwise.

PR #982, opened 2019-11-05
983 Extract config validation from ProxyServices and add unit tests

jonahx:
History is clean.

Description taken from the primary commit:

### Extract config validation from ProxyServices and add unit tests

ProxyServices had a method "validateRequestedPluginsExist".  This logic
in this method is really about config validation, and doesn't belong in
ProxyServices.  This commit fixes this problem.

It comprises the following changes:

1. Remove "validateRequestedPluginsExist()"
2. Create a new function object "ValidateConfig" to house this logic.
   Define it in the "config.v2" package, where it belongs.
3. Invoke the logic at the border, in entrypoint, right when receive a
   new config.  This lets us fail fast.
4. Create unit tests for this object.  We had none before. 
5. Add a new method "plugin.AvailIDs" in the "plugin" package so that
   information such as the static built in connector names "ssh" and
   "ssh-agent" are centralized, and so that the validation code doesn't
   need not concern itself with those details.


internal/proxyservice/proxy_service.go
```diff
@@ -391,17 +342,10 @@ func (s *proxyServices) credsRetriever(
...
 	}
 }
 
-// getSecrets returns the secret values for the requested credentials.
-// TODO: Move this up one level, pass it down as dep.  Danger: This has
-//   a hardcoded dependency on plugin and v1.
-func (s *proxyServices) getSecrets(secrets []*v2.Credential) (map[string][]byte, error) {
-	return s.resolver.Resolve(secrets)
-}
-
```
jonahx: @sgnn7 Any idea how this change go in here?  This change was made a while ago iirc and I didn't do this (I don't think) as part of this PR.

pkg/secretless/plugin/plugins.go
```diff
@@ -11,3 +11,24 @@ type AvailablePlugins interface {
 	HTTPPlugins() map[string]http.Plugin
 	TCPPlugins() map[string]tcp.Plugin
 }
+
+// AvailIDs returns a list of all connector names based on the
+// AvailablePlugins given as well as the built in connectors.  It's a pure
+// function that depends only on the AvailablePlugins interface, which is
+// why we define it here rather than the implementation packaged "sharedobj".
+func AvailIDs(availPlugins AvailablePlugins) []string {
```
sgnn7: There's no advantage to shortening the name here

pkg/secretless/plugin/plugins.go
```diff
@@ -11,3 +11,24 @@ type AvailablePlugins interface {
...
+
+	// Add the names of the static, built-in ssh plugins whose "connectors"
+	// don't even exist as separate objects, but are hardcoded into the ssh
+	// proxy services.
+	connectorNames := []string{"ssh", "ssh-agent"}
```
sgnn7: This seems like something that should be a module constant

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -83,10 +85,18 @@ func StartSecretless(params *SecretlessOptions) {
 	var allServices internal.Service
 
 	// Main service reload callback
+
+
```
sgnn7: Stray newline

pkg/secretless/entrypoint/entrypoint.go
```diff
@@ -60,6 +60,8 @@ func StartSecretless(params *SecretlessOptions) {
 		params.PluginChecksumsFile,
 		logger,
 	)
+	// Create the config validator
+	validateConfig := v2.NewValidateConfig(logger, availPlugins)
```
sgnn7: why is the config validator so far from the point of use?

sgnn7: I know you're not a fan of `-er` suffix but the current constructor name is really awkward - it's a double verb and it's almost implying that you are running the validation at this point rather than later.

pkg/secretless/config/v2/service.go
```diff
@@ -186,6 +188,60 @@ func connectorID(
...
+func NewValidateConfig(lg logapi.Logger, ap plugin.AvailablePlugins) ValidateConfig {
+	return ValidateConfig{
+		logger: lg,
+		availPlugins: ap,
+	}
+}
+
+// Call validates that the connectors of the given Config exist, according to
+// the AvailablePlugins.
+func (v *ValidateConfig) Call(cfg Config) error {
```
sgnn7: Not a blocker but we don't really use the `Call` pattern almost anywhere else and it doesn't describe what the method is doing to the params.

pkg/secretless/config/v2/validate_config_test.go
```diff
@@ -0,0 +1,67 @@
...
+				{Name: "Fake Service 1", Connector: "FAKE PLUGIN 1"},
+				{Name: "Fake Service 2", Connector: "FAKE PLUGIN 2"},
+			},
+		}
+
+		err := validateConfig.Call(cfg)
+		assert.NotNil(t, err)
+
+		// Ensure error message mentions which plugins are missing.
+		assert.Regexp(t, "FAKE PLUGIN 1", err)
```
sgnn7: Would these panic up if `err == nil` since the assert above won't bail if it's nil?

pkg/secretless/config/v2/service.go
```diff
@@ -186,6 +188,60 @@ func connectorID(
...
+// plugins available to create connectors.
+type ValidateConfig struct {
+	logger logapi.Logger
+	availPlugins plugin.AvailablePlugins
+}
+
+// NewValidateConfig creates a new instance of ValidateConfig.
+func NewValidateConfig(lg logapi.Logger, ap plugin.AvailablePlugins) ValidateConfig {
+	return ValidateConfig{
+		logger: lg,
```
sgnn7: Inconsistent variable naming - `l` should be fine here if you're using golang naming standard or you can use `logger` to match some of our other code.

pkg/secretless/entrypoint/entrypoint.go
jonahx: > why is the config validator so far from the point of use?

Thinking was to keep with the creation of the other dependencies.  Since it's essentially a derived dependency.  But I also considered putting it closer the point of use and since it was confusing to you I'll move it back there.

PR #987, opened 2019-11-08
Move "ensureSocketIsDeleted" logic into "ConfigEnv" and unit test it.

jonahx:
History is clean-ish but could be better.

- Moves `ensureSocketIsDeleted()` out of `ProxyServices` and into `ConfigEnv`
- Add unit tests for the logic
- Move all the logic for `ConfigEnv` into its own file
- Sepatately: Changes `Service` to use a proper `NetworkAddress` type for `ListenOn`
- (related to previous point) Updates existing unit tests involving `ListenOn` as needed
- Improves formatting on some existing unit tests with long lines (since I was in the file anyway)

### DIFF
--- a/pkg/secretless/config/v2/config_env.go
+++ b/pkg/secretless/config/v2/config_env.go
@@ -1,0 +1,148 @@
+package v2
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/log"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin"
+	"github.com/go-ozzo/ozzo-validation"
+)
+
+// DeleteFileFunc is a function that takes a filename, attempts to delete the
+// file, and returns an error if it can't.
+type DeleteFileFunc func(name string) error
+
+// FileInfoFunc is a function that takes a filename and returns information
+// about that file, or an error if it cannot be found or read.
+type FileInfoFunc func(name string) (os.FileInfo, error)
+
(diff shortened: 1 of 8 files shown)


pkg/secretless/config/v2/config_env.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// ConfigEnv represents the runtime environment that will fulfill the services
+// requested by the Config.  It has a single public method, Prepare, that
+// ensures the runtime environment supports the requested services.
+type ConfigEnv struct {
+	logger       log.Logger
+	availPlugins plugin.AvailablePlugins
+	getFileInfo  func(name string) (os.FileInfo, error)
+	deleteFile   func(name string) error
```
sgnn7: Field names should probably be sorted

sgnn7: Type-aliasing might increase readability for these two func types if it's feasible

pkg/secretless/config/v2/service.go
```diff
@@ -188,66 +186,22 @@ func connectorID(
...
+//func Stat(name string) (FileInfo, error) {
+//func Remove(name string) error {
+//type FileInfo interface {
+//	Name() string       // base name of the file
+//	Size() int64        // length in bytes for regular files; system-dependent for others
+//	Mode() FileMode     // file mode bits
+//	ModTime() time.Time // modification time
+//	IsDir() bool        // abbreviation for Mode().IsDir()
+//	Sys() interface{}   // underlying data source (can return nil)
+//}
```
sgnn7: Leftover stuff?

PR #989, opened 2019-11-08
Add support of `sslmode` parameter to the mysql plugin

sgnn7:
Old code had no way to pass this parameter into the connector so now we
support it in that capacity. This inclusion of the flag still does nto
mean that we are actively using it though.

Connected to #928 

### DIFF
--- a/internal/plugin/connectors/tcp/mysql/connection_details_test.go
+++ b/internal/plugin/connectors/tcp/mysql/connection_details_test.go
@@ -12,14 +12,17 @@ func TestExpectedFields(t *testing.T) {
 		"port":     []byte("1234"),
 		"username": []byte("myusername"),
 		"password": []byte("mypassword"),
+		"sslmode":  []byte("disable"),
 	}
 
 	expectedConnDetails := ConnectionDetails{
-		Host:     "myhost",
+		Host: "myhost",
+		Options: map[string]string{
+			"sslmode": "disable",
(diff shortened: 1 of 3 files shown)


internal/plugin/connectors/tcp/mysql/connection_details_test.go
```diff
@@ -12,14 +12,17 @@ func TestExpectedFields(t *testing.T) {
...
 		"username": []byte("myusername"),
 		"password": []byte("mypassword"),
+		"sslmode":  []byte("disable"),
 	}
 
 	expectedConnDetails := ConnectionDetails{
-		Host:     "myhost",
+		Host: "myhost",
+		Options: map[string]string{
+			"sslmode": "disable",
```
izgeri: is that actually the intended flow for this test? the default SSL mode is "require" (see [here](https://github.com/cyberark/secretless-broker/blob/2dd84c3d152ff878ed31b5b552dbc8c5d12ee2d8/internal/plugin/connectors/tcp/ssl/ssl.go#L34))

izgeri: leaving it blank would effectively be the same as "require"

sgnn7: This tests explicitly sets it to "disable" about 5 lines up from here

PR #1002, opened 2019-11-18
999 generic http connector

jonahx:
TBD but....

Quick explanation:

- Creates a generic http connector
- Integration test shows how to use it to recreated Basic Auth connector
- Includes unit tests, though they aren't exhaustive.

### DIFF
--- a/internal/plugin/connectors/http/generic/config_test.go
+++ b/internal/plugin/connectors/http/generic/config_test.go
@@ -1,0 +1,50 @@
+package generic
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func sampleConfig() []byte {
+	return []byte(`
+  credentialValidation:
+    username: '[^:]+'
+  headers:
+    Authorization: 'Basic {{ printf "%s:%s" .username .password | base64 }}'
+    Name-With-Dashes: '{{ .username }}'
+    SimpleConcatenation: '{{ .username }} - {{ .password }}'
+  forceSSL: true
+  authenticateURLsMatching:
+    - ^http
+`)
+}
+
+func Test_newConfig(t *testing.T) {
+	t.Run("creates expected headers", func(t *testing.T) {
+		cfg, err := newConfig(sampleConfig())
+
+		assert.NoError(t, err)
+		if err != nil {
+			return
+		}
+
+		headers, err := cfg.renderedHeaders(map[string][]byte{
+			"username": []byte("Jonah"),
+			"password": []byte("secret"),
+		})
+
+		assert.NoError(t, err)
+		if err != nil {
+			return
+		}
+
+		// A couple simple test cases
+		assert.Equal(t, "Jonah", headers["Name-With-Dashes"])
+		assert.Equal(t, "Jonah - secret", headers["SimpleConcatenation"])
+
+		// Assert against value calculated with independent base64 encoder.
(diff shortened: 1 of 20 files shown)


internal/plugin/connectors/http/generic/connector.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	gohttp "net/http"
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/log"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+)
+
+// Connector injects an HTTP request with AWS authorization headers.
```
BradleyBoutcher: Should this say AWS?

http-generic.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
```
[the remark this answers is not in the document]
jonahx: Good catch.  That's an error.  Should have been deleted.

internal/plugin/connectors/http/generic/config_test.go
```diff
@@ -0,0 +1,50 @@
+package generic
```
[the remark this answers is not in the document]
jonahx: No, in go white box unit test files (which is what this is), and often even black box unit tests as well, go in the same directory as the files they're testing.  Here's a random example from the stdlib:

https://golang.org/src/bytes/

pkg/secretless/plugin/sharedobj/internal_plugins.go
```diff
@@ -20,9 +21,10 @@ func GetInternalPluginsFunc() (plugin.AvailablePlugins, error) {
 	// New connectors should have an entry in the map below, according to their type (HTTP/TCP)
 	return &Plugins{
 		HTTPPluginsByID: map[string]http.Plugin{
-			"aws":        aws.GetHTTPPlugin(),
-			"basic_auth": basicauth.GetHTTPPlugin(),
-			"conjur":     conjur.GetHTTPPlugin(),
+			"aws":          aws.GetHTTPPlugin(),
+			"basic_auth":   basicauth.GetHTTPPlugin(),
+			"generic_http": generic.GetHTTPPlugin(),
+			"conjur":       conjur.GetHTTPPlugin(),
```
doodlesbykumbi: nit: sort keys

test/http_generic/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+      interval: 1s
+      timeout: 30s
+    volumes:
+      - ./http-basic-auth.conf:/etc/nginx/conf.d/default.conf:ro
+      - ./basic_auth.htpasswd:/etc/nginx/.htpasswd:ro
+
+  secretless:
+    build:
+      context: ../..
+    environment:
```
doodlesbykumbi: These envvars aren't use anywhere as far as I can tell. If not let's get rid of them

jonahx: I'll try it and see we'll see if anything breaks :)

test/mssql_connector/start
```diff
@@ -13,7 +13,7 @@ done
 # the order of the services is important. mssql must be up before we start secretless
 docker-compose up -d mssql
 
-./wait_for_mssql
+time ./wait_for_mssql
```
doodlesbykumbi: hm ?

PR #1008, opened 2019-11-20
1007 rewrite basic auth with generichttp

jonahx:
History is clean.

Commit message should provide context.


internal/plugin/connectors/http/generic/external_api.go
```diff
@@ -0,0 +1,75 @@
...
+// NewNewConnector is a meta-constructor: a function that returns a NewConnector
+// function. If that's confusing, here's an equivalent phrasing: it's a function
+// that creates a connector constructor.  It's intended to be used by other HTTP
+// connectors who delegate their implementation to the generic HTTP connector.
+//
+// It takes a static ConfigYAML -- static in the sense that its represents the
+// connector definition, not user configuration -- which is then merged at
+// runtime with the actual config from the .yml file, allowing ForceSSL to be
+// configured by the user.
+func NewNewConnector(staticCfgYAML *ConfigYAML) (http.NewConnectorFunc, error) {
```
BradleyBoutcher: Nit: Could we call this NewConnectorConstructor or something? I think NewNew might be a little confusing. It describes what it returns moreso than what it does, I feel.

pkg/secretless/plugin/connector/http/http.go
```diff
@@ -38,24 +38,24 @@ type Connector interface {
 }
 
 /*
-ConnectorConstructor makes it easy to define Secretless HTTP connector plugins.
+NewConnectorFunc makes it easy to define Secretless HTTP connector plugins.
```
BradleyBoutcher: good explanation 👍

pkg/secretless/plugin/sharedobj/internal_plugins.go
```diff
@@ -24,7 +24,7 @@ func GetInternalPluginsFunc() (plugin.AvailablePlugins, error) {
 			"aws":          aws.GetHTTPPlugin(),
 			"basic_auth":   basicauth.GetHTTPPlugin(),
 			"conjur":       conjur.GetHTTPPlugin(),
-			"generic_http": generic.GetHTTPPlugin(),
+			"generic_http": generichttp.GetHTTPPlugin(),
```
BradleyBoutcher: Nit: Should we change the other naming conventions to include the http suffix as well? Or change this back to just generic? It seems weird to have the 'http' suffix for just this one. I guess it's possible we add generic plugins elsewhere, but that being the case we should probably change the package names for anything in the http package.

jonahx: It's a good sign that your comments are all about decisions I was on the fence about.

So here's the argument for keeping it.  The correctness of a package name can only be determined by looking at how it's used by other packages.  In this, from above:
```go
newConnector, err := generichttp.NewNewConnector(
```
Let's make two comparisons.  First, with how I had before:
```go
newConnector, err := generic.NewNewConnector(
```
Hmmm... in this case, what is "generic".  I think a reader of the code would need to jump to definition or look at the import path to figure it out.

Now let's see how a reference to basic auth looks:
```go
info := basicauth.PluginInfo()
```
The ambiguity is not the same as in the generic case.  A typical dev may be expected to understand that `basicauth` is a term relating to HTTP Basic Authentication.  We may not expect an analogous understanding when reading `generic`.  Hence the additional `http` is warranted in one case and not the other.  

A counter-counterpoint might be: Anything ever using the generic http connector will already be an an HTTP context....

Wdyt?

internal/plugin/connectors/http/basicauth/plugin.go
```diff
@@ -16,16 +18,22 @@ func PluginInfo() map[string]string {
...
-	return &Connector{
-		logger:   conRes.Logger(),
-	}
-}
-
 // GetHTTPPlugin is required as part of the Secretless plugin spec for HTTP
 // connector plugins. It returns the HTTP plugin.
 func GetHTTPPlugin() http.Plugin {
-	return http.ConnectorConstructor(NewConnector)
+	newConnector, err := generichttp.NewNewConnector(
```
sgnn7: Already in slack but the `NewNew` is pretty hard to read

internal/plugin/connectors/http/basicauth/plugin.go
```diff
@@ -16,16 +18,22 @@ func PluginInfo() map[string]string {
...
+				"username": "[^:]+",
+			},
+			Headers: map[string]string{
+				"Authorization": " Basic {{ printf \"%s:%s\" .username .password | base64 }}",
+			},
+		},
+	)
+	// This should never occur
+	if err != nil {
+		log.Panicf("Failed to create generic HTTP NewConnector: %s", err)
```
sgnn7: Hmm... do we not have our logger passed into these? If not, should we?

internal/plugin/connectors/http/basicauth/plugin.go
```diff
@@ -16,16 +18,22 @@ func PluginInfo() map[string]string {
...
 // connector plugins. It returns the HTTP plugin.
 func GetHTTPPlugin() http.Plugin {
-	return http.ConnectorConstructor(NewConnector)
+	newConnector, err := generichttp.NewNewConnector(
+		&generichttp.ConfigYAML{
+			CredentialPatterns: map[string]string{
+				"username": "[^:]+",
+			},
+			Headers: map[string]string{
+				"Authorization": " Basic {{ printf \"%s:%s\" .username .password | base64 }}",
```
sgnn7: The prefix space in `" Basic ...` is either an error or we should figure out why it's needed - seems odd that it would be needed.

internal/plugin/connectors/http/generic/config.go
```diff
@@ -71,17 +64,10 @@ func (c *config) renderedHeaders(
 	return headers, nil
 }
 
-// newConfig takes the raw cfgBytes, unmarshals them into a temporary configYAML
-// struct, and then validates and converts that into a config -- which is what
-// our application wants to work with.
-func newConfig(cfgBytes []byte) (*config, error) {
+// newConfig takes a ConfigYAML, validates it, and converts it into a config --
```
sgnn7: `converts it into a config` - should indicate what type of config

internal/plugin/connectors/http/generic/external_api.go
```diff
@@ -0,0 +1,75 @@
+package generichttp
+
+import (
+	"fmt"
+
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/http"
+	"gopkg.in/yaml.v2"
```
sgnn7: Needs sorting

internal/plugin/connectors/http/generic/external_api.go
```diff
@@ -0,0 +1,75 @@
...
+// ConfigYAML is an almost-literal representation of the YAML config section
+// of a generic HTTP connector.  It has two purposes:
+//
+//     1.  It allows other internal connectors to configure a generic connector
+//         to which they are delegating their implementation.  That is, it's
+//         a code-level interface equivalent of a .yml file config section.
+//     2.  Technically, it acts as an intermediate form when parsing an actual
+//         .yml file.  That is, we convert from .yml --> ConfigYAML --> config.
+type ConfigYAML struct {
+	CredentialPatterns map[string]string `yaml:"credentialValidation"`
```
sgnn7: Wouldn't this be `CredentialValidations`?

k8s-ci/k8s_crds/test
```diff
@@ -51,7 +51,7 @@ function wait_for_CRD() {
   echo "waiting for CRD to be ready"
 
   local attempt_count=0
-  local max_attempts=100
+  local max_attempts=200
```
sgnn7: This should be nixed per slack thread I think

pkg/secretless/plugin/sharedobj/internal_plugins.go
sgnn7: I'm probably with @BradleyBoutcher that it doesn't seem to fit in nicely with the rest. I think the usage of that package can be fine with renaming in case of ambiguity which almost seems expected in a lot of other codebases that have this nested layout (eg. `import ( generichttp ".../http/generic" )`):
- https://github.com/kubernetes/sample-controller/blob/master/main.go#L30-L31
- https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/cmd.go

internal/plugin/connectors/http/basicauth/plugin.go
sgnn7: Total newb questions but is this evaluated here or is something else evaluating that dynamic string somewhere?

jonahx: It's a go text template and template itself is created in the `NewNewConnector` ctor.  That template remains in memory and is rendered with the credentials at runtime.

internal/plugin/connectors/http/basicauth/plugin.go
jonahx: Ah, notice where it is.  We're inside of `GetHTTPPlugin`, which  returns a _plugin_, which itself provides the `NewConnector` ctor, which is what would be passed the logger.  So this exists "above" anything that would a  logger.  This is framework code.  

To say it another way, you should think of the "generic http" connector as a sort of meta-programming tool (though not in the reflection sense). It allows us to "generate" other plugins "declaratively."  I use scare quotes because it's not code generation as pre-compile step, but conceptually you can think of it code generation.  And it's not declarative in the sense of "a yaml" file, but it is declarative in the sense of "you just define a struct that maps 1-1 to a yaml file."

internal/plugin/connectors/http/generic/config.go
jonahx: Ha!  Scroll up 10 lines above this comment.  See line 14.

pkg/secretless/plugin/sharedobj/internal_plugins.go
jonahx: alright.  `generic` it is!

internal/plugin/connectors/http/basicauth/plugin.go
sgnn7: k - I think we still need to remove the prefixed whitespace though, right?

internal/plugin/connectors/http/basicauth/plugin.go
sgnn7: 👍 though just a small comment that `It could never occur, unless the connector definition itself was broken` is what the tests should catch and not break the harness. If that code will break the test run, we should make sure that it doesn't.

internal/plugin/connectors/http/basicauth/plugin.go
jonahx: Yeah I'll try removing it.  It should be fine I think.  I only put it there because its generally there if you look at the raw headers.

internal/plugin/connectors/http/generic/external_api.go
BradleyBoutcher: Do we have any sort of check style for sorting imports, or a means of adding one?

PR #1022, opened 2019-11-25
Add REAMDE for Generic HTTP connector

jonahx:
This commit adds detailed docs for both end-users and developers.

### DIFF
--- a/internal/plugin/connectors/http/generic/README.md
+++ b/internal/plugin/connectors/http/generic/README.md
@@ -1,0 +1,200 @@
+# Generic HTTP Authentication
+
+Secretless comes with built-in HTTP connectors that let you to connect to
+services like AWS, or any service that uses Basic authentication.  But what if
+you want to connect to a service that Secretless has no connector for?
+
+The Generic HTTP connector can help you.  
+
+It can inject any credential you want directly into any HTTP header.  You
+define the header name and the credential you want injected.  You can even
+concatenate credentials together, or encode them in base64.
+
+## End User Documentation
+
+Using the Generic HTTP connector is just a matter of configuring your
+`secretless.yml`.
+
+It's easiest to learn how this works by example.
+
+### Example Configurations
+
+We'll start with two example configurations, and then explain each of the
+relevant details.
+
+#### Header with single credential
+
+Here's an simple example configuration for a fictional "Example Service" that
+requires an single API key credential in a header called `X-ApiKey`.
+
+```yaml
+services:
+  my_example_service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:8080
+    credentials:
+      apikey: 
+        from: conjur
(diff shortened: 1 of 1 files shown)


internal/plugin/connectors/http/generic/README.md
```diff
@@ -0,0 +1,201 @@
...
+It can inject any credential you want directly into any HTTP header.  You
+define the header name and the credential you want injected.  You can even
+concatenate credentials together, or encode them in base64.
+
+## End User Documentation
+
+Using the Generic HTTP connector is just a matter of configuring your
+`secretless.yml`.
+
+It's easiest to learn how this works by example.
```
BradleyBoutcher: I like the conversational tone, it makes for easy reading.

internal/plugin/connectors/http/generic/README.md
```diff
@@ -0,0 +1,201 @@
...
+      match:
+        - ^http                   # apply this connector to all requests
+```
+
+#### Reimplementing Basic Authentication
+
+Of course, if you needed Basic Authentication in a real application, you'd use
+the built-in `basic_auth` connector.  But the Generic HTTP connector is
+powerful enough to provide the same functionality, and it's instructive to see
+how it _could_ be done:
```
BradleyBoutcher: Good distinction

internal/plugin/connectors/http/generic/README.md
```diff
@@ -0,0 +1,201 @@
...
+described above, a few other pieces of boilerplate are required to complete
+the process.
+
+Refer to the Basic auth connector
+(`internal/plugin/connectors/http/basicauth/plugin.go`) as an example when
+working through the steps below.
+
+Here are the steps, in detail:
+
+1. Create a new directory for your connector under `/internal/plugin/connectors/http`.
```
BradleyBoutcher: Not a blocker: I feel like we've cut away so much code already with the success of the generic connector, perhaps we can reduce this to a single connector directory containing all the various plugin files, with appropriate naming conventions for each file. 

There's a part of me that wonders if there's value in reducing it further, since the reduced plugin files are so similar now. However, it might reduce readability.

jonahx: Interesting.  So the idea would be to have a directory like `/internal/plugin/connectors/http/generic_implementations` and then files `basic_auth.go` and `conjur.go`, etc in there?

I sort of like the idea but I'm not sure the incremental benefit makes it worth doing.  It also does have a couple downsides: eg, the test dirs would no longer mirror the impl dirs.  Also, perhaps more importantly, you're breaking the consistency of things, ie, making the generic connectors into a special case.  Whereas currently they're just like any other connector and they _happen_ to be implemented using some utilities provided by the `generic` classes.  That might be a preferable way to think of them.  It's a debatable point.

In any case, if we decide to do this, it will involve changes beyond this README so we should propose it as a followup.

Issue #1009, opened 2019-11-21
The HTTP Conjur Connector is rewritten to use the Generic HTTP Connector as a template


BradleyBoutcher:
We will be consuming the generic connector within the plugin, rather 
- [x] Change plugin import to use generic connector

This will be the more important part of the change. We decorate the generic HTTP headers here with Conjur specific parameters. 
- [x] Modify GetHTTPPlugin to use NewNewConnector

Add anything required by the Generic Connector
- [ ] Modify YAML as needed

The functionality here will be replaced by Generic, so this is unnecessary 
- [x] Remove connector.go
 
Run unit tests and integration tests to confirm we didn't break anything 
- [ ] Run tests to confirm preservation of functionality

PR #1024, opened 2019-11-26
1009 - The Conjur HTTP connector makes use of the generic connector 

BradleyBoutcher:
Several changes were made to the Conjur Connector to allow use of the generic connector.

Since @jonahx created the generic connector, allowing the injection of 
headers to create custom connectors for new services, we are 
attempting to reduce our code base and streamline our existing 
connectors. To that end, several changes were made to the
Conjur connector.

- The connector.go file was removed from the conjur directory,
with the NewConnectorConstructor taking the place of its functionality.

- The GetHTTPPlugin method in the Conjur Connector now defines the headers
used for authorization, such as the token.

- Relocaed the forceSSL property, per generic connector standards.

#1009 

secretless-broker/internal/plugin/connectors/http/conjur/plugin.go

- [X] Manually tested [Keychain provider](https://github.com/cyberark/secretless-broker/tree/master/test/manual/keychain_provider)

- [x] An updated README with instructions on how to manually test this feature
- [x] Utility `start` and `stop` scripts to spin up and tear down the test environments
- [x] A `test` script to run some basic manual tests (optional; if does not exist, the README should have detailed instructions)


internal/plugin/connectors/http/conjur/plugin.go
```diff
@@ -16,16 +18,26 @@ func PluginInfo() map[string]string {
...
-}
-
 // GetHTTPPlugin is required as part of the Secretless plugin spec for HTTP
 // connector plugins. It returns the HTTP plugin.
 func GetHTTPPlugin() http.Plugin {
-	return http.NewConnectorFunc(NewConnector)
+	newConnector, err := generichttp.NewConnectorConstructor(
+		&generichttp.ConfigYAML{
+			CredentialValidations: map[string]string{
+				// Unsure
```
izgeri: are you looking for the standard format of the Conjur token? that's probably worth asking about in Slack and adding here before this gets merged

BradleyBoutcher: Good catch - I checked with Jonah yesterday and forgot to remove it.

izgeri: if it helps it looks like it's just a JWT: https://docs.conjur.org/Latest/en/Content/Get%20Started/cryptography.html?Highlight=cryptography#Authenticationtokens

izgeri: whoops, missed your comment. 👍

test/connector/http/conjur/secretless.yml
```diff
@@ -11,3 +11,4 @@ services:
     config:
       authenticateURLsMatching:
         - ^http\:\/\/conjur\/
+      forceSSl: false
```
doodlesbykumbi: Missing newline at EOF

internal/plugin/connectors/http/conjur/plugin.go
```diff
@@ -16,16 +18,23 @@ func PluginInfo() map[string]string {
...
-}
-
 // GetHTTPPlugin is required as part of the Secretless plugin spec for HTTP
 // connector plugins. It returns the HTTP plugin.
 func GetHTTPPlugin() http.Plugin {
-	return http.NewConnectorFunc(NewConnector)
+	newConnector, err := generichttp.NewConnectorConstructor(
+		&generichttp.ConfigYAML{
+			Headers: map[string]string{
+				"Authorization": "Token token={{ `%s` .token | base64 }}",
```
doodlesbykumbi: Since the token value needs to be surrounded by quotation marks I'd suggest the following:
```
"Authorization": `Token token="{{ .token | base64 }}"`,
```

test/connector/http/conjur/secretless.yml
```diff
@@ -11,3 +11,5 @@ services:
...
```
jonahx: typo: `forceSSL`

internal/plugin/connectors/http/conjur/plugin.go
```diff
@@ -1,7 +1,9 @@
 package conjur
 
 import (
-	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+	"log"
+
+	generichttp "github.com/cyberark/secretless-broker/internal/plugin/connectors/http/generic"
```
jonahx: I had originally named the package `generichttp` but after feedback from Srdjan and (I thought) you  we decided to go with the name `generic` since we thought the "http" would be clear from context.

So I think either:
1. We revisit that decision and change the package.
2. Keep the package name and don't do the rename here.

At this point I'd lean toward 2.

Generally speaking, you should only be renaming a package because it has a conflict with another package.  Otherwise it's telling you that the package name is not good enough, and you should fix the root problem (assuming it's code we own).

test/connector/http/conjur/secretless.yml
```diff
@@ -11,3 +11,5 @@ services:
     config:
       authenticateURLsMatching:
         - ^http\:\/\/conjur\/
+      forceSSL: false
+
```
sgnn7: Drive-by comment: both ymls have a stray newline at the end of the files

Issue #1025, opened 2019-11-26
White box UTs for generic HTTP connector's credentials validator exist


diverdane:
## Description
The generic HTTP connector's validate function requires some basic UT test cases.

AC:
- [ ] Unit test includes successful validation with multiple credentials
- [ ] Unit test includes test for detection of missing required credential
- [ ] Unit test includes test for detection of credential that does not match a configured validation pattern

PR #1026, opened 2019-11-26
Adds UT test cases for generic connector validator

diverdane:
This change adds UT test cases for the generic HTTP connector's
validate method. This includes:
- Creating a common getSampleConfig() testing function for parsing
  the UT sampleConfig
- Add test cases for:
  * Valid username and password.
  * Missing required credential.
  * Credential that doesn't match corresponding validation string.

Fixes: https://github.com/cyberark/secretless-broker/issues/1025

Issue #1025 

internal/plugin/connectors/http/generic/config_test.go:Test_validate()

- [ Y ] Manually tested [Keychain provider](https://github.com/cyberark/secretless-broker/tree/master/test/manual/keychain_provider)
- [ N ] Manually run [Kubernetes-Conjur demo](https://github.com/conjurdemos/kubernetes-conjur-demo) with a local Secretless Broker image build of your branch

### DIFF
--- a/internal/plugin/connectors/http/generic/config_test.go
+++ b/internal/plugin/connectors/http/generic/config_test.go
@@ -1,15 +1,17 @@
 package generic
 
 import (
+	"strings"
 	"testing"
 
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 )
 
-func sampleConfig() []byte {
+func sampleConfigYAML() []byte {
 	return []byte(`
-  credentialValidation:
-    username: '[^:]+'
+  credentialValidations:
+    username: '^[^:]+$'
   headers:
     Authorization: 'Basic {{ printf "%s:%s" .username .password | base64 }}'
     Name-With-Dashes: '{{ .username }}'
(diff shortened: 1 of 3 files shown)


internal/plugin/connectors/http/generic/config_test.go
```diff
@@ -20,16 +21,21 @@ func sampleConfig() []byte {
...
+	if err != nil {
+		assert.Fail(t, "sampleConfig should never fail")
+		return nil, err
+	}
+	cfg, err := newConfig(cfgYAML)
+
+	assert.NoError(t, err)
+	return cfg, err
+}
+
```
jonahx: This is a relatively small instance of it, ofc, but i think there are a couple important conceptual issues here worth addressing.

1. We're mixing concerns.  Currently, this is both a utility for constructing a `config` as well as something that runs some tests.  It wants to be only the former.  We want all our tests to be directly inside `TestXXX` methods.  Fixing this will also let us to remove the paramter `t`.
2. Before this change, we had `sampleConfig`, which was a literal yaml string.  That name was ok because there was nothing to confuse it with.  Since we now have two separate helper functions for closely related but different things -- the literal yaml config, and the `config` _type_ -- we want to give them unambiguous names.  I'd suggest `sampleConfigYAML` and `sampleConfig`.
3. A broader principle related to point 2 -- we never use `getXXX` for value types.   Here's a [short video snippet on this best practice](https://youtu.be/MzTcsI6tn-0?t=1457).  Instead, the name should always be a noun that describes the thing being returned.

internal/plugin/connectors/http/generic/config_test.go
```diff
@@ -52,4 +58,54 @@ func Test_newConfig(t *testing.T) {
...
+			if tc.expErrStr == "" {
+				assert.NoError(t, err)
+			} else {
+				assert.Error(t, err)
+				if err != nil {
+					assert.True(t, strings.Contains(err.Error(), tc.expErrStr))
+				}
+			}
+		})
+	}
```
jonahx: We try to follow [line of sight programming](https://youtu.be/5DVV36uqQ4E?t=660) wherever possible.  This almost always makes the code cleaner and  the logic easier to follow code because the happy path and the "exceptions," in the colloquial sense of that word, are easily differentiated.  It also reduces indentation level. In this case the change is almost trivial:
```go
	for _, tc := range testCases {
		t.Run(tc.description, func(t *testing.T) {
			err = cfg.validate(tc.creds)
			if tc.expErrStr == "" {
				assert.NoError(t, err)
				return
			}
			assert.Error(t, err)
			if err != nil {
				assert.True(t, strings.Contains(err.Error(), tc.expErrStr))
			}
		})
	}
```

internal/plugin/connectors/http/generic/config_test.go
diverdane: Great suggestions! Thanks for the explanations and pointers.

internal/plugin/connectors/http/generic/config_test.go
diverdane: Fixed.

Issue #1016, opened 2019-11-22
The MSSQL connector faithfully passes on the Login response


izgeri:
At present the connector sends a hard-coded OK response message to client.
(internal/plugin/connectors/tcp/mssql/authentication.go:62)

In general, the OK response contains information about the target service. It is, therefore, important for it to be dynamic and be derived from the actual login response from the target service.

AC:
- [ ] The OK message for the MSSQL connector is constructed in a way similar to the PG connector
- [ ] Unit tests exist to validate that the OK message is constructed appropriately

Issue #1014, opened 2019-11-22
MSSQL Connector faithfully propagates PreLoginResponse between client and target service


izgeri:
AC:
- [x] The MSSQL PRELOGIN response is passed between the client and the target service via Secretless and is not hardcoded
- [ ] Tests exist to validate the transmission of the PRELOGIN response

Issue #1013, opened 2019-11-22
MSSQL Connector faithfully propagates errors between client and target service


izgeri:
AC:

- [ ] MSSQL connector is updated to faithfully propagate errors between client and target service
- [ ] Unit tests exist to validate that MSSQL transparently propagates errors

Issue #968, opened 2019-10-25
Secretless MSSQL Connector is Beta Quality


izgeri:
Based on the [draft quality guidelines](), to get the MSSQL connector to beta quality we will need to:
- Be clear about the known limitations of the MSSQL connector (cyberark/secretless-docs#206)
- Be clear about the supported MSSQL versions (cyberark/secretless-docs#206)
- A documentation page exists for the MSSQL connector (cyberark/secretless-docs#206)
- MSSQL connector has sufficient automated tests
  - MSSQL connector has integration tests that run in a docker-compose network and validate (at a minimum) that: 
    - A client can connect to MSSQL via Secretless and query a database 
  - MSSQL connector has unit tests that validate (at a minimum):  (#1010)
    - The most relevant messages sent between client and server are transmitted with high fidelity, eg 
      - Client options and arguments  
      - Errors  
      - Success messages 
  - All main positive tests exist. Many negative tests & other focuses. Might be some edge cases remaining to test.  
- All acceptance criteria are met (see below)
- MSSQL connector has initial security review (conjurinc/aam-rd-leads#31)
- Project with MSSQL has undergone security scan with [gosec](https://github.com/securego/gosec) with no high or critical issues (#976)
- MSSQL has undergone initial XA (#977)
- MSSQL connector no longer has hardcoded server handshake responses (see #975)
- MSSQL connector has troubleshooting documentation (cyberark/secretless-docs#210)

We are deferring until the GA stage:
- Initial performance metrics on this connector

## Acceptance Criteria for Beta Release of MSSQL Connector

- Users can use Secretless with the MSSQL connector to deploy applications to Kubernetes or OpenShift that require an MSSQL Server 2017 v14.x database that the apps authenticate to using using SQL Server authentication 
- Users can create applications that connect to MSSQL instance with no credential knowledge 
- New connections to MSSQL always use fresh credentials from the configured credential provider(s) 
- MSSQL connector config is consistent with config for other Secretless database connectors (eg `username`, `password`, `host`, and (optional) `port`), and the standard default for `port` (eg 1433) is enforced 
- MSSQL Connector meets the [Core Acceptance Criteria](https://github.com/conjurinc/docs/blob/master/reference/core_acceptance_criteria.md)
- Secretless MSSQL connector has had a security review, and has no high or critical bugs 
  - Secretless MSSQL connector has undergone STRIDE threat modeling 
[...]

PR #1034, opened 2019-12-05
MsSQL Connector no longer sends hardcoded prelogin responses

### LINKED ISSUE #975: MSSQL connector no longer has hardcoded server handshake responses
The "fake handshake response" needs to become real, and all magic values need to be replaced with values logically derived from the client request, or explained and properly named.  Kumbi's initial comment regarding this:
  > Replace hardcoded MSSQL server responses in fake handshake with a more truthful handshake response. Ie, intercept the bytes flowing between secretless and the target server and extract information like server version etc.

BradleyBoutcher:
Since we use the third party connector for much of this process, it makes sense to reduce the code base for our overhead as much as possible, and further reduce the complexity of our signatures and variable names

- Removed the authentication.go file, since we are not performing any real authentication there
- Renamed authentication.go to authentication_responses.go, since it currently contains pre made responses. This can be removed once we dynamically return responses from mssql (#1013, #1014, and #1016)
- Combined Connect and ConnectToBackend since their functionality, purpose, and signatures are the same. This can probably be further reduced or have better naming conventions.
- Added comments detailing the steps we take within the Connect phases performed by Secretless
- Moved performLoginHandshake() from authentication.go to connector.go, and renamed to customizePreloginRequest, since that is all it's actually doing and still has a purpose in that regard
- Added some comments specifying story work where applicable

Closes #975 and #1014 

internal/plugin/connectors/tcp/mssql/connector.go


internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,28 +116,84 @@ func (connector *SingleUseConnector) Connect(
 
 	connector.clientConn = clientConn
 
-	err := connector.performPreLoginHandshake()
+	// Read the prelogin request from the client, but we don't need to do anything
+	// we just need to make sure the client isn't blocked
+	clientBuffer := mssql.NewTdsBuffer(4096, connector.clientConn)
```
sgnn7: This constant (`4096`) should be pulled up and defined as a const

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,28 +116,84 @@ func (connector *SingleUseConnector) Connect(
...
+		driverConnSet <- struct{}{}
+
+		if err != nil {
+			connector.logger.Errorf("failed to connect to mssql server: %s", err)
+			connector.sendErrorToClient()
+		}
+	} ()
+
+	// Retrieve information from channels within the driver
+	preloginResponse := <- ch
```
sgnn7: This comment should indicate that it's a blocking request

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -4,26 +4,111 @@ import (
...
 //
 // It is the main method of the SingleUseConnector. It:
 //   1. Constructs connection details from the provided credentials map.
-//   2. Dials the backend using credentials.
-//   3. Runs through the connection phase steps to authenticate.
+//   2. Reads the client PreLogin request
+//   3. Adds channels to the context that can intercept data from the driver
+//   4. Initiates authentication and connection to MsSQL through the third-party driver
+//   6. Intercepts PreLogin response from the driver
+//	 5. Customizes the PreLogin response to meet Secretless standards
```
sgnn7: I think 5 and 6 are swapped

BradleyBoutcher: good catch, meant to change the numbers there

izgeri:
#968 is the epic, so I don't think this PR closes it - is there another card you meant to link to?

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -4,26 +4,111 @@ import (
...
 // It is the main method of the SingleUseConnector. It:
 //   1. Constructs connection details from the provided credentials map.
-//   2. Dials the backend using credentials.
-//   3. Runs through the connection phase steps to authenticate.
+//   2. Reads the client PreLogin request
+//   3. Adds channels to the context that can intercept data from the driver
+//   4. Initiates authentication and connection to MsSQL through the third-party driver
+//   5. Intercepts PreLogin response from the driver
+//	 6. Customizes the PreLogin response to meet Secretless standards
+//		and sends it to the user
```
jonahx: 👍

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
 
 	connector.clientConn = clientConn
 
-	err := connector.performPreLoginHandshake()
+	// Read the prelogin request from the client, but we don't need to do anything
+	// we just need to make sure the client isn't blocked
```
jonahx: I know this  was an existing comment, but it's a bit cryptic.  Let's change to:

```
Secretless _is_ the client with respect to the server, and there is nothing in the 
pre-login handshake that needs to be passed along.  Secretless simply reads
it from the client and throws it away, so that client can advance to the next 
stage of the process.  Otherwise the client would block forever waiting for its
pre-login handshake to be read.
```
cc/ @doodlesbykumbi Is there anything we should add to this for maximum
clarity?

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
...
 	// Verify the driverConn is an mssql driverConn object and get its underlying transport
 	mssqlConn := driverConn.(*mssql.Conn)
 	connector.backendConn = mssqlConn.NetConn()
 
-	return nil
+	return connector.backendConn, nil
+}
+
+
+func (connector *SingleUseConnector) sendErrorWithMessage(message string, err error) (net.Conn, error){
```
jonahx: Name improvement: `sendErrorWithMessage` -> `sendError` (or `sendClientError`)

General naming principle: Avoid repeating what is already implicit in the parameters.  That is, anyone using this method would have to know that it included a message, because it has a message param, so repeating that fact in the name is redundant.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
...
-	if _, err = clientConn.Write(connector.CreateAuthenticationOKMessage()); err != nil {
-		connector.logger.Errorf("Failed to send a successful authentication"+
-			" response to the client"+
-			": %s", err)
-		connector.sendErrorToClient()
-		return nil, err
-	}
+	// Create the context for our connection
+	parentContext := context.Background()
+	ch := make(chan map[uint8][]byte)
```
jonahx: `ch` was our temporary name, we should make it more descriptive before mergering.  How about `preLoginResponseCh`?

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
...
-}
+	// Set a 'marker' for when the driver has finished connecting
+	driverConnSet := make(chan struct {})
 
-// ConnectToBackend establishes the connection to the target database and sets
-// the backendConnection field.
-func (connector *SingleUseConnector) ConnectToBackend(dataSourceName string) error {
-	var err error
+	// Add channels to the context, to retrive information from the driver
+	childContext := context.WithValue(parentContext, PreLoginResponseKey, ch)
```
jonahx: Similarly, `preLoginContext` seems better here.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
...
-	// go-mssqldb driver package.  The driver package has its own concept of a
-	// "connector", and its connectors also have a "Connect" method.
-	driverConnector, err := mssql.NewConnector(dataSourceName)
+	// Build a new driver connection
+	var driverConn driver.Conn
+
+	go func() {
+		// Kick off authentication through our third party connector
+		driverConn, err = driverConnector.Connect(childContext)
+		driverConnSet <- struct{}{}
```
jonahx: I originally came up with `driverConnSet` but it's another instance of a temporary name.  Better to describe what it's telling us in terms of the high-level connection steps (rather than in terms of the implementation detail of a variable being set).  Hence I suggest:

`driverConnSet` -> `connectPhaseFinished`

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
...
+		if err != nil {
+			_, err = connector.sendErrorWithMessage("failed to connect to mssql server: %s", err)
+		}
+	} ()
+
+	// Blocks continuation until we've received the preLoginResponse from the driver
+	preloginResponse := <- ch
+
+	// We don't want the driver to use encryption when talking to the server
+	preloginResponse[mssql.PreloginENCRYPTION] = []byte{mssql.EncryptNotSup}
```
jonahx: Isn't this one saying that we don't want the client to use encryption when talking to Secretless?

cc/ @doodlesbykumbi

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
...
 	if err != nil {
-		connector.logger.Errorf("Failed to create a go-mssqldb connector: %s", err)
-		return err
+		return connector.sendErrorWithMessage("failed to write prelogin response: %s", err)
 	}
 
-	driverConn, err := driverConnector.Connect(context.Background())
+	// We actually don't need the client's handshake response;
+	// we just need for them to not be blocked
+	err = clientBuffer.ReadNextPacket()
```
jonahx: Similar to my comment above, I think we should elaborate in this comment.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +118,102 @@ func (connector *SingleUseConnector) Connect(
...
+	// TODO: 	Replace this with an actual 'ok' message from the server
+	//			once login completes within the driver
+	if _, err = clientConn.Write(connector.CreateAuthenticationOKMessage()); err != nil {
+		return connector.sendErrorWithMessage("Failed to send a successful authentication"+
+			" response to the client"+
+			": %s", err)
+	}
+
+	// Block continuation until driver has completed connection
+	<- driverConnSet
```
jonahx: Fun aside:

I didn't see this until making my other comment above on renaming `drivingConnSet`,  but note how the comment is naturally using the language ("completed connection") that I suggested for the variable name.  This is a signal to lookout for: when your comments use different language from your variables, it's often a sign that the variable should be using the comment language.  Because our comments typically use the natural, informal words with which we really think.

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: I agree that the new messaging is clearer. The original comment I wrote was inaccurate and the new one inherits from it. The truth is:

1. Secretless parses the The TDS7 prelogin-message from the client
2. Secretless incorporates content from (1) in the response to the client. At present, there's nothing from the prelogin-message that is needed for making requests to the server.
3. Secretless (doesn't currently but will do so very soon) parses the [TDS7 login](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/773a62b6-ee89-4c02-9e5e-344882630aac) from the client. This is where Secretless gets some client parameters that are then uses in generating Secretless' TDS7 login directed to the server.

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: Correct. This is entirely a client to Secretless concern.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -4,26 +4,111 @@ import (
...
+     |                          | | for intercepting data     |    |                        |
+     |                          | |---------------------------|    |                        |
+     |                          |                                  | Prelogin Request       |
+     |                          |                                  |----------------------->|
+     |                          |                                  |                        |
+     |                          |                                  |      Prelogin Response |
+     |                          |                                  |<-----------------------|
+     |                          |                                  |                        |
+     |                          |                               Intercept Prelogin Response |
+     |                          |<----------------------------------------------------------|
```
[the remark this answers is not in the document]
BradleyBoutcher: You believe correctly!

internal/plugin/connectors/tcp/mssql/connector.go
BradleyBoutcher: very good point

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -33,69 +121,112 @@ func (connector *SingleUseConnector) Connect(
...
+		if err != nil {
+			_, err = connector.sendError("failed to connect to mssql server: %s", err)
+		}
+	} ()
+
+	// Blocks continuation until we've received the preLoginResponse from the driver
+	preloginResponse := <- preLoginResponseChannel
+
+	// This encryption setting should only apply to the client -> server, but since
+	// secretless is in-between, we want to ignore it and handle this ourselves
```
jonahx: This is correct now but we've sort of buried the lede.  Can you change to:
```
// Since the communication between the client and Secretless must be unencrypted,
// we fool the client into thinking that it's talking to a server that does not support
// encryption.  
```

Issue #1039, opened 2019-12-09
Integration tests for mssql have been rewritten using sqlcmd


jonahx:
## Description
The current integration tests for the mssql connector are written using the Go unit test framework and depend on the `go-mssql` to work.  This caused conflicts and a race condition once we used `replace` on the official `go-mssql` package to use our modified fork of it.

In addition, depending on that driver package was a brittle strategy to begin with, and it's better if our integration tests rely on the official `sqlcmd` tool.

AC:
- Rewrite the ITs in bash to test the end to end flow using `sqlcmd` inside a docker container.
    - Test that we cannot connect to with bad creds when connecting directly to the DB
    - Test that we _can_ connnect with bad creds when connecting through Secretless
- Remove existing tests in the file `mssql_connector_test.go`
- MSSQL has additional integration tests that validate:
  - An app is able to connect to MSSQL via Secretless using:
    - correct username, wrong password
    - wrong username, wrong password
    - empty username, empty password
  - An app is unable to connect to MSSQL via Secretless if Secretless is configured with the wrong credentials

PR #1043, opened 2019-12-11
Integration tests for mssql have been rewritten to use sqlcmd

diverdane:
This change modifies the integration tests for the mssql connector
such that the sqlcmd is used to test access to an MSSQL server
both directly (to establish a testing baseline) and indirectly
via secretless broker.

A 'b' benchmark command line flag is introduced so that access can
be timed and compared for:
- Direct access to the MSSQL server
- Indirect access to the MSSQL server via secretless

If the 'b' benchmark flag is not selected, the test cases include:
- Direct access to the MSSQL server with correct credentials succeeds
- Direct access to the MSSQL server with bad credentials fails
- Indirect access to MSSQL server via secretless with bad credentials
  succeeds.

The test detects when the `secretless-dev` service is available, and if
so, uses that for testing (instead of `secretless`).

This change fixes secretless-broker Issue #1039.

Dockerfile and test file.

- [ No ] Manually tested [Keychain provider](https://github.com/cyberark/secretless-broker/tree/master/test/providers/keychain)
- [ No ] Manually run [Kubernetes-Conjur demo](https://github.com/conjurdemos/kubernetes-conjur-demo) with a local Secretless Broker image build of your branch

- [ n/a ] An updated README with instructions on how to manually test this feature
- [ n/a ] Utility `start` and `stop` scripts to spin up and tear down the test environments
- [ n/a ] A `test` script to run some basic manual tests (optional; if does not exist, the README should have detailed instructions)


test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,133 @@
...
+
+query_sql_service() {
+  # query_sql_service <server> <port> <user> <password>
+  server=$1
+  port=$2
+  user=$3
+  pw=$4
+
+  docker_cmd="docker-compose run --rm --no-deps"
+  sql_cmd="sqlcmd -S \"$server,$port\" -U \"$user\" -P \"$pw\" -Q \"SELECT 1+1\""
```
jonahx: Fly by comment:

First, this is a very clean bash.  A trick to avoid all the quote escaping when you have this kind of double substitution situation -- creating a variable (that itself has expanded variables) that will be expanded later-- is to use bash arrays.  If you do:
```sh
sql_cmd=(sqlcmd -S "$server,$port" -U "$user" -P "$pw" -Q "SELECT 1+1")
```
and then later when you use it (note the required quotes around the array which expands it with each element quoted):
```sh
$docker_cmd test bash -c "${sql_cmd[@]}" 2>&1 || true
```
everything will be kosher.   A small improvement imo.

diverdane: Ah, yes, thanks! I had tried that earlier using bash arrays, but it wasn't working because I was passing SQL command via `bash -c`. Removing the `bash -c` then allows the bash array trick to work.
Thanks for the explanation `note the required quotes around the array which expands it with each element quoted`, that was the piece of the puzzle that I was missing to understand how this bash array expansion works!

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,133 @@
...
-docker-compose run \
-  -e SECRETLESS_HOST="$SECRETLESS_HOST" \
-  --rm \
-  --no-deps \
-  test bash -c "go test -v ./test/connector/tcp/mssql"
+echo "Waiting for '$secretless_host' service to start"
+
+# single quotes are intentional:
+# shellcheck disable=SC2016
+docker-compose run --rm --no-deps test bash -c '
```
sgnn7: You want `-ec`, not just `-c` I think to ensure that errors propagate back up

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,133 @@
...
+  $docker_cmd test "${sql_cmd[@]}" 2>&1 || true
+}
+
+# Automatically detect if we're devmode based on the existence
+# of the secretless-dev container.  We assume that you started
+# your workflow using `./dev` if you are developing, and this
+# command will use the secretless-dev container.
 if [[ ! -z $(docker-compose ps -q secretless-dev) ]]; then
-  export SECRETLESS_HOST=secretless-dev
+  secretless_host="secretless-dev"
```
jonahx: comment on existing code, but since we're here we should change:
```sh
if [[ ! -z $(docker-compose ps -q secretless-dev) ]]; then
```
to (untested):
```sh
if docker-compose ps -q secretless-dev 2>&1; then
```

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,133 @@
...
+# shellcheck disable=SC2016
+docker-compose run --rm --no-deps test bash -c '
+counter=0
+while ! wget --quiet --output-document - http://'$secretless_host':5335/ready > /dev/null 2>&1; do
+    if expr $counter = 5 > /dev/null; then
+      echo ""
+      echo "Timed out waiting for Secretless"
+      exit 1
+    fi
+    counter=$counter+1
```
sgnn7: I don't think the code here does what you want it to do:
```
$ foo=0
$ foo=$foo+1
$ echo $foo
0+1
```

jonahx:
```sh
let "counter = $counter + 1"
```
will work or you can use an arithmetic expansion `$(( ... ))`

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,133 @@
...
+  assert_contains "$good_resp" "1 rows affected"
+
+  echo "Test: Query mssql service directly with incorrect password fails"
+  bad_resp=$(query_sql_service $db_host_tls $db_port "BAD_USER" "BAD_PASSWORD")
+  assert_contains "$bad_resp" "Login failed for user"
+
+  echo "Test: Query mssql via secretless with incorrect password succeeds"
+  good_resp=$(query_sql_service $secretless_host $secretless_port "NONEXISTENT_USER" "BOGUS_PASSWORD")
+  assert_contains "$good_resp" "1 rows affected"
+fi
```
jonahx: Might be worth raising in the main channel, but even though it's cool you were able to add this in a simple way, I think we're mixing concerns by putting the benchmark code here.  I think this should be purely the IT itself.

We already have another tool for benchmarking, so this may be redundant, but it we do want to keep it just a quick and simple benchmark for this code, I think we should move it into its own `benchmark` file that would be run manually and separate from the IT.

test/connector/tcp/mssql/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM mcr.microsoft.com/mssql-tools:latest
+
+ENV PATH $PATH:/opt/mssql-tools/bin
+
+# Install wget
+RUN apt-get update -y && \
```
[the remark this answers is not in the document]
diverdane: Yes, will fix that.

test/connector/tcp/mssql/test
diverdane: That doesn't seem to work. What I'm seeing when I try this is that conditional that you're suggesting looks at the `exit value` of the `docker-compose ps ...` command. In this case, the `docker-compose ps ...` command returns a success exit value of 0 whether the `secretless-dev` container is running or not. So we need to look at stdout from this command, i.e. use command substitution. Re. the stderr, I *think* we want to ignore stderr in this case, and base our decision solely on whether or not there's stdout for the command. One minor improvement, though, would be to use `-n` instead of `! -z`:
```
if [[ -n $(docker-compose ps -q secretless-dev) ]]; then
```

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,133 @@
...
+  echo "The string:"
+  echo ""
+  echo "    ${string}"
+  echo ""
+  echo "Did not contain expected substring:"
+  echo ""
+  echo "    ${substring}"
+  echo ""
+
+  exit 1
```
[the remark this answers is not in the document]
diverdane: Ooh, yeah. This is a pattern that exists in other plugin tests (e.g. https://github.com/cyberark/secretless-broker/blob/master/test/connector/http/basicauth/test#L19-L40). I'll fix this in the mssql test. Should I open a tech-debt issue for the others?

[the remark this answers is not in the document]
jonahx: Definitely.  Thanks.

test/connector/tcp/mssql/test
diverdane: Right!

test/connector/tcp/mssql/test
jonahx: From what I'm seeing the exit code should work:

```sh
~/cyberark/secretless-broker/src/test/connector/tcp/mssql (master) $ dc ps
         Name                       Command                State             Ports         
-------------------------------------------------------------------------------------------
mssql_mssql_1            /opt/mssql/bin/nonroot_msg ...   Up         0.0.0.0:1433->1433/tcp
mssql_secretless-dev_1   ./bin/reflex                     Up         0.0.0.0:2223->2223/tcp
mssql_secretless_1       /usr/local/bin/secretless- ...   Exit 128                         
mssql_test_1             sleep 999d                       Up                               
~/cyberark/secretless-broker/src/test/connector/tcp/mssql (master) $ dc ps -q secretless-dev; echo $?
8ca8cacd29772cb5f294db45f6d4227ecb4eb88a8eeff40a846ab00fec26d373
0
~/cyberark/secretless-broker/src/test/connector/tcp/mssql (master) $ dc ps -q xxxxxx; echo $?
ERROR: No such service: xxxxxx
1
```

test/connector/tcp/mssql/test
diverdane: Yep, if this is redundant, I'll remove it. The only (limited) value to adding a separate `benchmark` script here would be that we'd have something really quick to check that we're not somehow adding obscene amounts of latency with the insertion of secretless into the mix... but if this is being checked regularly in IT, there's not much value gained.

test/connector/tcp/mssql/test
jonahx: Oh also, since quiet still displays the IDs we want:
```sh
if docker-compose ps --quiet secretless-dev 1>/dev/null 2>&1; then
```

jonahx: Nevermind, you're correct.  Ignore my above comments.  It still succeeds if the service exists but it not running.

test/connector/tcp/mssql/test
diverdane: Yikes! Good catch!

test/connector/tcp/mssql/test
jonahx: Well the `juxtaposer` benchmarks I believe are only run before releases (ofc you _can_ run them whenever you want but normally people don't), so there's some value to moving it into a little benchmark script.

diverdane: I think I'll leave this out for now... Since there would be a lot of shared code between `test` and `benchmark`, it would take a bit of refactoring, e.g creating a test_common file with shared functions, constants, etc.

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,124 @@
...
+echo "Running tests"
+echo ""
 
-# print secretless logs after tests run
-ret=$?
-docker-compose logs $SECRETLESS_HOST
-exit $ret
+echo "Test: Query mssql service directly with correct password succeeds"
+good_resp=$(query_sql_service $db_host_tls $db_port $db_user $db_password)
+contains_substring "$good_resp" "1 rows affected"
```
BradleyBoutcher: Nit: Should we store the expected output like '1 rows affected' as local variables, for reuse?

diverdane: Yes, I like that idea. It would make it easier to change if we ever need to.

test/connector/tcp/mssql/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM mcr.microsoft.com/mssql-tools:latest
+
+ENV PATH $PATH:/opt/mssql-tools/bin
+
+# Install wget. But before we do the required `apt-get update`, we first
+# remove the Microsoft package distro list that was created in the base
+# image, so that we're not subject to connection/checksum issues in that
+# distro (as we've bitten by in the past).
```
jonahx: Could you add a little more context?  I'm still a little hazy on exactly what happened.  I will definitely have no idea when I read this in 2 months.

diverdane: I reworded the comment, adding more details.

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,126 @@
...
+  # and we need to "|| true" because we don't want the script to exit when this
+  # happens
+  $docker_cmd test "${sql_cmd[@]}" 2>&1 || true
+}
+
+# Automatically detect if we're devmode based on the existence
+# of the secretless-dev container.  We assume that you started
+# your workflow using `./dev` if you are developing, and this
+# command will use the secretless-dev container.
+if [[ -n $(docker-compose ps -q secretless-dev) ]]; then
```
jonahx: Let's move this to the top, right after line 12, since it's basically global script configuration, and is essentially part of the "definition" of `secretless_host`

test/connector/tcp/mssql/test
```diff
@@ -1,17 +1,126 @@
...
+}
+
+while getopts "h" opt; do
+    case $opt in
+        h) help_and_exit 0;;
+        *) echo "Unknown option -$OPTARG"; help_and_exit 1;;
+    esac
+done
+
+contains_substring() {
```
jonahx: We should rename this to `assert_contains_substring` since the name `contains substring` suggest that it's a pure function that returns a boolean.  In fact, it's part of a little mini "test framework" we've built that checks an assertion, prints output, and alters the state of a global variable.

diverdane: Yes, that makes sense.

test/connector/tcp/mssql/test
diverdane: Yes, also makes sense.

PR #1045, opened 2019-12-12
client params passed to backend via Secretless

doodlesbykumbi:
This PR makes it so that client's params are propagated via Secretless. At present, it only focuses on app name and database.

This PR relies on the ReadLogin method introduced over at https://github.com/cyberark/go-mssqldb/pull/6. The build will fail until that dependency is merged.


doodlesbykumbi:
Superseded by https://github.com/cyberark/secretless-broker/pull/1049

PR #1049, opened 2019-12-16
propagate client params via secretless

doodlesbykumbi:
This PR makes enables client's param propagation via Secretless. This propagates the params that go-mssqldb supports via the Login packet:

+ Database
+ HostName
+ ServerName
+ AppName
+ TypeFlags

See https://github.com/cyberark/go-mssqldb/blob/master/tds.go#L988-L999.

~This PR relies on the ReadLogin method introduced over at cyberark/go-mssqldb#6. The build will fail until that dependency is merged.~

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connector.go
+++ b/internal/plugin/connectors/tcp/mssql/connector.go
@@ -9,6 +9,7 @@ import (
 	"github.com/cyberark/secretless-broker/pkg/secretless/log"
 	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
 	"github.com/cyberark/secretless-broker/third_party/ctxtypes"
+	"github.com/pkg/errors"
 
 	mssql "github.com/denisenkom/go-mssqldb"
 )
@@ -129,13 +130,17 @@ func (connector *SingleUseConnector) Connect(
 	clientBuffer := mssql.NewTdsBuffer(bufferSize, connector.clientConn)
 	_, err := mssql.ReadPreloginWithPacketType(clientBuffer, mssql.PackPrelogin)
 	if err != nil {
-		return connector.sendError("failed to read prelogin request: %s", err)
+		wrappedError := errors.Wrap(err, "failed to read prelogin request")
+		connector.sendError(wrappedError)
+		return nil, wrappedError
 	}
 
 	// Prepare connection details from the client, formatted for MsSQL
 	connDetails, err := NewConnectionDetails(credentialValuesByID)
 	if err != nil {
-		return connector.sendError("Unable to create new connection details: %s", err)
+		wrappedError := errors.Wrap(err, "unable to create new connection details")
+		connector.sendError(wrappedError)
+		return nil, wrappedError
 	}
 
 	// Create a new MsSQL connector
@@ -146,7 +151,9 @@ func (connector *SingleUseConnector) Connect(
 	// "connector", and its connectors also have a "Connect" method.
 	driverConnector, err := mssql.NewConnector(dataSourceName(connDetails))
 	if err != nil {
-		return connector.sendError("Failed to create a go-mssqldb connector: %s", err)
+		wrappedError := errors.Wrap(err, "failed to create a go-mssqldb connector")
+		connector.sendError(wrappedError)
+		return nil, wrappedError
 	}
 
 	// Create the context for our connection
@@ -155,24 +162,32 @@ func (connector *SingleUseConnector) Connect(
 	// Create a channel for receiving the prelogin response through the context
 	preLoginResponseChannel := make(chan map[uint8][]byte)
 
-	// Set a 'marker' for when the driver has finished connecting to the server
-	connectPhaseFinished := make(chan struct {})
+	// Create a channel for send the client login to the driver through the context
+	clientLoginChannel := make(chan mssql.Login)
 
-	// Add channels to the context, to retrive information from the driver
-	loginContext := context.WithValue(ctx, ctxtypes.PreLoginResponseKey,
+	// Set a 'marker' for when the driver has finished connecting to the server
+	connectPhaseFinished := make(chan struct{})
+
+	// Add channels to the context, to exchange information with the driver
+	loginContext := context.WithValue(
+		ctx,
+		ctxtypes.PreLoginResponseKey,
 		preLoginResponseChannel)
+	loginContext = context.WithValue(
+		loginContext,
+		ctxtypes.ClientLoginKey,
+		clientLoginChannel)
 
 	// Build a new driver connection
 	var driverConn driver.Conn
 
 	go func() {
 		// Kick off authentication through our third party connector
+		// TODO: this is generally dangerous
+		//  we're sharing state between the main routine and this, and have the channel
+		//  working as an implicit lock
 		driverConn, err = driverConnector.Connect(loginContext)
 		connectPhaseFinished <- struct{}{}
-
-		if err != nil {
-			_, err = connector.sendError("failed to connect to mssql server: %s", err)
-		}
 	} ()
 
 	// Blocks continuation until we've received the preLoginResponse from the driver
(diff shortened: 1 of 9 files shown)


internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -129,13 +129,21 @@ func (connector *SingleUseConnector) Connect(
 	clientBuffer := mssql.NewTdsBuffer(bufferSize, connector.clientConn)
 	_, err := mssql.ReadPreloginWithPacketType(clientBuffer, mssql.PackPrelogin)
 	if err != nil {
-		return connector.sendError("failed to read prelogin request: %s", err)
+		newError := fmt.Errorf(
```
diverdane: The errors.Wrap() function might be handy here:
```
import "github.com/pkg/errors"
```
and:
```
    newError := errors.Wrap(err, "failed to read prelogin request")
```
and similarly for other error wrappings below.

diverdane: Or could you redefine connector.sendError() to do the wrapping?:
```
func (connector *SingleUseConnector) sendError(err error, wrapStr string) {
   . . .
}
```

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -204,14 +242,12 @@ func (connector *SingleUseConnector) Connect(
 	// TODO: 	Replace this with an actual 'ok' message from the server
```
diverdane: These are existing comments (before your change), but can't these TODOs be combined (i.e. isn't the 'ok' message the same as the login response)?

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: wrappers the errors themselves so that the wrapped error could be returned

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: you're rght

test/connector/tcp/mssql/mssql_connector_test.go
```diff
@@ -50,6 +50,42 @@ func TestMSSQLConnector(t *testing.T) {
...
+		err = conn.Ping()
+		assert.NoError(t, err, "valid db should not error")
+	})
+
+	t.Run("Passes invalid database name to MSSQL through Secretless", func(t *testing.T) {
+		port := 2223
+		server := testutil.SecretlessHost
+
+		// Open the connection
+		connString := fmt.Sprintf("server=%s;port=%d;encrypt=disable;database=meow", server, port)
```
diverdane: 😸

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -186,32 +201,40 @@ func (connector *SingleUseConnector) Connect(
...
+		return nil, wrappedError
 	}
 
-	// Just like above, we don't need to concern ourselves with what the client would
-	// typically be sending to the server, since we are handling all communication with
-	// the server ourselves. We read the next packet so the client isn't blocked,
-	// while the driver handles the handshake.
-	err = clientBuffer.ReadNextPacket()
+	// We parse the client's Login packet so that we can pass on params to the server.
+	clientLogin, err := mssql.ReadLogin(clientBuffer)
```
diverdane: Just a thought that's not directly tied to your change here, sparked by a comment that you made last week on whether our approach for the mssql connector is too prescriptive. We're assuming everything between client and server falls in lock-step, but I wonder if we'll recover (eventually) if one side or the other gets out of step.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -186,32 +201,40 @@ func (connector *SingleUseConnector) Connect(
...
-
-	// TODO: 	Replace this with an actual 'ok' message from the server
-	//			once login completes within the driver
-	if _, err = clientConn.Write(connector.CreateAuthenticationOKMessage()); err != nil {
-		return connector.sendError("Failed to send a successful authentication"+
-			" response to the client"+
-			": %s", err)
-	}
+	// Send the client login so that secretless can send honest client params to the server
+	clientLoginChannel <- *clientLogin
```
jonahx: I'd like to see a clearer comment here, to call out that this channel will by read inside of the mssql driver, in code that exists in another repo.  Even though I _knew_ this coming in, I still did a double take before realizing what was happening.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -186,32 +201,40 @@ func (connector *SingleUseConnector) Connect(
...
 
 	// Block continuation until driver has completed connection
 	<- connectPhaseFinished
+	if err != nil {
+		wrappedError := errors.Wrap(err, "failed to connect to mssql server")
+		connector.sendError(wrappedError)
+		return nil, wrappedError
+	}
```
jonahx: I don't see any way this err could have been set since line 215...

izgeri:
@doodlesbykumbi do we need to log a follow-on issue to transmit the rest of the client params? if not, this should be in the README and in the docs as a known limitation (that we only support a couple of specific ones). we should probably also check in with @alexkalish about that - so if you have more info about the path forward / any challenges around passing all params, please lmk tomorrow

test/connector/tcp/mssql/mssql_connector_test.go
```diff
@@ -50,6 +50,42 @@ func TestMSSQLConnector(t *testing.T) {
...
+		if err != nil {
+			return
+		}
+		defer conn.Close()
+
+		// Ping the DB
+		err = conn.Ping()
+		assert.Error(t, err, "invalid db should error")
+	})
+
```
jonahx: Was thinking about this a bit more... both of these feel more like integration tests that we should be doing through `sqlcmd`.  They're really not unit tests.  There are no doubles or mocks.  We're not testing our application logic directly.  We're testing that the whole thing works end to end.

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: @jonahx The go-routine that supplies a value to `connectPhaseFinished`. I left a comment about how this isn't a clear way of doing things.

alexkalish:
Regarding client params, I think it's fine to not support every single one, assuming they cannot be handled generically and support needs to be added individually.  But as @izgeri, please be clear about this in limitations.  Thanks.

test/connector/tcp/mssql/mssql_connector_test.go
doodlesbykumbi: @jonahx 

sqlcmd and go-mssqldb are equivalent clients. The idea of the fork is for it to remain the same as the original but behave differently iff Secretless calls it. In the test cases the client is being used in its original form. You're right that we might inadvertently introduce changes to the fork that might influence the tests but I think the probability is quite low. That would be a massive bug. 

If anything we ought to run the tests with both clients... there've been bugs that have show themselves only via one client.

You're also right that switching to `sqlcmd` tets wouldn't be much work after that PR is mergedd.

jonahx:
@doodlesbykumbi Reminder to update the sequence diagram before your final rebase/merge.

test/connector/tcp/mssql/mssql_connector_test.go
jonahx: Summary of offline discussion where we hammered this out:

1. `sqlcmd` will be used for our canonical tests, since that's the official MS client and is the most reliable, independent verification that our code works.
2. Kumbi will try to use the (unforked) denis repo with his existing tests, so that we have cross validation with 2 clients.

internal/plugin/connectors/tcp/mssql/connector.go
jonahx: Oooooooh, yeah I see now.  I think we should definitely change this.  Easy fix:

`connectPhaseFinished` -> `connectResult`, and make it's type `chan error`, which can be nil or an error.

doodlesbykumbi: @jonahx i thought to do that but the issue would still exist for driverConn. i think this is code that'll need reevaluating altogether.

jonahx: how about?

`chan ConnectResult` where `ConnectResult` is a struct that holds the `(driver, err)` pair?

test/connector/tcp/mssql/mssql_connector_test.go
jonahx: Worth noting 2. could be pushed to a followon issue that could be completed any time...

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: This is resolved in a following PR ->  https://github.com/cyberark/secretless-broker/pull/1050

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: This is resolved in a following PR -> #1050

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: I think timeouts resolve this. There's a built in timeout.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -221,11 +244,11 @@ func (connector *SingleUseConnector) Connect(
...
 
 // TODO: Add ability to receive an MSSQL error and send it to the client (#1013)
-func (connector *SingleUseConnector) sendError(message string, err error) (net.Conn, 
-	error){
-	connector.logger.Errorf(message, err)
+func (connector *SingleUseConnector) sendError(err error) {
+	//NOTE: no need to log, Secretless already does this for every error sent back
+	//from the Connect method
+	//connector.logger.Error(err)
 	connector.sendErrorToClient()
```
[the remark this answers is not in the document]
BradleyBoutcher: Just saw the other PR - never mind!

test/connector/tcp/mssql/mssql_connector_utils.go
```diff
@@ -0,0 +1,144 @@
...
+	Host string
+	Port int
+	Username string
+	Password string
+	Database string
+}
+
+type dbQueryExecutor func(cfg dbConfig, query string) (string, error)
+
+func defaltSecretlessDbConfig() dbConfig {
```
diverdane: Misspelling: `defaltSecretlessDbConfig` should be `defaultSecretlessDbConfig`

test/connector/tcp/mssql/mssql_connector_test.go
```diff
@@ -0,0 +1,82 @@
...
+		_, err := queryExec(
+			cfg,
+			"",
+		)
+
+		assert.Error(t, err, "invalid db should error")
+		assert.Contains(t, err.Error(), "Generic SQL Error")
+	})
+
+	t.Run("Cannot connect directly to MSSQL", func(t *testing.T) {
```
jonahx: move next to other secretless test

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -155,24 +162,32 @@ func (connector *SingleUseConnector) Connect(
 	// Create a channel for receiving the prelogin response through the context
 	preLoginResponseChannel := make(chan map[uint8][]byte)
 
+	// Create a channel for send the client login to the driver through the context
+	clientLoginChannel := make(chan mssql.Login)
```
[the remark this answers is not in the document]
doodlesbykumbi: something to address in an upcoming refactor PR

PR #1050, opened 2019-12-17
mssql: use server responses to respond to client

doodlesbykumbi:
Resolves https://github.com/cyberark/secretless-broker/issues/1013 and https://github.com/cyberark/secretless-broker/issues/1016

1. receives protocol errors and loginAck via context
2. able to encode loginAck and error to send to client
3. move ctxtypes to go-mssqldb to avoid unnecessarily dep on secretless (twas' semi-cyclic)

Depends on https://github.com/cyberark/go-mssqldb/pull/12


internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -96,13 +98,19 @@ Overview of the connection process
...
 const bufferSize uint16 = 4096
 
+type connectResult struct {
+	conn *mssql.Conn
+	err error
+}
+
 // Connect implements the tcp.Connector func signature
 //
 // It is the main method of the SingleUseConnector. It:
```
BradleyBoutcher: We should update this class comment with any new steps

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -162,11 +173,17 @@ func (connector *SingleUseConnector) Connect(
...
 
+	// Create a channel for receiving a server error from the driver through the context
+	errChannel := make(chan mssql.Error)
+
+	// Create a channel for receiving a server loginAck from the driver through the context
+	loginAckChannel := make(chan mssql.LoginAckStruct)
+
 	// Set a 'marker' for when the driver has finished connecting to the server
-	connectPhaseFinished := make(chan struct{})
+	connectPhaseFinished := make(chan connectResult)
```
BradleyBoutcher: We've slightly modified the function of this channel (which I like), so we might need to give it a more appropriate name. Maybe connectPhaseResponseChannel or connectionResultChannel?

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -199,56 +228,99 @@ func (connector *SingleUseConnector) Connect(
...
-	connector.backendConn = mssqlConn.NetConn()
+
+	// DONE: 	Send the login response to the client 	 (#1016)
+	// DONE: 	Verify appropriate errors are passed to the client (#1013)
 
 	return connector.backendConn, nil
 }
 
 
 // TODO: Add ability to receive an MSSQL error and send it to the client (#1013)
```
BradleyBoutcher: Love this method - we can also strike this comment

internal/plugin/connectors/tcp/proxy_service.go
```diff
@@ -110,6 +110,9 @@ func (proxy *proxyService) handleConnection(clientConn net.Conn) error {
 	logger.Infof("New connection on %v.\n", clientConn.LocalAddr())
 
 	targetConn, err = proxy.connector.Connect(clientConn, backendCredentials)
+	if err == io.EOF {
```
jonahx: I'd like to see a comment explaining why we need to handle these two cases differently and (perhaps) why you chose the word "care".

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -177,6 +194,14 @@ func (connector *SingleUseConnector) Connect(
...
 		ctxtypes.ClientLoginKey,
 		clientLoginChannel)
+	loginContext = context.WithValue(
+		loginContext,
+		ctxtypes.ServerLoginAckKey,
+		loginAckChannel)
+	loginContext = context.WithValue(
+		loginContext,
+		ctxtypes.ServerErrorKey,
+		errChannel)
```
jonahx: Similar to how we created a `ConnectResult` to treat "the returned value and err" as a single unit, we should create a struct to hold all of these different channel types.  Then we'll only need to do a single `WithValue` here.  And we'll have a single type to represent "all our communication with the driver" rather than having 4 different scattered channels.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -199,56 +228,99 @@ func (connector *SingleUseConnector) Connect(
...
+			State:      0x01,
+			// Severity Class - 16 indicates a general error that can be corrected by the user.
+			Class:      16,
+			Message:    errors.Wrap(err, "secretless").Error(),
+			ServerName: "secretless",
+			ProcName:   "",
+			LineNo:     0,
+		}
+	}
+	_ = mssql.WriteError72(connector.clientBuff, protocolErr)
```
jonahx: Great example where an inability to follow line of sight indicates a deeper problem.

There are two separate things happening here:
1. Creation of the `mssql.Error`
2. Sending the error to the client

If you create one func for 1. only (call it, say, `protocolError(err error) mssql.Error`, line of sight will naturally fall out of that.

Then this function will become simply:

```go
func (connector *SingleUseConnector) sendError(err error) {
	_ = mssql.WriteError72(connector.clientBuff, protocolError(err))
}
```
---
Totally separate issue: Can't  `mssql.WriteError72` fail?  And I understand that there's perhaps no point in trying to send _that_ failure to the client, but I'd think we want to at least log that we were unable to send error X to the client.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -159,96 +167,126 @@ func (connector *SingleUseConnector) Connect(
...
+			if res.err != nil {
+				err = res.err
+			} else {
+				// Verify the driverConn is an mssql driverConn object and get its
+				// underlying transport
+				connector.backendConn = res.conn.NetConn()
+			}
+
+			break responseLoop
+		}
```
jonahx: I wrote this response to your previous version of this code, but everything still applies, double so now....

---

You must have been waiting to see my reaction on this one.... I will _not_ disappoint you :)

I think 1. the labelled loop is both unnecessary and 2. it obscured what is actual pretty simple logic.

As you note in the comments:

```go
// a loginAck is sent from the server, this will be followed by connectPhaseFinished
// unless there's an issue
```
Thus we should be able to rewrite this as:

1. a single (no loop) select which waits for just the loginAck or error.
2. an if which returns if there was an error
3. a 2nd single select (no loop) which waits for just the `connectPhaseFinished` or another error

This makes the expected sequential actions explicit, gets rid of the loop and breaks and label, and as a bonus will eliminate the `if... else`.

internal/plugin/connectors/ssh/proxy_service.go
```diff
@@ -148,7 +149,7 @@ func (proxy *proxyService) Start() error {
 			}()
 
 			go func() {
-				if err := proxy.handleConnections(chans); err != nil {
+				if err := proxy.handleConnections(chans); err != nil && err != io.EOF {
```
diverdane: Nice. This is to resiliently handle the client shutting down the connection?

internal/plugin/connectors/tcp/mssql/connector.go
diverdane: Adding to Bradley's comment, the sequence diagram should also be updated to show login being read from client and (real) login response being sent back to client. (https://github.com/cyberark/secretless-broker/pull/1050/files#diff-a4bfca951ffb2d0c843eb664663a402dL71)

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -127,8 +132,11 @@ func (connector *SingleUseConnector) Connect(
 	// it from the client and throws it away, so that client can advance to the next
 	// stage of the process.  Otherwise the client would block forever waiting for its
 	// pre-login handshake to be read.
-	clientBuffer := mssql.NewTdsBuffer(bufferSize, connector.clientConn)
-	_, err := mssql.ReadPreloginWithPacketType(clientBuffer, mssql.PackPrelogin)
+	connector.clientBuff = mssql.NewTdsBuffer(bufferSize, connector.clientConn)
+	_, err := mssql.ReadPreloginRequest(connector.clientBuff)
+	if err == io.EOF {
+		return nil, err
```
diverdane: Do we need to log anything here indicating an EOF was seen, maybe at info level? Otherwise, it seems the connection attempt is dropped silently.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -159,96 +167,126 @@ func (connector *SingleUseConnector) Connect(
...
+			connInterceptor.ClientLogin <- *clientLogin
+			// TODO: where this is consumed inside go-mssqldb, it real
+			break
+		// a protocol error is received from the server
+		case protocolErr := <- connInterceptor.ServerError:
+			err = protocolErr
+			break responseLoop
+		// a loginAck is sent from the server, this will be followed by connectionResultChan
+		// unless there's an issue
+		case loginAck = <- connInterceptor.ServerLoginAck:
```
diverdane: gofmt nit: should be:
```
	case loginAck = <-connInterceptor.ServerLoginAck:
```
(no space before connInterceptor).

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: I think that's something that can take place in the Secretless "framework" code. I think an error of `io.EOF` from any service connector should generally mean the same thing.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -159,99 +164,121 @@ func (connector *SingleUseConnector) Connect(
...
-	if err != nil {
-		wrappedError := errors.Wrap(err, "failed to handle client login")
-		connector.sendError(wrappedError)
-		return nil, wrappedError
+	var loginAck mssql.LoginAckStruct
+	// Block continuation until error or driver has completed connection
+responseLoop:	for {
+		select {
+		// preLoginResponse is received from the server
+		case preloginResponse := <-connInterceptor.PreLoginResponse:
```
diverdane: Looking at this a second (or third?) time... The code before the select does a call to driverConnector.Connect() and then sends a connection result in the connectionResultChan. We're assuming that a preloginResponse will be received before the result is seen in the connectionResultChan, but there could be a race. Should we defer the sending of the connection result, or defer the checking for it, or something like that to ensure we're processing things in the required order?

internal/plugin/connectors/ssh/proxy_service.go
jonahx: @doodlesbykumbi this looks like it's from a different PR and somehow got mixed in here?

internal/plugin/connectors/tcp/mssql/connector.go
jonahx: @diverdane It looks to me like line 181 can't be hit until (among other things) the internal code of `driverConnector.Connect` puts a value on the preloginResponse channel _and_ that value is pulled off (which can only happen via line 192).  So there should be no race.

That said, the fact that this is confusing enough that you read it that way does cement my believe we should flatten the logic here (though I'm still fine with doing that in a followon)

diverdane: Ah, right. I think I realized the first time that I reviewed this part that there was synchronization built in.

internal/plugin/connectors/ssh/proxy_service.go
doodlesbykumbi: took it out. yeah it should be another PR

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -239,111 +275,138 @@ func (connector *SingleUseConnector) Connect(
...
+	// Set a 'marker' for when the driver has finished connecting to the server
+	connectionResultChan := make(chan connectResult)
 
-	// Create a channel to send the client login to the driver through the
-	// context. See types.ReadLoginFunc for an explanation of why we need to use
-	// interface{} here, even though we're sending actually sending an
-	// mssql.Login.
-	clientLoginChannel := make(chan interface{})
+	// Create a ConnectIntercept for exchanging values with the driver via context and
+	// ensure it gets cleaned up when this method returns
```
jonahx: There is no `defer`, so it's not currently doing:

> ensure it gets cleaned up when this method returns

which confused me.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -239,111 +275,138 @@ func (connector *SingleUseConnector) Connect(
...
-	clientLoginChannel <- clientLogin
+		connectionResultChan <-connectResult{
+			conn: driverConn,
+			err:  err,
+		}
+	} ()
 
-	// Block continuation until driver has completed connection
-	<-connectPhaseFinished
+	loginResp, backendConn, err := connector.waitForLoginResponse(connInterceptor, connectionResultChan)
```
jonahx: line too long

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -239,111 +275,138 @@ func (connector *SingleUseConnector) Connect(
...
+	var loginResponse mssql.LoginResponse
+	var err error
+
+// SECTION: wait for prelogin response
+	select {
+	// preloginResponse is received from the server
+	case preloginResponse := <-interceptor.ServerPreLoginResponse:
+		// Since the communication between the client and Secretless must be unencrypted,
+		// we fool the client into thinking that it's talking to a server that does not support
+		// encryption.
```
jonahx: nit: wrap lines

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -239,111 +275,138 @@ func (connector *SingleUseConnector) Connect(
...
-	if _, e := connector.clientConn.Write(mssqlError); e != nil {
-		connector.logger.Errorf("failed to write error %s to MSSQL client", e)
+func (connector *SingleUseConnector) waitForLoginResponse(
+	interceptor *mssql.ConnectInterceptor,
+	connResChan chan connectResult,
+) (*mssql.LoginResponse, net.Conn, error) {
+	var loginResponse mssql.LoginResponse
+	var err error
+
+// SECTION: wait for prelogin response
```
jonahx: I'd like to see a blank line after the SECTION comment, which visually makes it feel more section like and distinguishes it from explanatory comments which touch the line below them.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -239,111 +275,138 @@ func (connector *SingleUseConnector) Connect(
...
+			return nil, nil, wrappedError
+		}
+
+		// Send the client login to the mssql driver which it can use to pass client params
+		// to the server. This channel is set as a value on the context passed to the mssql
+		// driver on construction.
+		interceptor.ClientLoginRequest <- clientLoginRequest
+		break
+	// connect is finished. this case is visited when there is non-protocol error or
+	// after loginResponse
```
jonahx: I didn't understand this comment

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -239,111 +275,138 @@ func (connector *SingleUseConnector) Connect(
...
+		LineNo:     0,
+	}
 }
 
-// TODO: Add ability to receive an MSSQL error and send it to the client (#1013)
-func (connector *SingleUseConnector) sendErrorToClient() {
-	mssqlError := connector.CreateGenericErrorMessage()
-	if _, e := connector.clientConn.Write(mssqlError); e != nil {
-		connector.logger.Errorf("failed to write error %s to MSSQL client", e)
+func (connector *SingleUseConnector) waitForLoginResponse(
```
jonahx: I think we want a more accurate name here, since we're not merely waiting for a login response, we're performing a number of stages in the protocol sequence.  

Would `DoPreloginAndLoginSteps` be accurate?  Generally "and" names are a code smell, but it's nbd and that's something we can address later if we want when we refactor.  Here my only concern is that the method name accurately described what the method does.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -239,111 +275,138 @@ func (connector *SingleUseConnector) Connect(
...
+		// We parse the client's LoginRequest packet so that we can pass on params to the server.
+		clientLoginRequest, err := connector.readLoginRequest(connector.clientBuff)
+		if err != nil {
+			wrappedError := errors.Wrap(err, "failed to handle login from client")
+			return nil, nil, wrappedError
+		}
+
+		// Send the client login to the mssql driver which it can use to pass client params
+		// to the server. This channel is set as a value on the context passed to the mssql
+		// driver on construction.
```
jonahx: same here, wrap lines

internal/plugin/connectors/tcp/proxy_service.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net"
 
 	validation "github.com/go-ozzo/ozzo-validation"
+	"github.com/pkg/errors"
```
jonahx: The changes in this file really feel like a separate PR, but that's ok... At the very least we should clean the history so that these changes are there own commit with a clear message explaining why we're making them and what benefits the wrapping provides us.

test/connector/tcp/mssql/mssql_connector_test.go
```diff
@@ -53,36 +54,46 @@ func RunTests(t *testing.T, queryExec dbQueryExecutor) {
...
+		// Test the returned values
 		assert.Contains(t, err.Error(), "Login failed")
 	})
 
 	t.Run("Passes valid database name to MSSQL through Secretless", func(t *testing.T) {
-		// Open connection and run test query
 		cfg := defaultSecretlessDbConfig()
-		cfg.Database = "master"
-		_, err := queryExec(
+		// existing database name, see https://docs.microsoft.com/en-us/sql/relational-databases/databases/tempdb-database?view=sql-server-ver15
```
jonahx: nit, but i'd but a linebreak after "see:"

test/connector/tcp/mssql/mssql_query.go
```diff
@@ -8,6 +8,8 @@ import (
 	"os/exec"
 	"text/tabwriter"
 
+	_ "github.com/denisenkom/go-mssqldb" // registers the "mssql" driver to database/sql
+
```
jonahx: How was this working previously?  (ie, seems like we would have always needed this...)

test/connector/tcp/mssql/mssql_connector_test.go
```diff
@@ -53,36 +54,46 @@ func RunTests(t *testing.T, queryExec dbQueryExecutor) {
 			cfg.Host = "127.0.0.1"
 		}
 
+		// Execute Query
 		_, err = queryExec(
 			cfg,
 			"",
 		)
 
+		// Test the returned values
```
jonahx: 👍 to these section comments

also, similar to my comment above, i think these integration test changes need to be in their own commit.

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: went for `WaitForServerConnection`. i think when the refactor takes place then there'll be a better arrangement but for now that method is essentially waiting

Issue #1017, opened 2019-11-22
MSSQL connector OK message has tests against different protocol and server versions


izgeri:
Note: This User Story depends upon Issue #1016  "The MSSQL connector faithfully passes the login response"

Some Background:
The OK message is a.k.a. the TDS login response, and this response message contains a LOGINACK token stream (details in links below).

The TDS login response is documented here:
[MS-TDS Login Response](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/da5a73bb-e492-47ab-9d9c-23ac00292efc)
with details about fields here:
[LOGINACK](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/490e563d-cc6e-4c86-bb95-ef0186b98032)

Here's a sample LOGINACK token stream, taken from a Wireshark capture:
Token - LoginAck
    Token length: 54
    Interface: 1
    TDS version: 0x74000004
    Server name: Microsoft SQL Server
    Server Version: 14.0.3257

Here is the current published version of TDS:
[current published version of TDS](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/b46a581a-39de-4745-b076-ec4dbb7d13ec#published-version) and the [previous versions of TDS](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/b46a581a-39de-4745-b076-ec4dbb7d13ec#previous-versions).

## Versions of TDS to test:
Comparing previous versions of the Login response, the syntax has not changed since 2013 (or perhaps earlier). The TDS version that is included in the MSSQL server docker images hasn't changed since at least 2017 (all tagged images use 7.4).

## Versions of MSSQL server to test:
Microsoft has published various images of the [SQL server docker image](https://hub.docker.com/_/microsoft-mssql-server). Some tags that can be used:
`2017-latest`
`2019-latest`   (same as `latest`?)
`2017-CU1-ubuntu` (oldest image?)

AC:
- [x] Test the OK message with different TDS versions and Server versions

PR #1051, opened 2019-12-18
Adds IT tests for different versions of MSSQL servers

diverdane:
Adds IT tests for different versions of MSSQL servers

Adds IT tests for different versions of MSSQL servers

This test extends the MSSQL integration tests by adding test for
running the existing test suite against the following MSSQL server
versions:
- 2019-latest
- 2017-CU1-ubuntu
in addition to the `2017-latest` version that is used in the existing
test.

The integration tests for each new server version have been added in their
own subdirectory in the MSSQL test directory. This will allow for
different server versions to be incrementally added/removed over time
without affecting the default testing that uses MSSQL server version
2017-latest.

Each MSSQL server version test subdirectory has its own versions of
docker-compose.yml and secretless.yml that use different values of
host ports for the various services so that they can run concurrently.
Each subdirectory also has versions of the 'start', 'stop', 'test',
'test-local', and 'wait_for_mssql' scripts, but these scripts mostly
delegate to their counterparts in the parent test directory (so that any
changes to scripts in the parent directory will not have to be duplicated
in the subdirectories).

Addresses serverless-broker Issue #1017.

Connected to #1017 

In the server-2017-CU1 and server-2019 subdirectories.

- [ No ] Manually tested [Keychain provider](https://github.com/cyberark/secretless-broker/tree/master/test/providers/keychain)
- [ No ] Manually run [Kubernetes-Conjur demo](https://github.com/conjurdemos/kubernetes-conjur-demo) with a local Secretless Broker image build of your branch

- [ n/a ] An updated README with instructions on how to manually test this feature
- [ n/a ] Utility `start` and `stop` scripts to spin up and tear down the test environments
- [ n/a ] A `test` script to run some basic manual tests (optional; if does not exist, the README should have detailed instructions)

### DIFF
--- a/test/connector/tcp/mssql/mssql_connector_test.go
+++ b/test/connector/tcp/mssql/mssql_connector_test.go
@@ -2,6 +2,7 @@ package mssqltest
 
 import (
 	"fmt"
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -37,9 +38,14 @@ func RunTests(t *testing.T, queryExec dbQueryExecutor) {
 	})
 
 	t.Run("Cannot connect directly to MSSQL", func(t *testing.T) {
+		// Set Host and Port to $DB_HOST_TLS and $DB_PORT environment
+		// variables, respectively.
+		envCfg := testutil.NewDbConfigFromEnv()
 		cfg := defaultSecretlessDbConfig()
-		cfg.Port = 1433
-		cfg.Host = "mssql"
+		cfg.Host = envCfg.HostWithTLS
+		var err error
+		cfg.Port, err = strconv.Atoi(envCfg.Port)
+		assert.NoError(t, err)
 
 		// This is for local testing. Locally, Secretless and and the target service
 		// are exposed on 127.0.0.1 via port mappings
@@ -47,7 +53,7 @@ func RunTests(t *testing.T, queryExec dbQueryExecutor) {
 			cfg.Host = "127.0.0.1"
 		}
 
-		_, err := queryExec(
+		_, err = queryExec(
 			cfg,
 			"",
 		)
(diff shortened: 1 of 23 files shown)


test/connector/tcp/mssql/start
```diff
@@ -1,21 +1,33 @@
...
+while getopts :dm:s: opt; do
     case $opt in
-        d) SECRETLESS_HOST=secretless-dev;;
+        d) dev_mode=true;;
+        m) mssql_host=${OPTARG};;
+        s) secretless_host=${OPTARG};;
        \?) echo "Unknown option -$OPTARG"; exit 1;;
     esac
 done
+# If the secretless host is not explicitly set on the command line,
```
BradleyBoutcher: 👍

test/connector/tcp/mssql/mssql_connector_test.go
```diff
@@ -37,17 +38,22 @@ func RunTests(t *testing.T, queryExec dbQueryExecutor) {
 	})
 
 	t.Run("Cannot connect directly to MSSQL", func(t *testing.T) {
+		// Set Host and Port to $DB_HOST_TLS and $DB_PORT environment
```
BradleyBoutcher: nit: for some reason this little chunk looks messy, but that could just just be me

diverdane: Roger that... What's going on here is that I need to read 2 environment variables ($DB_HOST_TLS and $DB_PORT), and merge those into the default secretless DB config. One alternative would have been to read these env variables directly via os, and merge in with defaults. But I figured there was already a testutil.NewDbConfigFromEnv() function to read config from env, so I just used that function, and merged results. The other gotcha was that $DB_PORT env is a string, so it needed an Atoi() call with an error check.

PR #1052, opened 2019-12-19
Unit test scaffolding

jonahx:
Jonah's PR


ut-plan.txt
```diff
@@ -0,0 +1,47 @@
...
+Note: These cover every possible error scenario in our code (they will, when
+I'm done with the list), and we should have a test or assertion for each of
+them)
+
+- Expected error returned and sent to client if Prelogin packet can't be read
+  from client.
+  - General note: everything sent to client is is Generic now, but that's
+    changing with @kumbirai upcoming work, so we may want to coordinate these
+    efforts.
+- Expected error returned and sent to cleint if credentials are missing or
```
diverdane: s/cleint/client/

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -89,13 +89,42 @@ Overview of the connection process
...
+// swapped out in unit tests.  Note we keep MSSQL in the name to prevent
+// confusion with the Secretless Connector.
+type NewMSSQLConnectorFunc func(dsn string) (*mssql.Connector, error)
+
+// NewSingleUseConnector creates a new SingleUseConnector
+func NewSingleUseConnector(logger log.Logger) *SingleUseConnector {
+	return NewSingleUseConnectorWithOptions(logger, mssql.NewConnector)
+}
+
+// NewSingleUseConnector creates a new SingleUseConnector, and allows you to
```
diverdane: s/NewSingleUseConnector/NewSingleUseConectorWithOptions/

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -89,19 +90,100 @@ Overview of the connection process
...
+	backendConn       net.Conn
+	clientConn        net.Conn
+	logger            log.Logger
+	newMSSQLConnector types.NewMSSQLConnectorFunc
+	readPrelogin      types.ReadPreloginFunc
+	writePrelogin     types.WritePreloginFunc
+	newTdsBuffer      types.NewTdsBufferFunc
+}
+
+// NewSingleUseConnector creates a new SingleUseConnector
```
diverdane: Maybe useful to add comment that `NewSingleUseConnector` creates a production version of a SingleUseConnector?

internal/plugin/connectors/tcp/mssql/types/types.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"io"
+	"net"
+)
+
+// NewMSSQLConnectorFunc represents the constructor of an mssqlConnector. It
+// exists so that its production implementation (mssql.NewConnector) can be
+// swapped out in unit tests.  Note we keep MSSQL in the name to prevent
+// confusion with the Secretless Connector.
+type NewMSSQLConnectorFunc func(dsn string) (MSSQLConnector, error)
```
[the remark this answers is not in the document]
jonahx: Per the slack thread started, let's go with `MSSQLConnectorCtor`.

internal/plugin/connectors/tcp/mssql/types/types.go
```diff
@@ -0,0 +1,57 @@
...
+	NetConn() net.Conn
+}
+
+// MSSQLConnectorFunc lets us treat a pure function (that matches the "Connect"
+// signature) as an MSSQLConnector interface.
+type MSSQLConnectorFunc func(context.Context) (NetConner, error)
+
+// Connect implements the MSSQLConnector interface on a pure function.
+func (fn MSSQLConnectorFunc) Connect(ctx context.Context) (NetConner, error) {
+	return fn(ctx)
```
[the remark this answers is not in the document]
jonahx: Yeah, the word "pure" here is misleading, especially since the phrase "pure function" is generally used to make a distinction around mutation / side effects.  I think I should delete that word.

> functions names can be named clearly and flexibly, they don't have to be named Connect()

Actually, the concern is more purely structural.  That is, to fulfill an interface you must define a method on some type.   Typically that type is a struct, which holds some data.  But what if there is not data to hold?  You could define a type as `struct{}` and then a method on that, but that's rather verbose boilerplate because really you only care about the method.  This trick let's you treat the plain go `func` as a type which fulfills the interface.  

This comes up a lot, and you'll see this trick other places in our code base as well as in the std lib, eg here:

https://golang.org/src/net/http/server.go?s=61509:61556#L1993

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -89,19 +90,100 @@ Overview of the connection process
...
 
 // SingleUseConnector is used to create an authenticated connection to an MSSQL target
 type SingleUseConnector struct {
-	backendConn net.Conn
-	clientConn  net.Conn
-	logger      log.Logger
+	backendConn       net.Conn
+	clientConn        net.Conn
+	logger            log.Logger
+	newMSSQLConnector types.NewMSSQLConnectorFunc
```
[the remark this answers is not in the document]
jonahx: Ah, ok this one is a little different from the previous one above, where we both agreed that `MSSQLConnectorCtor` is a better name for the _type_.  

But here we're talking about an instance of a `MSSQLConnectorCtor`, that is, a constructor function itself.  For those we have a very clear convention in Go with the `NewXXX`.

Consider how it looks when we use it:

```go
driverConnector, err := connector.newMSSQLConnector(dataSourceName(connDetails))
```

That reads exactly like normal code where we do `foo := NewFoo(blah)`.  But if we where to name the instance using ctor, we'd get:

```go
driverConnector, err := connector.mssqlConnectorCtor(dataSourceName(connDetails))
```
which makes it sound as if we're _returning_ a ctor, rather than invoking one to return a connector instance.  It also breaks the standard `NewXXX` convention.

So in this case I vote to keep the name as is -- wdyt?

And as a side note, I don't "contructor" in the name violates the rule which prohibits including the _data type_ in a name.  "constructor" is an abstract description telling you what the thing is.  The rule is more about names like `FirstNameStr` and so on.  

Arguably the `Func` suffix should be avoided though for the reason you state.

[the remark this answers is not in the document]
diverdane: Ah, yes, I absolutely agree. The sample invocation that you provided above makes it abundantly clear.

diverdane:
All looks good, 'cept for the codeclimate errors. I noticed that one of the errors says that a function is limited to a max of 5 arguments??? Can we raise this? Seems a little over the top.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -94,15 +96,108 @@ Overview of the connection process
...
 type SingleUseConnector struct {
-	backendConn net.Conn
-	clientConn  net.Conn
-	logger      log.Logger
+	backendConn       net.Conn
+	clientConn        net.Conn
+	logger            log.Logger
+	// Note: We're following standard ctor naming practices with this field.
+	newMSSQLConnector types.MSSQLConnectorCtor
+	readPrelogin      types.ReadPreloginFunc
```
diverdane: I should have thought of this before... for the `readPrelogin`, `writePrelogin`, and `readLogin` fields of the SingleUseConnector, would it be clearer to include whether the messages are request vs. acks, e.g. `readPreloginReq`, `writePreloginAck`, `readLoginReq`, or is it already clear that these apply on the client side, so request vs. ack is obvious?
When we merge this with Kumbi's message pass-through, I suspect we'll need to add a `writeLoginAck` (Or `writeLogin`).

jonahx: Yes, it would be clearer.  Let's wait till we merge with Kumbi though since we'll be making many other changes as well.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -97,12 +99,109 @@ type SingleUseConnector struct {
...
+
+// WritePreloginWithPacketType is the production version of our writePrelogin
+// dependency, which delegates to the actual 3rd party driver.
+func WritePreloginWithPacketType(
+	rawTdsBuffer interface{},
+	fields map[uint8][]byte,
+	rawPktType interface{},
+) error {
+	// This can never fail unless mssql package changes: panicking is fine
+	tdsBuffer := rawTdsBuffer.(*mssql.TdsBuffer)
```
doodlesbykumbi: 1. I'm curious why rawTdsBuffer and rawPktType are not their expected types and instead of type `interface{}` ? 
2. What was the motivation for making `PacketType` public ? I ask mostly because that's a change that moves the fork further from the original

internal/plugin/connectors/tcp/mssql/types/types.go
```diff
@@ -0,0 +1,70 @@
...
+	tdsBuffer interface{},
+	fields map[uint8][]byte,
+	pktType interface{}) error
+
+// ReadLoginFunc defins the type of the func that reads the client's login
+// packet.  The production version is implemented by:
+//     mssql.ReadLogin(r *TdsBuffer) (*Login, error)
+// Note that, in order to avoid a concrete dependency on mssql in this package,
+// we must replace TdsBuffer with ReadNextPacketer and *Login with interface{}.
+// That interface{} will then be case back to a *Login by the receiving code
```
doodlesbykumbi: s/case/cast

internal/plugin/connectors/tcp/mssql/types/types.go
```diff
@@ -0,0 +1,70 @@
...
+// mssql.WritePreloginWithPacketType.
+type WritePreloginFunc func(
+	tdsBuffer interface{},
+	fields map[uint8][]byte,
+	pktType interface{}) error
+
+// ReadLoginFunc defins the type of the func that reads the client's login
+// packet.  The production version is implemented by:
+//     mssql.ReadLogin(r *TdsBuffer) (*Login, error)
+// Note that, in order to avoid a concrete dependency on mssql in this package,
```
doodlesbykumbi: Why do we want to avoid a concrete dependency ?

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -97,12 +99,109 @@ type SingleUseConnector struct {
...
+		ReadPreloginWithPacketType,
+		WritePreloginWithPacketType,
+		ReadLogin,
+		NewTdsBuffer,
+	)
+}
+
+// NewMSSQLConnector is the production implementation of MSSQLConnectorCtor,
+// used for creating mssql.Connector instances.  We need to wrap the raw
+// constructor provided by mssql (ie, mssql.NewConnector) in this function so
```
doodlesbykumbi: s/ie,/i.e./g (there's a few other occurrences)

not sure if i'm being pedantic and what you have here is a legitimate other way of representing "id est".

internal/plugin/connectors/tcp/mssql/connector.go
jonahx: > I'm curious why rawTdsBuffer and rawPktType are not their expected types and instead of type interface{}

Good question.  Short answer: Because then our doubles would have a dep on the 3rd party driver.

> What was the motivation for making PacketType public

Because for our production impl we have to do this:

```
return mssql.ReadPreloginWithPacketType(tdsBuffer, pktType)
```
and `mssql.ReadPreloginWithPacketType` requires an argument of that type.  We can't cast to it without having the type available here.  

If you seen alternative I'm all ears....

internal/plugin/connectors/tcp/mssql/types/types.go
jonahx: This is [this whole thing](https://youtu.be/MzTcsI6tn-0?t=736) which we've discussed before.  If you have further questions let's discuss in real time in slack.

Issue #988, opened 2019-11-08
Medium severity issues flagged by gosec are resolved


izgeri:
AC:
- [ ] Scan the project with [gosec](https://github.com/securego/gosec)
- [ ] Resolve any medium (or higher) severity issues

izgeri:
optionally, you can add gosec scanning to our pipeline and fail builds with HIGH or CRITICAL severity issues, and save the output of the scan so that it can be reviewed in Jenkins.

Issue #976, opened 2019-10-30
Project with MSSQL has undergone security scan with no high or critical issues


izgeri:
AC:
- [ ] Project with MSSQL has undergone security scan with [gosec](https://github.com/securego/gosec) with no high or critical issues

PR #1053, opened 2019-12-19
976 - Introduce Security Scans for Go Packages

BradleyBoutcher:
A new bash script which runs gosec on our packages

We only flag issues of high severity with 'medium' or 'high' confidence by Gosec
Gosec only scans directories modified by checking the Git diff first. If the branch is master, it scans the entire project. This way we save time in our pipeline while developing.
Finally, the reports are exported as xml and parsed using Junit.

Connected to #976 and #988

bin/build_security_scan

NA

NA

NA


BradleyBoutcher:
@hughsaunders Thanks for the review! Addressing your comments in order:
- It absolutely can be parallelized, I'll make a few changes and re-submit it.
- So far I've tested locally and made sure that issues are detected and output correctly, but currently there are none in our repository that meet the criteria for an issue. When there's no issues, the log should be empty.
- On a related note, I realized there is an issue. Since we're scanning the git diff, we won't see which files are changed in that commit since technically nothing has been changed locally. I'll have to make a change there for sure.

Jenkinsfile
```diff
@@ -31,6 +31,13 @@ pipeline {
       }
     }
 
+    stage('Scan Go Packages') {
```
izgeri: This is a little vague - worth explicitly referring to gosec in the stage name

Jenkinsfile
```diff
@@ -31,6 +31,13 @@ pipeline {
...
+      steps {
+        sh "./bin/build_security_scan -s High -c 'Medium' -b ${env.BRANCH_NAME}"
```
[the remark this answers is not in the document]
BradleyBoutcher: Looking at the other bin/ files, it looks like we use the 'check_' format for file names. So for golint, it's check_style. How about for this we use 'check_security'?

bin/security_scan
```diff
@@ -0,0 +1,64 @@
...
+print_usage() {
+    echo "Security Scanner"
+    echo " "
+    echo "Description:"
+    echo "Runs gosec on directories which git detects and marks in the diff."
+    echo "If the branch is detected as 'master', it will scan all"
+    echo "directories regardless of what has been modified locally."
+    echo " "
+    echo "Format:"
+    echo "security_scan [arguments]"
```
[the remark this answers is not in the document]
BradleyBoutcher: Since we wanted to run it within a docker container, I thought it would be easiest to have two separate files. I'm not very familiar with bash, but I assume that's the easiest way to run a large script like this inside a docker run command?

bin/build_security_scan
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+ 
+set -eo pipefail
+
+project_dir=$PWD
```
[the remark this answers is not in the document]
BradleyBoutcher: I'm not sure I understand what you mean, can you provide an example? I was using this line to mount the parent working directory into the docker container.

bin/check_security
```diff
@@ -0,0 +1,66 @@
...
+    c) confidence="${OPTARG}" ;;
+    s) severity="${OPTARG}" ;;
+    h) print_usage ;;
+    *) print_usage ;;
+  esac
+done
+
+# This script is designed to be ran within the Secretless-Broker repository,
+# to use elsewhere, you will need to change this.
+working_directory=${PWD}/..
```
sgnn7: You should use the abspath [here](https://github.com/cyberark/secretless-broker/blob/master/bin/abspath) to figure out the path. Generally you also want to use the program `pwd` rather than the variable `PWD` if you want to use pwd.

sgnn7: If you leave as-is, you need some quotes

bin/check_security
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
sgnn7: missing `set -eo pipefail` with optional `-x`

bin/check_security
```diff
@@ -0,0 +1,66 @@
...
+# to use elsewhere, you will need to change this.
+working_directory=${PWD}/..
+
+
+# If we are on master, scan the entire repository
+modified_directories="./..."
+
+# Get an array of directories containing modified files
+if [[ ${current_branch} != 'master' ]]
+then
```
sgnn7: nit: not a blocker but we generally used the `if [[ ... ]]; then` pattern instead of the broken-by-lines one

bin/check_security
```diff
@@ -0,0 +1,66 @@
...
+then 
+    git fetch origin master:refs/remotes/origin/master
+    modified_directories=$(git diff origin/master...origin/"${current_branch}" --name-only | xargs -L1 dirname | uniq)
+fi
+
+# Exclude test files
+test_dir=${working_directory}/test
+
+# Reset output file just in case it exists
+true > "gosec.output"
```
sgnn7: Why not just remove it?

bin/run_check_security
```diff
@@ -0,0 +1,28 @@
...
+  esac
+done
+
+# gosec => Scans go packages and flags security vulnerabilities
+docker run --rm \
+    -v "$toplevel_dir/:/secretless/" \
+    secretless-dev \
+    bash -exc "
+    go get github.com/securego/gosec/cmd/gosec
+    ./bin/check_security -c ${confidence} -s ${severity} -b ${current_branch}
```
sgnn7: This line and the one above should be indented

bin/check_security
```diff
@@ -0,0 +1,67 @@
...
+# Get an array of directories containing modified files
+if [[ ${current_branch} != 'master' ]]; then 
+    git fetch origin master:refs/remotes/origin/master
+    modified_directories=$(git diff origin/master...origin/"${current_branch}" --name-only | xargs -L1 dirname | uniq)
+fi
+
+# Exclude test files
+test_dir=${toplevel_dir}/test
+
+# Reset output file just in case it exists
```
diverdane: Nittiest of nits: s/Reset/Remove/

bin/check_security
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+print_usage() {
+    echo "Security Scanner"
+    echo " "
```
diverdane: Do we need to echo a space: `echo " "`, or will just `echo` suffice?

Jenkinsfile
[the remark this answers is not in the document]
izgeri: is it your intention that the file becomes a general script to run security checks, so that when we have another we'd like to run we also do it from this file?

I'm still not sure I like `check_security` - security means a lot of different things to different people, so if we can find a way to be more clear I think that'd be good

bin/build_security_scan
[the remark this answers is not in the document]
izgeri: this script has an example: https://github.com/cyberark/secretless-broker/blob/master/bin/build_website#L19

it defines `TOPLEVEL_DIR` and uses that throughout

bin/security_scan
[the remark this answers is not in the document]
izgeri: ok, so there are two scripts because one is meant to be run in a docker container (which docker container?) and the other one actually runs the the container?

I feel like there might be some patterns for how to name and organize these that would make their intended use more clear, but I'm not close enough to the code lately to remember exactly what to point you to. @sgnn7 @jonahx do you have any examples of how we've done something like this in other cases? in particular, I'm concerned that if I look at the `bin/` dir, it's not going to be obvious which script I should run to do a `gosec` scan if we have two with similar names. I don't think this is the first time we've had to think about something like this.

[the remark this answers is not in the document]
BradleyBoutcher: Currently I have them named as such:
- check_security: performs the gosec scan with whatever parameters you need
- run_check_security: creates a docker container with secretless mounted as a volume, and runs check_security on the volume

bin/run_gosec
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
izgeri: as a follow-on to this PR, does it make sense to include this as a bash-lib script? cc @hughsaunders

PR #1056, opened 2019-12-24
Adds remote debug capability for secretless broker

diverdane:

(LOW PRIORITY)

This change adds remote debug capability for the secretless broker.
WIth this change, a Delve-capable IDE such as Intellij/Goland can attach
to a running secretless broker container, allowing breakpoints to be set,
single-step debugging, etc.

The remote debug capability will be helpful for debugging integration tests.
(Unit tests, on the other hand, can already be easily debugged directly from
the source code without remotely connecting to into a container).

This change includes a new "debug" image for the secretless broker that
adds a Delve binary (to allow remote debug connections between the secretless
broker and an IDE), and compiles the secretless broker with code optimizations
disabled (for better debugging experience).

To use remote debugging on the secretless broker, follow instructions
in the test/connector/mssql/remote_debug.md file.

Connected to #[relevant GitHub issues, eg 76]

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -1,0 +1,75 @@
+FROM golang:1.13-stretch as secretless-builder
+MAINTAINER Conjur Inc.
+LABEL builder="secretless-builder"
+
+WORKDIR /secretless
+
+# TODO: Expand this with build args when we support other arches
+ENV GOOS=linux \
+    GOARCH=amd64 \
+    CGO_ENABLED=1
+
+COPY go.mod go.sum /secretless/
+COPY third_party/ /secretless/third_party
+
+RUN go mod download
+
+# Compile Delve (for debugging)
+RUN go get github.com/go-delve/delve/cmd/dlv
+
+# secretless source files
+COPY ./cmd /secretless/cmd
+COPY ./internal /secretless/internal
+COPY ./pkg /secretless/pkg
+COPY ./resource-definitions /secretless/resource-definitions
+
+ARG TAG="dev"
+
+# The `Tag` override is there to provide the git commit information in the
+# final binary. See `Static long version tags` in the `Building` section
+# of `CONTRIBUTING.md` for more information.
+RUN go build -ldflags="-X github.com/cyberark/secretless-broker/pkg/secretless.Tag=$TAG" \
+             -gcflags="all=-N -l" \
+             -o dist/$GOOS/$GOARCH/secretless-broker ./cmd/secretless-broker && \
+    go build -o dist/$GOOS/$GOARCH/summon2 ./cmd/summon2
+
+
+# =================== MAIN CONTAINER ===================
+FROM alpine:3.8 as secretless-broker
+MAINTAINER CyberArk Software, Inc.
+
+RUN apk add -u shadow libc6-compat && \
+    # Add Limited user
+    groupadd -r secretless \
+             -g 777 && \
+    useradd -c "secretless runner account" \
+            -g secretless \
+            -u 777 \
+            -m \
+            -r \
+            secretless && \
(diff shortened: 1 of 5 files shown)


test/connector/tcp/mssql/secretless-debug/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - 40000:40000
+    security_opt:
+      - apparmor:unconfined
+      - seccomp:unconfined
+    cap_add:
+      - SYS_PTRACE
+    volumes:
+      - ../secretless.yml:/secretless.yml
+    depends_on:
+      - mssql
```
izgeri: why can't we just add this image to the `test/connector/tcp/mssql` docker-compose.yml file and update the README in that dir to include these instructions? it seems not very DRY to me to have a separate dir within the MSSQL test suite for this stuff

diverdane: Yes, good point! I moved the secretless-debug docker-compose stuff into the parent directory, and eliminated duplication in the test-local scripts. As for the README, there's no README in the parent. Should I move it up and add some notes for generic secretless testing?

go.mod
```diff
@@ -24,7 +24,6 @@ require (
 	github.com/joho/godotenv v1.2.0
 	github.com/json-iterator/go v1.1.8 // indirect
 	github.com/lib/pq v0.0.0-20180123210206-19c8e9ad0095
-	github.com/modern-go/reflect2 v1.0.1 // indirect
```
izgeri: Any idea why these deps got removed?

diverdane: Hmmm... I don't believe that they should be getting removed. The reflect2 dependency should be there because of the `github.com/json-iterator/go v1.1.8 // indirect` that shows up 2 lines earlier. Coincidentally, this dependency is the one I used as an example for 'git mod why ...' command for the Go Mod section in our Go style guide: https://github.com/conjurinc/docs/pull/139/files#diff-b48d7f4cab90a224552e7e2f355e0a71R329-R337

diverdane: My best guess is that these deletions were due to some cruft leftover in my local workspace. To check this, I undid those deletions and forced a build (which should update go.mod as necessary), and those entries stayed in go.mod (as they should). Change set has been updated.

test/connector/tcp/mssql/remote_debug.md
```diff
@@ -0,0 +1,47 @@
+# Using the Secretless-Broker Remote Debug Image
```
izgeri: can we make this just README.md in this directory? for now, it can just include remote debug instructions - but having this here would make it easier for devs to add other mssql-specific dev notes, and that can be quite handy later

izgeri: I think with that these changes could go in

diverdane: Okay, I moved the remote-debug.md to README.md, and added an intro into the README.md to describe what remote debugging is.

go.mod
izgeri: I'm still seeing them as deleted, but your comment implies this file should not have changed in this PR - can you please clarify?

diverdane: After rebase with latest master, this issue goes away.

I think the go command had been "doing the right thing" in removing these dependencies from go mod. In fact, in the latest master, these have actually been deleted. I suspect that someone may have done a prior merge of changes with a go.mod that wasn't quite up-to-date. This could happen, I suppose, if someone does a merge in a workspace where they haven't executed any go commands, or they did 'git checkout go.mod' for some reason before committing. Either way, the go command was trying to update go.mod to what it should be.

PR #1068, opened 2020-01-06
Fix security scan exiting early on master

BradleyBoutcher:
For some odd reason, gosec exits early when pipefail is used.
Pipefail is now removed from the script that runs gosec, but is included
in the script that runs the docker container.

### DIFF
--- a/bin/run_gosec
+++ b/bin/run_gosec
@@ -8,7 +8,6 @@ set -eo pipefail
 # Performs a gosec scan with given parameters on 
 # the entire local repository (in the case of master branch)
 # or on just files modified, as detected in the git diff.
-
 print_usage() {
     echo "Security Scanner"
     echo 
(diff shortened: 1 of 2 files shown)


bin/run_gosec
```diff
@@ -1,14 +1,11 @@
 #!/usr/bin/env bash
 
-set -eo pipefail
```
sgnn7: At best, this set of flags should be removed for the exact place where errors are expected and not for the whole pipeline step otherwise a bug in any part of this file will result in the whole security checking passing in Jenkins

BradleyBoutcher: So the current call pattern is
Jenkinsfile calls check_golang_security
check_golang_security builds a docker container and runs run_gosec
run_gosec runs gosec

I only removed `set -eo pipefail` from run_gosec, it still exists in check_golang_security

BradleyBoutcher: Currently, it looks like the issue is that pipefail is getting an exit code from gosec that is non zero, but isn't a failure, so gosec is ended abruptly. The only way I can get it to finish scanning master is to remove pipefail from the file that only runs gosec.

BradleyBoutcher: Currently, it looks like the issue is that pipefail is getting an exit code from gosec that is non zero, but isn't a failure, so gosec is ended abruptly. The only way I can get it to finish scanning master is to remove pipefail from the file that only runs gosec.

Issue #1070, opened 2020-01-07
MsSQL WritePrelogin Unit Test Exists


BradleyBoutcher:
TestWritePrelogin
  Test cases & AC :
- [x] expected error propogates when all other deps succeed and this one errors
-   Error is returned from connect
-   Error is written to client buffer
-   Error written to log
- [x] If the prelogin response we return from our MSSQLConnector double (by setting up the channels using context) has the encryption flag set (mssql.PreloginENCRYPTION), it is unset when it's written to the client.
- [x] All the other content that we supply is also passed to the client.

Issue #1010, opened 2019-11-21
MSSQL connector has unit tests


izgeri:
MSSQL connector has unit tests that validate (at a minimum): 
    - The most relevant messages sent between client and server are transmitted with high fidelity, eg 
      - Client options and arguments  
      - Errors  
      - Success messages 

When you pick up this card, it should be broken down into at least a couple of smaller scope cards. Please write up a short proposal of how to break it up and have the team review it in #secretless-dev

> General note: These two assertions will occur together for all or most of the
test cases.  We should write a simple helper function to avoid duplication.
  - Error is returned from connect
  - Error is written to client buffer
  - Error written to log
I'll refer to these two assetions together as "expected error propogates" below.

- [ ] TestReadPrelogin (#1069)
  - Test cases
    - expected error propogates when all other deps succeed and this one errors
    - the clientConn we pass in is passed as a first argument to this func
      (note: currently we're passing it a TdsBuffer but this will change in
      Jonah's PR)
- [ ] TestInvalidConnectionDetails (#1072)
  - Test cases
    - expected error propogates when all other deps succeed and this one errors
  - Notes
    - The details of NewConnectionDetails should not be tested here.  We can
      write a separate UT for that func if desired (see mysql connector for an
      example of that)
    - Just create a double for newConnectionDetails which is an error returning
      func.
- [ ] TestWritePrelogin (#1070)
  - Test cases
    - expected error propogates when all other deps succeed and this one errors
    - If the prelogin response we return from our MSSQLConnector double (by
      setting up the channels using context) has the encryption flag set
      (mssql.PreloginENCRYPTION), it is unset when it's written to the client.
    - All the other content that we supply is also passed to the client.
- [ ] TestThirdPartyConnect (#1073)
  - Test Cases
    - If the driverConnector.Connect never returns a prelogin response,
      Secretless fails cleanly with a timeout error which is propogated. (Come
      up with the reasonable time for this to occur... 5 seconds?  10?) 
    - Same thing but for "prelogin response suceeds, login response fails"
- [ ] TestReadLogin (#1074)
  - Test Cases
    - expected error propogates when all other deps succeed and this one errors
    - when this succeeds, we do indeed drain the client buffer of what our
      double reads
- [ ] TestWriteLogin (#1075)
  - Test Cases
    - expected error propogates when all other deps succeed and this one errors
    - when it succeeds clientBuffer does receive what our double writes to it
* Some specfics of the above will change as a result of Kumbi's
[...]

PR #1077, opened 2020-01-08
1070 MsSQL WritePrelogin Unit Test Exists

BradleyBoutcher:
This test handles the following cases
  - The production version of WritePreloginWithPacketType
  succeeds when all others succeed
  - We receive the prelogin fields from the go-mssqldb driver
  - We modify the prelogin fields before sending them to the user

Connected to #1070 

internal/plugin/connectors/tcp/mssql/connector_test.go

NA

#1010 

NA

NA

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connector_test.go
+++ b/internal/plugin/connectors/tcp/mssql/connector_test.go
@@ -2,6 +2,7 @@ package mssql
 
 import (
 	"context"
+	"io"
 	"net"
 	"testing"
 
@@ -9,7 +10,7 @@ import (
 
 	"github.com/cyberark/secretless-broker/internal/plugin/connectors/tcp/mssql/mock"
 	logmock "github.com/cyberark/secretless-broker/pkg/secretless/log/mock"
-	pluginconnector "github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
+	pluginConnector "github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
 	"github.com/denisenkom/go-mssqldb"
 )
 
@@ -17,7 +18,7 @@ func TestHappyPath(t *testing.T) {
 	logger := logmock.NewLogger()
 	clientConn := mock.NewNetConn(nil)
 	expectedBackendConn := mock.NewNetConn(nil)
-	creds := pluginconnector.CredentialValuesByID{
+	creds := pluginConnector.CredentialValuesByID{
 		"credName": []byte("secret"),
 	}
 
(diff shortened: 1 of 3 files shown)


internal/plugin/connectors/tcp/mssql/mock/mock.go
```diff
@@ -83,3 +85,16 @@ func (n *NetConn) Write([]byte) (numBytes int, err error) {
 func FakeTdsBufferCtor(r io.ReadWriteCloser) io.ReadWriteCloser {
 	return r
 }
+
+// MockTdsBufferCtor returns a Buffer (with a fake transport) that can be written & read
+func MockTdsBufferCtor(r io.ReadWriteCloser) io.ReadWriteCloser {
```
diverdane: Should drop the 'Mock' and call this 'TdsBufferCtor', otherwise some linters complain about s-s-s-stuttering, since a call would be mock.MockTdsBufferCtor. Same for 'MockTransport' below.

internal/plugin/connectors/tcp/mssql/mock/mock.go
```diff
@@ -83,3 +85,16 @@ func (n *NetConn) Write([]byte) (numBytes int, err error) {
...
+	return mssql.NewTdsBuffer(1024, new(MockTransport))
+}
+
+type MockTransport struct {
```
diverdane: Exported MockTransport will need a comment, or linter complains.

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -49,6 +51,76 @@ func TestHappyPath(t *testing.T) {
...
+		ctor,
+		mock.SuccessfulReadPrelogin,
+		writePreloginWithAssertion,
+		mock.SuccessfulReadLogin,
+		mock.MockTdsBufferCtor,
+	)
+
+	_, err := connector.Connect(clientConn, creds)
+	assert.Nil(t, err)
+	if err != nil {
```
diverdane: I think the check for `err != nil` can be removed here. There was a check for 'err != nil' in the happy path test because there was an assert that followed that needed to be skipped.

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -49,6 +51,76 @@ func TestHappyPath(t *testing.T) {
...
+	}
+
+	// Perform the actual write operation and
+	// assert that we are writing the correct fields back to the client
+	writePreloginWithAssertion := func(
+		tdsBuffer io.ReadWriteCloser,
+		fields map[uint8][]byte,
+		pktType uint8) error {
+			_ = mssql.WritePreloginWithPacketType(tdsBuffer, fields, pktType)
+			assert.Equal(t, expectedPreloginResponse, fields)
```
diverdane: COOL BEANS!!!

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -49,6 +51,76 @@ func TestHappyPath(t *testing.T) {
...
+		mssql.PreloginMARS:       {0}, // MARS disabled
+	}
+
+	// Perform the actual write operation and
+	// assert that we are writing the correct fields back to the client
+	writePreloginWithAssertion := func(
+		tdsBuffer io.ReadWriteCloser,
+		fields map[uint8][]byte,
+		pktType uint8) error {
+			_ = mssql.WritePreloginWithPacketType(tdsBuffer, fields, pktType)
```
jonahx: We don't want any references to `mssql` in any of the UTs.

Also, you should rarely (never?) need assertions _inside_ of your dependency doubles. 

I believe for the test that you're trying to do here all you need to do in this double is write your `fields` into the `tdsBuffer` (which isn't really a `tdsBuffer`, it's just a `ReadWriteCloser`).  To make that last parenthetical point clear, you might want to name it `fakeTdsBuffer`.  We can completely ignore `pktType` as that is only needed by a real `tdsBuffer`.

Now you'll have a double that does nothing but write the bytes you give it  into the buffer you give it.  In the test, you'll assert on a buffer mock that it received a `Write` call with the parameter you expect.  Or on a buffer double that the you actual  bytes you expect were written to it.  I prefer the latter approach.

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -49,6 +51,76 @@ func TestHappyPath(t *testing.T) {
...
+			assert.Equal(t, expectedPreloginResponse, fields)
+			return nil
+	}
+
+	ctor := mock.NewSuccessfulMSSQLConnectorCtor(
+		func(ctx context.Context) (net.Conn, error) {
+
+			// Pass in our mocked out preloginResponse
+			if preLoginResponse := ctx.Value(ctxtypes.PreLoginResponseKey); preLoginResponse != nil {
+				// A panic will never occur here unless the calling code is wrong
```
jonahx: Doesn't change anything conceptually, but heads up: you'll have to replace `ctxtypes.PreLoginResponseKey` with Kumbi's connectorInterceptor thingy since that will land first thing tomorrow.

internal/plugin/connectors/tcp/mssql/connector_test.go
jonahx: Indeed, the func can just end on line 118.

internal/plugin/connectors/tcp/mssql/mock/mock.go
```diff
@@ -94,3 +95,18 @@ func (n *NetConn) Write([]byte) (numBytes int, err error) {
 func FakeTdsBufferCtor(r io.ReadWriteCloser) io.ReadWriteCloser {
 	return r
 }
+
+// TdsBufferCtor returns a Buffer (with a fake transport) that can be written & read
+func TdsBufferCtor(r io.ReadWriteCloser) io.ReadWriteCloser {
+	return mssql.NewTdsBuffer(1024, new(Transport))
```
doodlesbykumbi: I'm wondering if an additional Transport Type is necessary, whatever is passed to TdsBufferCtor should be the transport.

BradleyBoutcher: I'm not sure I understand,  can you explain this more? We just need the mock to perform write operations without failing, we don't need an actual transport though.

Issue #1069, opened 2020-01-06
MSSQL plugins ReadPrelogin method has unit tests


diverdane:
## Description
The readPrelogin() method for the MSSQL plugin's SingleUseConnector needs
unit tests to cover error conditions.

Test cases should include:

- Expected error "propogates" when all other dependencies succeed and this
   one errors. In this case, the error propogation should include:
   * Expected error is returned from the SingleUseConnector's Connect() method
   * Expected error is written to client buffer
   * Expected error is written to log
  - Error written to log

AC:
- [ ] Test case available: Expected error is returned from Connect() method
- [ ] Test case available: Expected error is written to client buff
- [ ] Test case available: Expected error is written to log

PR #1079, opened 2020-01-09
Add failing readPrelogin unit test

diverdane:
This change adds a test case for a failing readPrelogin call.

Some more code will need to be added after Kumbi's PR #1050 merges
in order to actually test that an error is being sent to the client.

This change also adds a change to the secretless mock logger to
record errors being logged. This is added as a "straw man"
proposal.

Connected to #1069

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connector_test.go
+++ b/internal/plugin/connectors/tcp/mssql/connector_test.go
@@ -2,9 +2,12 @@ package mssql
 
 import (
 	"context"
+	"github.com/cyberark/secretless-broker/internal/plugin/connectors/tcp/mssql/types"
+	"io"
 	"net"
 	"testing"
 
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/cyberark/secretless-broker/internal/plugin/connectors/tcp/mssql/mock"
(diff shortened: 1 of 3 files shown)


internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -49,6 +50,33 @@ func TestHappyPath(t *testing.T) {
...
+
+func TestFailingReadPrelogin(t *testing.T) {
+	logger := logmock.NewLogger()
+	clientConn := mock.NewNetConn(nil)
+	creds := connector.CredentialValuesByID{}
+
+	testError := "injected readPrelogin error"
+	connector := NewSingleUseConnectorWithOptions(
+		logger,
+		nil,
```
BradleyBoutcher: Don't we want to test that ReadPrelogin fails when all other dependencies succeed? Shouldn't we still mock those our, or will nil be sufficient?

diverdane: I suppose it depends on how black-box-ish we want to be with UT. When the readPreloginRequest fails, it bails out of our Connect() fairly early. So technically, the other dependencies are a don't care for this test. My preference would be to leave these as nil to leave the testing as minimalist as possible.

doodlesbykumbi: @diverdane this makes the tests less black-box-ish though, which makes them brittle in the event of arbitrary rearrangement of code inside Connect.

diverdane: @BradleyBoutcher, @doodlesbykumbi  I refactored the happy path test and the failing read login tests so that there are default functions being used/shared. So testing is more black box-ish, although I'm sure we can replace what I have with what @doodlesbykumbi is working on.

PR #1087, opened 2020-01-13
Introduce functional options for connector

doodlesbykumbi:


internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -153,24 +153,24 @@ func NewMSSQLConnector(dsn string) (types.MSSQLConnector, error) {
...
 func NewSingleUseConnectorWithOptions(
-	logger log.Logger,
-	newMSSQLConnector types.MSSQLConnectorCtor,
-	readPreloginRequest types.ReadPreloginRequestFunc,
-	writePreloginResponse types.WritePreloginResponseFunc,
-	readLoginRequest types.ReadLoginRequestFunc,
-	writeLoginRequest types.WriteLoginResponseFunc,
-	writeError types.WriteErrorFunc,
-	newTdsBuffer types.TdsBufferCtor,
+	setters ...types.ConnectorOption,
```
jonahx: I think if we're naming the type `ConnectorOption` (which is reasonable since it is the "functional options" pattern ) then we should stay consistent and this line should be:
```go
options  ...types.ConnectorOption
```
and then in your loop below you do `option(args)`.  

This is the naming Dave uses in that article.  But unnecessarily introducing a new name for a thing (having  "options" and "setters" which mean the same, in this case) should, as a general principle, always be avoided when possible.

Normally I try avoid naming where we "call" a noun (as in `option(args)`) and opt instead for `verb(noun)`, but in this case it seems reasonable to break the rule because of the pattern name, the type name, and the benefits of consistency.

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -1,59 +1,372 @@
...
+		connectorOption,
+	)
+
+	_, actualErr = runDefaultTestConnect(connector)
+	expectedClient := connector.clientConn
+
+	// confirms error returned by #Connect contains expected error
+	assert.Equal(t, expectedErr, errorspkg.Cause(actualErr))
+	// confirms that errors are always written to the client and not something else
+	assert.Equal(t, expectedClient, actualClient)
```
diverdane: Is it possible for the errors to be written to something else besides the client?

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -1,59 +1,372 @@
...
+
+	assert.Equal(t, actualLoginRequest, expectedPreLoginRequest)
+}
+
+func TestReadPreLoginRequestFails(t *testing.T) {
+	methodFails(t, func(connectorOptions *types.ConnectorOptions) {
+		connectorOptions.ReadPreloginRequest = func(
+			io.ReadWriteCloser,
+		) (map[uint8][]byte, error) {
+			return nil, methodFailsExpectedErr
```
diverdane: The methodFails method depends upon (or assumes that) the test that calls provides function options that must return 'methodFailsExpectedErr'. This does minimize the code, but do we want to keep this implicit dependency for methodFails, or alternatively pass the expected error to methodFails?

internal/plugin/connectors/tcp/mssql/mock/options.go
```diff
@@ -0,0 +1,103 @@
...
+// func Connect(...) (net.Conn, error)
+type MSSQLConnectorCtorOptions struct {
+	// Backend connection returned from the call to Connect
+	BackendConn net.Conn
+	// Error returned from the call to Connect
+	Err error
+	// PreloginResponse from the server passed to interceptor during Connect
+	ServerPreloginResponse map[uint8][]byte
+	// Pointer to unload client login request to, taken from interceptor during Connect
+	ClientLoginRequestPtr **mssql.LoginRequest
```
jonahx: There should be no references to `mssql` in any mocks.  This is a very important invariant to maintain.

internal/plugin/connectors/tcp/mssql/mock/options.go
```diff
@@ -0,0 +1,103 @@
...
+//
+// Empty values of MSSQLConnectorCtorOption result default:
+// ServerPreloginResponse => map[uint8][]byte{}
+// ServerLoginResponse => &mssql.LoginResponse{}
+// BackendConn => nil
+// BackendConn => nil
+// ClientLoginRequestPtr => nil (this means ClientLoginRequest is read from the channel to
+// a vacuum)
+func MSSQLConnectorCtor(setters ...MSSQLConnectorCtorOption) types.ConnectorOption {
+	args := &MSSQLConnectorCtorOptions{}
```
jonahx: Create a `New` for this which takes care of all the defaults for you, so you can remove all the `if` statements in the code below.

internal/plugin/connectors/tcp/mssql/mock/options.go
```diff
@@ -0,0 +1,103 @@
...
+	ClientLoginRequestPtr **mssql.LoginRequest
+	// LoginResponse from the server passed to interceptor during Connect
+	ServerLoginResponse *mssql.LoginResponse
+}
+
+// MSSQLConnectorCtorOption is the 'functional option' complement to
+// MSSQLConnectorCtorOptions.
+type MSSQLConnectorCtorOption func(*MSSQLConnectorCtorOptions)
+
+// MSSQLConnectorCtor is a 'functional option' of types.ConnectorOption. It generates
```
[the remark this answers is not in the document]
jonahx: I thought about this for a long time, and wrote up a whole long thing about how to make it clearer, improve names, etc, but at the end I threw it all away because I think this is a mistake.

Even being very familiar with these patterns, it took me too long to understand the code, and there are too many types to keep in your head.  It reminds of:

"All problems in computer science can be solved by another level of indirection, except too much indirection."

This is something I struggle with all the time and think about a lot.  For my money, though, the nested functional options veers clearly into unnecessary abstraction territory here.  If I were to revisit this code in 6mos, I would regret having written it that way.

And if you look at [Rob's original article](https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html) on this, the problem it's trying to solve is to make an API growable and flexible.  These conditions don't apply here.  This mock code is private and unlikely to grow. (It's even debatable if the main usage of it in the connector is justified, but that part reads clearly to me and I think I like it.)

I would make these changes:

1. Rename `mock.MSSQLConnectorCtorOptions` to `MSSQLConnectorCtor`
1. Rename what is currently `MSSQLConnectorCtor` to `MSSQLConnectorCtorOpt`.  It returns an option after all, per the first line of your comment.
2. Rewrite usages such as the following:

```go
connector := NewSingleUseConnectorWithOptions(
  mock.DefaultConnectorOptions(),
  mock.MSSQLConnectorCtor(
    func(options *mock.MSSQLConnectorCtorOptions) {
      options.BackendConn = expectedBackendConn
    },
  ),
)
```
to something like:

```go
connector := NewSingleUseConnectorWithOptions(
  mock.DefaultConnectorOptions(),
  mock.MSSQLConnectorCtorOpt( // returns a connector option
    mock.MSSQLConnectorCtor{
      BackendConn: expectedBackendConn
    }
  ),
)
```
You can fill in the details so that default values are taken care of as elegantly as they are with the functional options pattern.

Issue #1073, opened 2020-01-07
MSSQL TestThirdPartyConnect is implemented


izgeri:
- Test Cases
    - If the driverConnector.Connect never returns a prelogin response,
      Secretless fails cleanly with a timeout error which is propogated. (Come
      up with the reasonable time for this to occur... 5 seconds?  10?) 
    - Same thing but for "prelogin response suceeds, login response fails"

PR #1089, opened 2020-01-14
Introduce Functional Options for MsSQL Connector

BradleyBoutcher:
This PR is focused on improving the foundation of our current MsSQL tests, reducing code reuse and time needed to build new unit tests for the connector. It additionally includes a division between tests for our connector, and tests for our production options.
- NewSingleUseConnectorWithOptions allows custom injection of dependencies without needing to specify all of them
- ConnectorOptions struct captures all the configuration options for a SingleUseConnector
- ConnectorOption is the 'functional option' complement to ConnectorOptions
- Introduce tests that cover the 'production' options - dependencies we will use during production.
- Covers each option specified in the singleuseconnector
- Introduces mock options structure used for unit testing, including a function which returns default options
- Unit tests now make use of the t.Run structure, where each test can be ran from within a single connect function
- Introduces helper functions for testing around failure
- Makes use of the new functional options feature in the connector

This is a re-uploaded PR of @doodlesbykumbi's work. All credit should go to him.

Connected to:
#1010 
#1073 

internal/plugin/connectors/tcp/mssql/connector_test.go

NA

NA

NA

NA
NA

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connector_test.go
+++ b/internal/plugin/connectors/tcp/mssql/connector_test.go
@@ -1,59 +1,108 @@
 package mssql
 
 import (
-	"context"
+	"errors"
+	"io"
 	"net"
 	"testing"
 
+	errorspkg "github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/cyberark/secretless-broker/internal/plugin/connectors/tcp/mssql/mock"
-	logmock "github.com/cyberark/secretless-broker/pkg/secretless/log/mock"
+	"github.com/cyberark/secretless-broker/internal/plugin/connectors/tcp/mssql/types"
 	pluginconnector "github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
 	"github.com/denisenkom/go-mssqldb"
 )
 
-func TestHappyPath(t *testing.T) {
-	logger := logmock.NewLogger()
+func TestSingleUseConnector_Connect(t *testing.T) {
+	t.Run("singleUseConnector.driver#Connect success", func(t *testing.T) {
+		expectedBackendConn := mock.NewNetConn(nil)
+
+		connector := NewSingleUseConnectorWithOptions(
+			mock.DefaultConnectorOptions(),
+			mock.MSSQLConnectorCtor(
+				func(options *mock.MSSQLConnectorCtorOptions) {
+					options.BackendConn = expectedBackendConn
+				},
+			),
+		)
+		actualBackendConn, err := runDefaultTestConnect(connector)
+
+		assert.Nil(t, err)
+		if err != nil {
+			return
+		}
+
+		assert.Equal(t, expectedBackendConn, actualBackendConn)
+	})
+
+	t.Run("singleUseConnector.driver#Connect fail", func(t *testing.T) {
+		methodFails(t, mock.MSSQLConnectorCtor(
+			func(ctorOptions *mock.MSSQLConnectorCtorOptions) {
+				ctorOptions.Err = methodFailsExpectedErr
+			}),
+		)
+	})
+}
+// Test helpers
+//
+
+// runDefaultTestConnect passes in default values to the Connect method of a connector.
+// This helps avoid boilerplate
+func runDefaultTestConnect(
+	connector *SingleUseConnector,
+) (net.Conn, error) {
 	clientConn := mock.NewNetConn(nil)
-	expectedBackendConn := mock.NewNetConn(nil)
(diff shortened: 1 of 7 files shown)


internal/plugin/connectors/tcp/mssql/mock/options.go
```diff
@@ -0,0 +1,103 @@
...
+
+// MSSQLConnectorCtor is a 'functional option' of types.ConnectorOption. It generates
+// a customizable MSSQLConnectorCtor using MSSQLConnectorCtorOptions, then sets it on the
+// ConnectorOption passed to it.
+//
+// Empty values of MSSQLConnectorCtorOption result default:
+// ServerPreloginResponse => map[uint8][]byte{}
+// ServerLoginResponse => &mssql.LoginResponse{}
+// BackendConn => nil
+// BackendConn => nil
```
BradleyBoutcher: Quote @diverdane 
> Do we want to use 'options' here instead of 'setters', per Johah's previous comment, or is this used intentionally to distinguish this from ConnectorOptions?

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -1,59 +1,108 @@
...
 
-			<- interceptor.ClientLoginRequest
+// methodFailsExpectedErr is the error value used inside methodFails
+var methodFailsExpectedErr = errors.New("failed for the test")
 
-			interceptor.ServerLoginResponse <- &mssql.LoginResponse{}
+// methodFails checks that the expected error is present in:
+// 1. the error returned from the call to the Connect method
+// 2. the error propagated to the client
+func methodFails(
```
BradleyBoutcher: Quote @diverdane 
> The methodFails method depends upon (or assumes that) the test that calls provides function options that must return 'methodFailsExpectedErr'. This does minimize the code, but do we want to keep this implicit dependency for methodFails, or alternatively pass the expected error to methodFails?

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -1,59 +1,108 @@
...
+	t *testing.T,
+	connectorOption types.ConnectorOption,
```
BradleyBoutcher: Quote @diverdane 
> Just a thought... but could we use a variadic connectorOptions here instead of a single connectorOption? Not sure if this would ever come into play, but if it's cheap and easy to change

internal/plugin/connectors/tcp/mssql/production_options_test.go
```diff
@@ -0,0 +1,133 @@
...
+	actualLoginResponse, err := mssql.ReadLoginResponse(r)
+	assert.Nil(t, err)
+	if err != nil {
+		return
+	}
+
+	assert.Equal(t, expectedLoginResponse, actualLoginResponse)
+}
+
+func TestProductionWritePreLoginResponse(t *testing.T) {
```
diverdane: Am I being *really* anal to want to move the `PreLogin` test cases before the `Login` test cases, and have the `TestProductionReadPreLoginResponse` test case before the `TestProductionWritePreloginResponse`? (i.e. chronological order?)

internal/plugin/connectors/tcp/mssql/production_options_test.go
```diff
@@ -0,0 +1,133 @@
...
+	expectedLoginResponse.TDSVersion = 12
+
+	r, w := net.Pipe()
+	go func() {
+		writeLoginResponse(w, expectedLoginResponse)
+	}()
+
+	// login response returned from ReadLoginResponse
+	actualLoginResponse, err := mssql.ReadLoginResponse(r)
+	assert.Nil(t, err)
```
diverdane: Would `assert.NoError` be clearer than `assert.Nil`?

internal/plugin/connectors/tcp/mssql/connector_test.go
BradleyBoutcher: @diverdane What do you think of this modification?
```
t.Run("singleUseConnector.driver#Connect fail", func(t *testing.T) {
		var methodFailsExpectedErr = errors.New("failed to complete connection")

		methodFails(t, methodFailsExpectedErr, mock.MSSQLConnectorCtor(
			func(ctorOptions *mock.MSSQLConnectorCtorOptions) {
				ctorOptions.Err = methodFailsExpectedErr
			}),
		)
	})
```

internal/plugin/connectors/tcp/mssql/production_options_test.go
BradleyBoutcher: It's definitely a small change, but an important one 👍

PR #1090, opened 2020-01-14
1070 MsSQL WritePrelogin Unit Test Exists

BradleyBoutcher:
- WritePreloginSucceeds test
  - The production version of WritePreloginWithPacketType
succeeds when all others succeed
  - We receive the prelogin fields from the go-mssqldb driver
  - We modify the prelogin fields before sending them to the user

- WritePreloginFails test
  - Error is returned from connect
  - Error is written to client buffer
  - Error written to log

Connected to #1070


diverdane:
This looks good to go.

PR #1093, opened 2020-01-14
Add option to specify MSSQL edition in tests

izgeri:
- Add MSSQL_PID env var option to MSSQL server container (defaults to Developer)
- Removed need for Secretless host from test script - this makes it easier to
  run the test scripts, bc you only have to supply config to the `start`
  script. All it took was leaving the `test` container up in the `start`
  script and verifying the `secretless` container was up in `start`, which
  logically is reasonable anyway.
- Updated 2017-cu1 and 2019 tests to not send Secretless host to test script


If interested in trying this out, a reviewer can run `./bin/build` from the project root and from the [mssql dir](https://github.com/cyberark/secretless-broker/tree/test-mssql-editions/test/connector/tcp/mssql) run `./start -e Enterprise` to start the containers and `./test` to run the tests.

You can validate that the MSSQL container is running the Enterprise edition by running
```
$ docker exec -it {MSSQL_CONTAINER_ID} /opt/mssql-tools/bin/sqlcmd -S localhost -U SA -P 'yourStrong()Password' -Q 'SELECT @@VERSION'
                                                                                                                                                                                                                                                                                                            
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Microsoft SQL Server 2017 (RTM-CU18) (KB4527377) - 14.0.3257.3 (X64) 
	Nov 16 2019 01:14:50 
	Copyright (C) 2017 Microsoft Corporation
	Enterprise Edition (64-bit) on Linux (Ubuntu 16.04.6 LTS)   
```

I have used this to validate that our test cases pass against all server editions (Developer, Express, Standard, Enterprise, EnterpriseCore).

### DIFF
--- a/test/connector/tcp/mssql/start
+++ b/test/connector/tcp/mssql/start
@@ -1,26 +1,32 @@
 #!/bin/bash -ex
 
 mssql_host=mssql
-while getopts :dm:s: opt; do
+secretless_host=secretless
+mssql_edition=Developer # can also be Express, Standard, Enterprise, EnterpriseCore
+while getopts :dm:s:e: opt; do
     case $opt in
         d) dev_mode=true;;
         m) mssql_host=${OPTARG};;
         s) secretless_host=${OPTARG};;
+        e) mssql_edition=${OPTARG};;
        \?) echo "Unknown option -$OPTARG"; exit 1;;
     esac
 done
-# If the secretless host is not explicitly set on the command line,
-# then use one of the default names (either secretless or
-# secretless-dev, depending on whether testing is being done in
-# development mode) for the secretless host.
-if [[ -z $secretless_host ]]; then
-    secretless_host=secretless
-    if [[ "$dev_mode" = true ]]; then
-        secretless_host=secretless-dev
-    fi
+
+# the secretless host pulls from either the default value or the command line
+# input, but if dev mode is on it will be overriden by that flag to use the
+# dev container
+export SECRETLESS_HOST=$secretless_host
+
+if [[ "$dev_mode" = true ]]; then
+  echo "Using secretless-dev container"
+  export SECRETLESS_HOST=secretless-dev
 fi
 
 ./stop
+
+echo -e "\nStarting containers for the $mssql_edition edition\n"
+export MSSQL_PID=$mssql_edition
 
 docker-compose build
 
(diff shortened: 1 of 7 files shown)


test/connector/tcp/mssql/start
```diff
@@ -1,27 +1,34 @@
...
     case $opt in
         d) dev_mode=true;;
         m) mssql_host=${OPTARG};;
         s) secretless_host=${OPTARG};;
+        e) mssql_edition=${OPTARG};;
        \?) echo "Unknown option -$OPTARG"; exit 1;;
     esac
 done
+
 # If the secretless host is not explicitly set on the command line,
```
diverdane: Nit: This comment needs to be updated. The `If the secretless host is not explicitly set on the command line...` isn't checked anymore.

diverdane: Nit: Comment should be updated since the `If the secretless host is not explicitly set on the command line` no longer applies.

test/connector/tcp/mssql/test
```diff
@@ -1,55 +1,6 @@
...
-docker-compose run \
-  -e SECRETLESS_HOST="$SECRETLESS_HOST" \
-  --rm \
-  --no-deps \
+docker-compose exec -T \
   test bash -c "go test -v ./test/connector/tcp/mssql"
-
-# print secretless logs after tests run
-ret=$?
-docker-compose logs $SECRETLESS_HOST
```
diverdane: Do we no longer need to print logs from the secretless container? Might be handy to have for CI failures.

izgeri: I am actually not sure if I like this, but I will share what I was thinking and you can tell me if you think I should put it back anyway :)

Pros of adding them to script output:
- The Jenkins logs will show the container logs, and it might help when trying to understand a test failure

Cons of adding them to the script:
- Every time I run the scripts locally I get a flood of container logs in my terminal, which makes it hard to see the rest of the important info that I can't get from running `docker logs`
- When you run in Jenkins, there is a ton of info from the container logs that pollutes the log output, and even in the blue ocean view it's hard to distinguish the container log output from the rest of the test script output. It may make it harder to scan for pass/fail at a glance
- In practice I haven't seen an integration test failure that I didn't need to reproduce locally before fixing anyway, at which point I would have access to the container logs on my local machine

I do think having the logs in the Jenkins output may help with some edge case test failures, but the experience of using the scripts in the meantime is degraded enough for me that I prefer to leave this out. That being said, I'm not the one running them most of the time - so if you feel strongly this should stay, I can find a way to get something back in.

diverdane: Your bullet item in the `Pros` has me completely sold on leaving this out... and I very much agree with the "hard to distinguish the container log output from the rest of the test script output". It took me a while to get used to that stream of secretless log showing up after test results.

PR #1095, opened 2020-01-15
Postgres connector validates packet lengths

### LINKED ISSUE #1088: Postgres connector validates packet lengths
The postgres protocol implementation reads a message without validating its length. In particular, it assumes that the length is at least 4 based on [this line](https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/tcp/pg/protocol/protocol.go#L88).

Additionally, the `ParseStartupMessage` function calls `ReadString` twice in a row even if it errors the first time - we should update this to check for errors after the first call as well.

AC:
- [x] Add unit tests for `readMessage` that at least verify its behavior when the message length is valid, invalid but > 4, or invalid and < 4.
- [x] Update [`readMessage`](https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/tcp/pg/protocol/protocol.go#L80) to appropriately validate message length, and error if there's a problem with the packet
- [x] Update [`ParseStartupMessage`](https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/tcp/pg/protocol/startup.go#L26) to check for errors after the first `ReadString` as well

doodlesbykumbi:
resolves #1088

### DIFF
--- a/internal/plugin/connectors/tcp/pg/protocol/protocol_test.go
+++ b/internal/plugin/connectors/tcp/pg/protocol/protocol_test.go
@@ -1,0 +1,65 @@
+package protocol
+
+import (
+	"encoding/binary"
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReadMessage(t *testing.T) {
+	t.Run("parses contents", func(t *testing.T) {
+		r, w := net.Pipe()
+		expectedMessageType := byte(12)
+		expectedMessage := []byte{0,1,2,3,4}
+
+		go func() {
+			err := binary.Write(w, binary.BigEndian, expectedMessageType)
+			if err != nil {
+				panic(err)
+			}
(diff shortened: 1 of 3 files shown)


internal/plugin/connectors/tcp/pg/protocol/protocol_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+		if !assert.NoError(t, err) {
+			return
+		}
+
+		assert.Equal(t, expectedMessage, message)
+		assert.Equal(t, expectedMessageType, messageType)
+	})
+
+	t.Run("validates message length", func(t *testing.T) {
```
BradleyBoutcher: I may be mistaken, but it looks like we're missing a test case. It could fail to write, or it could fail because the length is wrong. Shouldn't we have two separate fail case tests?
>  Add unit tests for readMessage that at least verify its behavior when the message length is valid, invalid but > 4, or invalid and < 4

PR #1098, opened 2020-01-16
Reduce indirection of MsSQL Unit Test Framework

BradleyBoutcher:
- MSSQLConnectorCtor struct represents MSSQLConnectorCtor options
- Default constructor for an MSSQLConnectorCtor
- Wrapper function for constructing a Connector and its default methods, which can both be customized as needed

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connector_test.go
+++ b/internal/plugin/connectors/tcp/mssql/connector_test.go
@@ -17,15 +17,10 @@ import (
 
 func TestSingleUseConnector_Connect(t *testing.T) {
 	t.Run("singleUseConnector.driver#Connect success", func(t *testing.T) {
-		expectedBackendConn := mock.NewNetConn(nil)
+		expectedBackendConn := mock.DefaultMSSQLConnectorCtor.BackendConn
 
 		connector := newSingleUseConnectorWithOptions(
 			mock.DefaultConnectorOptions,
-			mock.MSSQLConnectorCtor(
-				func(options *mock.MSSQLConnectorCtorOptions) {
-					options.BackendConn = expectedBackendConn
-				},
-			),
 		)
 		actualBackendConn, err := runDefaultTestConnect(connector)
 
@@ -40,11 +35,10 @@ func TestSingleUseConnector_Connect(t *testing.T) {
 	t.Run("singleUseConnector.driver#Connect fail", func(t *testing.T) {
 		var methodFailsExpectedErr = errors.New("failed to complete connection")
 
-		methodFails(t, methodFailsExpectedErr, mock.MSSQLConnectorCtor(
-			func(ctorOptions *mock.MSSQLConnectorCtorOptions) {
-				ctorOptions.Err = methodFailsExpectedErr
-			}),
-		)
+		customCtor := mock.DefaultMSSQLConnectorCtor
+		customCtor.Err = methodFailsExpectedErr
+
+		methodFails(t, methodFailsExpectedErr, mock.CustomNewMSSQLConnectorOption(customCtor))
 	})
 
 	t.Run("singleUseConnector#ReadPreLoginRequest fail", func(t *testing.T) {
@@ -84,8 +78,12 @@ func TestSingleUseConnector_Connect(t *testing.T) {
 
 		var actualClient io.ReadWriteCloser
 
+		customCtor := mock.DefaultMSSQLConnectorCtor
+		customCtor.ServerPreloginResponse = fakePreLoginResponse
+
 		connector := newSingleUseConnectorWithOptions(
 			mock.DefaultConnectorOptions,
+			mock.CustomNewMSSQLConnectorOption(customCtor),
 			func(connectorOptions *types.ConnectorOptions) {
 				connectorOptions.WritePreloginResponse = func(
 					w io.ReadWriteCloser,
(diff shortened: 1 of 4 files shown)


internal/plugin/connectors/tcp/mssql/mock/options.go
```diff
@@ -26,76 +26,43 @@ type MSSQLConnectorCtorOptions struct {
...
-// ClientLoginRequestPtr => nil (this means ClientLoginRequest is read from the channel to
-// a vacuum)
-func MSSQLConnectorCtor(setters ...MSSQLConnectorCtorOption) types.ConnectorOption {
-	args := &MSSQLConnectorCtorOptions{}
-
-	for _, setter := range setters {
-		setter(args)
-	}
+// NewMSSQLConnectorCtor is the default constructor for MSSQLConnectorCtor
+var NewMSSQLConnectorCtor = MSSQLConnectorCtor{
```
doodlesbykumbi: I'm confused about this. I don't see it used anywhere ?

internal/plugin/connectors/tcp/mssql/mock/options.go
```diff
@@ -26,76 +26,43 @@ type MSSQLConnectorCtorOptions struct {
...
+		connectOptions.Logger = logmock.NewLogger()
+		connectOptions.NewMSSQLConnector = NewSuccessfulMSSQLConnectorCtor(
 			func(ctx context.Context) (net.Conn, error) {
 				interceptor := mssql.ConnectInterceptorFromContext(ctx)
 
-				if args.ServerPreloginResponse == nil {
-					args.ServerPreloginResponse = map[uint8][]byte{}
-				}
-				interceptor.ServerPreLoginResponse <- args.ServerPreloginResponse
+				interceptor.ServerPreLoginResponse <- ctor.ServerPreloginResponse
```
doodlesbykumbi: How come we no longer have the nil protection for `ctor.ServerPreloginResponse` and `interceptor.ServerLoginResponse` ?

internal/plugin/connectors/tcp/mssql/mock/options.go
BradleyBoutcher: Essentially this is a constructor used everywhere you made the comments above - it fills in default values so we don’t have to worry about checking for nil. That was part of Jonah’s comment.

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -181,8 +174,11 @@ func TestSingleUseConnector_Connect(t *testing.T) {
 		var actualLoginResponse *mssql.LoginResponse
 		var actualClient io.ReadWriteCloser
 
+		ctor := mock.NewMSSQLConnectorCtor
```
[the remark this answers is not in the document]
BradleyBoutcher: It’s a constructor that is filling in a MSSQLConnectorCtor with default values and returning it. I’m not familiar with Go Syntax, but can we pass values in that way as well?

internal/plugin/connectors/tcp/mssql/mock/options.go
BradleyBoutcher: It’s no longer needed, since the values there will never be nil. This method is never used for a fail case so that shouldn’t be necessary.

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -19,13 +19,12 @@ func TestSingleUseConnector_Connect(t *testing.T) {
 	t.Run("singleUseConnector.driver#Connect success", func(t *testing.T) {
 		expectedBackendConn := mock.NewNetConn(nil)
 
+		customCtor := mock.DefaultMSSQLConnectorCtor
```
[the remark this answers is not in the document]
BradleyBoutcher: Good catch - thanks!

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -272,6 +267,10 @@ func methodFails(
 	// expected error on method
 	expectedErr := methodFailsExpectedErr
 
+	// We build our connector in three parts
+	// 1. Pass in our Mock connector with default methods
+	// 2. Pass in our failing method for this test case
```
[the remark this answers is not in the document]
BradleyBoutcher: Ah yes that should be updated, once again good catch!

internal/plugin/connectors/tcp/mssql/mock/options.go
```diff
@@ -10,10 +10,10 @@ import (
 	mssql "github.com/denisenkom/go-mssqldb"
 )
 
-// MSSQLConnectorCtorOptions represents options when creating a MSSQLConnectorCtor. This
+// MSSQLConnectorCtor represents options when creating a MSSQLConnectorCtor. This
```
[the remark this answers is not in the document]
BradleyBoutcher: That's an important distinction. I added a few words to make it clear what we're talking about here.

internal/plugin/connectors/tcp/mssql/connector_test.go
```diff
@@ -21,11 +21,6 @@ func TestSingleUseConnector_Connect(t *testing.T) {
```
diverdane: In the line right above this (line 20), we have:
```
    expectedBackendConn := mock.NewNetConn(nil)
```
This is allocating a new netConn object. Later, when mock.DefaultCOnnectorOptions is called, that will allocate another new netConn object. After we run connect, it asserts that these two objects are "Equal".

Would it be better to set `expectedBackendConn` to the actual value (object) we expect?:
```
expectedBackendConn := mock.DefaultMSSQLConnectorCtor.BackendConn
```
The test works, so I'm thinking assert.Equal() only checks that 2 objects have the same content.

Issue #1085, opened 2020-01-10
Remove the ability to override an internal plugin with an external one


izgeri:
At current we allow internal [http](https://github.com/cyberark/secretless-broker/blob/master/pkg/secretless/plugin/sharedobj/available_plugins.go#L74) and [tcp](https://github.com/cyberark/secretless-broker/blob/master/pkg/secretless/plugin/sharedobj/available_plugins.go#L90) connectors to be replaced by external connector plugins.

We would like to remove this, and force all external plugins to use unique names distinct from internal plugin names.

In addition, once this change is made, as external plugins are loaded we should also verify if their name matches an existing (already loaded) HTTP or TCP connector, whether external or internal, and fail to start if there is a name overlap.

AC:
- [x] Update Secretless to no longer overwrite internal plugins with external plugins, if the external plugins have the same name as an internal plugin. External plugin names should be checked against all internal TCP and HTTP connectors.
- [x] If Secretless attempts to load an external connector plugin that has the same name as an internal TCP or HTTP connector, Secretless should panic and fail with an error that makes it clear which external connector plugin caused the problem.
- [x] If Secretless attempts to load an external connector plugin that has the same name as an already-loaded external TCP or HTTP plugin, Secretless should panic and fail with an error that makes it clear that multiple external plugins with the same name were included.
- [x] Ideally there are tests for these scenarios, though there is some debt for the unit tests on the available plugins module - see #937, #934

PR #1099, opened 2020-01-17
Remove the ability to override an internal plugin with external one

diverdane:
Currently, an external http or tcp connector plugin is allowed to replace an
internal connector plugin that has the same plugin ID.

This change removes this capability, and forces all external plugins to
use plugin IDs that are unique among all other plugin IDs.
If there is a plugin ID conflict between an external plugin and any other
plugin, then a panic occurs.

Note that there is no need to check for duplicate plugin IDs among `internal`
plugins at run time because these plugins are
hardcoded,  so that the only way that duplicate plugin IDs could occur is if the
same ID  was used twice erroneously.

The checks for uniqueness of external plugin IDs is done in 2 places:

- Right after the external plugin file content is read in and parsed. Here, each external plugin ID is compared against all other external plugins (either HTTP or TCP) that have been loaded so far.
- When available plugins are assembled. Here, each external plugin ID is compared against all internal plugin IDs (either HTTP or TCP). 

This change also adds UT for the external_plugins.go file (which was previously
non-existent), and adds some mocks for parsing go plugins.

This change also adds history to the Secretless mock logger for log calls
for errors, warns, and panics. This will allow unit tests to check that the
appropriate errors, warnings, or panics are being logged for each test
case. History is maintained as a slice of strings for each log type
(errors/warn/panic).

Addresses Issue #1085

### DIFF
--- a/pkg/secretless/plugin/sharedobj/available_plugins.go
+++ b/pkg/secretless/plugin/sharedobj/available_plugins.go
@@ -7,6 +7,8 @@ import (
 	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
 )
 
+const pluginConflictMessage = "%s plugin ID '%s' conflicts with an existing internal plugin"
+
 // CompatiblePluginAPIVersion indicates what matching API version an external plugin
 // must have so that Secretless is capable of loading it.
 var CompatiblePluginAPIVersion = "0.1.0"
@@ -39,6 +41,24 @@ func AllAvailablePlugins(
 	)
 }
 
+// checkPluginIDConflicts asserts that a given plugin ID is not used
+// by any internal HTTP or TCP plugin.
+func checkPluginIDConflicts(
+	pluginType string, // "HTTP" or "TCP"
+	pluginID string,
+	internalPlugins plugin.AvailablePlugins,
+	logger log.Logger) {
+
+	httpPlugins := internalPlugins.HTTPPlugins()
+	if _, ok := httpPlugins[pluginID]; ok {
+		logger.Panicf(pluginConflictMessage, pluginType, pluginID)
+	}
+	tcpPlugins := internalPlugins.TCPPlugins()
+	if _, ok := tcpPlugins[pluginID]; ok {
+		logger.Panicf(pluginConflictMessage, pluginType, pluginID)
+	}
+}
+
 // AllAvailablePluginsWithOptions returns the full list of internal and external
 // plugins available to the broker using explicitly-defined lookup functions.
 func AllAvailablePluginsWithOptions(
@@ -49,52 +69,42 @@ func AllAvailablePluginsWithOptions(
 	logger log.Logger,
 ) (plugin.AvailablePlugins, error) {
 
+	allHTTPPlugins := map[string]http.Plugin{}
+	allTCPPlugins := map[string]tcp.Plugin{}
+
+	// Assemble internal plugins. Plugin IDs for internal plugins are
+	// assumed to be unique because their definitions are hardcoded.
 	internalPlugins, err := InternalPlugins(internalLookupFunc)
 	if err != nil {
 		return nil, err
 	}
+	for pluginID, httpPlugin := range internalPlugins.HTTPPlugins() {
+		allHTTPPlugins[pluginID] = httpPlugin
+	}
+	for pluginID, tcpPlugin := range internalPlugins.TCPPlugins() {
+		allTCPPlugins[pluginID] = tcpPlugin
+	}
 
+	// Assemble external plugins. Check whether the plugin ID for each
+	// external plugin conflicts with any plugin IDs of internal plugins.
+	// (Checks for uniqueness among external HTTP and TCP plugins is
+	// done elsewhere, i.e. as external plugins are discovered.)
 	externalPlugins, err := externalLookupfunc(pluginDir, checksumsFile, logger)
 	if err != nil {
 		return nil, err
 	}
-
-	httpPlugins := map[string]http.Plugin{}
-
-	for name, httpPlugin := range internalPlugins.HTTPPlugins() {
-		if _, ok := httpPlugins[name]; ok {
-			// TODO: Should this ever happen?  Do we need this check?  Should it panic?
-			logger.Warnf("Internal plugin '%s' replaced by internal plugin", name)
-		}
-		httpPlugins[name] = httpPlugin
+	for pluginID, httpPlugin := range externalPlugins.HTTPPlugins() {
+		checkPluginIDConflicts("HTTP", pluginID, internalPlugins, logger)
+		allHTTPPlugins[pluginID] = httpPlugin
 	}
-
-	for name, httpPlugin := range externalPlugins.HTTPPlugins() {
-		if _, ok := httpPlugins[name]; ok {
-			logger.Warnf("Internal plugin '%s' replaced by external plugin", name)
-		}
-		httpPlugins[name] = httpPlugin
-	}
-
-	tcpPlugins := map[string]tcp.Plugin{}
-
-	for name, tcpPlugin := range internalPlugins.TCPPlugins() {
-		if _, ok := tcpPlugins[name]; ok {
-			logger.Warnf("Internal plugin '%s' replaced by internal plugin", name)
-		}
-		tcpPlugins[name] = tcpPlugin
-	}
-
-	for name, tcpPlugin := range externalPlugins.TCPPlugins() {
-		if _, ok := tcpPlugins[name]; ok {
-			logger.Warnf("Internal plugin '%s' replaced by external plugin", name)
-		}
-		tcpPlugins[name] = tcpPlugin
+	for pluginID, tcpPlugin := range externalPlugins.TCPPlugins() {
+		checkPluginIDConflicts("TCP", pluginID, internalPlugins, logger)
+		allTCPPlugins[pluginID] = tcpPlugin
 	}
 
 	return &Plugins{
-		HTTPPluginsByID: httpPlugins,
-		TCPPluginsByID:  tcpPlugins,
+		HTTPPluginsByID: allHTTPPlugins,
+		TCPPluginsByID:  allTCPPlugins,
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/secretless/plugin/sharedobj/available_plugins.go
```diff
@@ -59,35 +59,30 @@ func AllAvailablePluginsWithOptions(
...
 
+	// Assemble HTTP plugins
 	httpPlugins := map[string]http.Plugin{}
-
 	for name, httpPlugin := range internalPlugins.HTTPPlugins() {
-		if _, ok := httpPlugins[name]; ok {
-			// TODO: Should this ever happen?  Do we need this check?  Should it panic?
-			logger.Warnf("Internal plugin '%s' replaced by internal plugin", name)
-		}
+		// Since we're ranging over a map with plugin name as key,
```
izgeri: the key point is that the map is hardcoded, and we create it - the only way for duplicate names to happen is if we erroneously add the same name twice

your comment sort of gets to that, but it could be a bit more clear about how there is no possibility

diverdane: I've reworded the comment... hopefully it's clear enough, but please let me know if it can be clearer.

diverdane:
Okay, I've added checks and UT for the different combinations.
One thing that I need clarification on is re. this:

> In addition, it will show a misleading error ("external plugin conflicts with internal plugin") when
> 
> * two external HTTP plugins are loaded with the same name
> * two external TCP plugins are loaded with the same name

In the the code I'm modifying, it's not possible to detect either of these cases, since the set of available HTTP plugins and the set of available TCP plugins are maintained in a map with name as key. I *think* this is okay, since, AFAICT, the external plugins are defined by creating a subdirectory for each external plugin. If this is the case (???), then wouldn't uniqueness be guaranteed among external HTTP plugins and among external TCP plugins, since the subdirectory names would need to be unique? If this assumption is incorrect, we'd have to check for uniqueness wherever these maps are instantiated.

pkg/secretless/plugin/sharedobj/available_plugins.go
```diff
@@ -60,35 +65,44 @@ func AllAvailablePluginsWithOptions(
...
 	httpPlugins := map[string]http.Plugin{}
+	tcpPlugins := map[string]tcp.Plugin{}
 
-	for name, httpPlugin := range internalPlugins.HTTPPlugins() {
-		if _, ok := httpPlugins[name]; ok {
-			// TODO: Should this ever happen?  Do we need this check?  Should it panic?
-			logger.Warnf("Internal plugin '%s' replaced by internal plugin", name)
+	checkHTTPNameConflict := func(pluginType, pluginName string) {
+		if _, ok := httpPlugins[pluginName]; ok {
+			logger.Panicf("%s plugin name '%s' %s", pluginType, pluginName, httpNameConflict)
```
diverdane: I'd like to hear opinions on whether this panic message is precise enough for indicating whether the plugin being checked conflicts with an `internal` vs. `external` plugin. For example, the current panic message might look like this:
```
external TCP plugin name 'FooTCP' conflicts with same-named HTTP plugin
```
It doesn't specify whether the HTTP plugin it conflicts with is internal or external. I can add this preciseness in, but it would take a little bit more bookkeeping in this function to pull it off.

izgeri: this seems good enough to ship for me!

izgeri:
@diverdane I'm not quite sure I follow [this comment](https://github.com/cyberark/secretless-broker/pull/1099#issuecomment-575733974)

but I did dive through the code and found that config-to-connector matching is _primarily_ done by the connector name in the config. I wrote:

> based on [this](https://github.com/cyberark/secretless-broker/blob/master/pkg/secretless/plugin/sharedobj/available_plugins.go#L16-L23) and [this](https://github.com/cyberark/secretless-broker/blob/master/pkg/secretless/config/v2/config.go#L83-L95) it looks like we determine the connector type to create from the config by checking if there is a name match for HTTP connectors first, then checking if the configured connector is ssh or ssh-agent, and if it’s none of those creating a TCP connector
> note that it does not check if the configured connector name matches that of an available TCP connector at this point (eg it just assumes it must be TCP, since it was none of the above)

so it _is_ theoretically possible to have an HTTP connector with `name1` and a TCP connector with `name1`, and whenever you run Secretless if your config points to name `name1` it will _always_ default to the HTTP `name1` connector and never hit the TCP `name1` connector

so imo it's better to just be sure there are no duplicates for now when the external plugins are loaded. I don't think we need to check for this anywhere but here - if you think we might need to, can you add more details about that?

pkg/secretless/plugin/sharedobj/available_plugins.go
```diff
@@ -60,35 +65,39 @@ func AllAvailablePluginsWithOptions(
...
 		httpPlugins[name] = httpPlugin
 	}
-
-	tcpPlugins := map[string]tcp.Plugin{}
-
 	for name, tcpPlugin := range internalPlugins.TCPPlugins() {
-		if _, ok := tcpPlugins[name]; ok {
-			logger.Warnf("Internal plugin '%s' replaced by internal plugin", name)
-		}
+		checkNameConflicts("internal TCP", name)
```
izgeri: it doesn't hurt to check this, but I think the same comment as above applies
> 		// Since internal plugins are hard coded, the only way that duplicate
>		// internal plugin names could occur is if the same name was
>		// erroneously used twice. So no checking for duplicate internal
>		// plugin names is done here at run time.

pkg/secretless/plugin/sharedobj/available_plugins_test.go
```diff
@@ -74,63 +75,144 @@ func TestAllAvailablePlugins(t *testing.T) {
...
+			addPluginName:     "intHTTP1",
+			expectedPanic:     httpNameConflict,
+		},
+		{
+			description:       "An external TCP plugin and an internal TCP plugin is loaded with the same name",
+			addPluginType:     "external",
+			addPluginProtocol: "TCP",
+			addPluginName:     "intTCP1",
+			expectedPanic:     tcpNameConflict,
+		},
```
izgeri: I think this is still missing
- two external HTTP
- two external TCP
from the test cases

pkg/secretless/plugin/sharedobj/available_plugins.go
diverdane: Yes, agreed. I removed the uniqueness check, and changed and moved the comment since it applies to both internal HTTP and internal TCP.

izgeri:
just wanted to write here to note that we realized today that the place to actually test for this is when the external tcp/http plugin maps are built, eg in [this switch](https://github.com/cyberark/secretless-broker/blob/master/pkg/secretless/plugin/sharedobj/external_plugins.go#L169)

stay tuned ...

pkg/secretless/plugin/sharedobj/available_plugins_test.go
diverdane: Per our phone conversation, I'll move the uniqueness checks to happen earlier in the plugin discovery process, and retrofit/move the UT to apply to the place where those uniqueness checks land. Thanks for walking through this!

diverdane:
Update has been pushed. This includes checks for external plugins having unique plugin IDs as soon as they are parsed, and another check that the external plugins' plugin IDs don't conflict with internal plugins. This adds a bunch of UT and mocks for external_plugins.go. Previously, there was no UT for this... now coverage is at 42%.

diverdane:
@izgeri I don't think I can remove the `Changes Requested` status, so I sent another review request. No
 changes have been made since the last time you reviewed this.

pkg/secretless/plugin/sharedobj/mock/plugins.go
```diff
@@ -121,3 +124,72 @@ func GetExternalPlugins(
...
+	}
+	return nil, fmt.Errorf("unknown symbolName %s", symbolName)
+}
+
+// RawPlugins instantiates a map of RawPlugins for testing. Values of entries
+// in this map should reflect values returned by HTTPExternalPluginsByID()
+// and TCPExternalPluginsByID() above.
+var RawPlugins = map[string]RawPlugin{
+	"http1": RawPlugin{
+		PluginAPIVersion: "0.1.0",
```
sgnn7: This string can be a constant since it's repeated a bunch

pkg/secretless/plugin/sharedobj/external_plugins_test.go
```diff
@@ -0,0 +1,155 @@
...
+			return
+		}
+
+		expExternalPlugins := mock.ExternalPlugins()
+		assert.EqualValues(t, expExternalPlugins.HTTPPlugins(), externalPlugins.HTTPPlugins())
+		assert.EqualValues(t, expExternalPlugins.TCPPlugins(), externalPlugins.TCPPlugins())
+	})
+
+	// Test detection of conflicting external plugin names.
+	extHTTPOrTCPIDConflict := "conflicts with external"
```
sgnn7: This variable name is _really_ hard to read due to casing

pkg/secretless/plugin/sharedobj/external_plugins.go
```diff
@@ -130,15 +135,39 @@ func ExternalPlugins(
...
+	// Check whether this ID is a duplicate of a loaded external TCP plugin.
+	if _, ok := externalPlugins.TCPPluginsByID[pluginID]; ok {
+		logger.Panicf("%s plugin name '%s' %s", pluginType, pluginID, externalTCPIDConflict)
+	}
+}
+
+// rawPlugin defines an interface that includes the only method of a Go plugin
+// that we care about. This provides an abstraction layer that we can use
+// to mock Go plugins.
+type rawPlugin interface {
```
sgnn7: Type def should probably be at the top of the file

pkg/secretless/plugin/sharedobj/external_plugins.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector/tcp"
 )
 
+const (
+	externalHTTPIDConflict = "conflicts with external HTTP plugin"
+	externalTCPIDConflict  = "conflicts with external TCP plugin"
+)
```
sgnn7: Ditto on use of these constants as a pattern

pkg/secretless/plugin/sharedobj/mock/plugins.go
diverdane: Fixed.

PR #1105, opened 2020-01-24
fixes mssql poc

doodlesbykumbi:
Tries to address https://github.com/cyberark/secretless-broker/issues/1106 and https://github.com/cyberark/secretless-broker/issues/1107.

The faithful propagation of MSSQL packets:
See the parts of the code with `CALLOUT:`
1. Login Request from client to server (tds.go)
1. Login Response from server to client (early returns in connector.go and tds.go) 

In trying to figure out what was wrong I also tried other things so you'll also see code for:
1. Secure auth
1. Passing client prelogin request to the server

***Definitely not expecting Jenkins to pass but the container image builds fine :)

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connector.go
+++ b/internal/plugin/connectors/tcp/mssql/connector.go
@@ -100,6 +100,7 @@ Overview of the connection process
 type SingleUseConnector struct {
 	clientConn net.Conn
 	clientBuff io.ReadWriteCloser
+	preloginRequest map[uint8][]byte
 
 	types.ConnectorOptions
 }
@@ -171,6 +172,7 @@ func (connector *SingleUseConnector) Connect(
 	clientConn net.Conn,
 	credentialValuesByID connector.CredentialValuesByID,
 ) (net.Conn, error) {
+	var err error
 
 	connector.clientConn = clientConn
 
@@ -181,7 +183,8 @@ func (connector *SingleUseConnector) Connect(
 	// client would block forever waiting for its pre-login handshake to be
 	// read.
 	connector.clientBuff = connector.NewTdsBuffer(connector.clientConn)
-	_, err := connector.ReadPreloginRequest(connector.clientBuff)
+
+	connector.preloginRequest, err = connector.ReadPreloginRequest(connector.clientBuff)
 	if err != nil {
 		wrappedError := errors.Wrap(err, "failed to read prelogin request from client")
 		connector.writeErrorToClient(wrappedError)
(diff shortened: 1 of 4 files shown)


internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -285,7 +297,19 @@ func (connector *SingleUseConnector) waitForServerConnection(
 		// Since the communication between the client and Secretless must be unencrypted,
 		// we fool the client into thinking that it's talking to a server that does not
 		// support encryption.
```
BradleyBoutcher: I believe this logic is already performed in the driver, is it not? That’s why we originally overrode the encryption here.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -229,6 +232,11 @@ func (connector *SingleUseConnector) Connect(
 		connInterceptor,
 		connectionResultChan)
 
+	// CALLOUT: this was one of the sources of the jdbc issues
+	// NOTE: this means that we pass on the client login request and call it a day
```
diverdane: This is cool! I think to match this change, the sequence diagram above needs to change (e.g. remove the "Intercept login response", and have the login response being sent from server all the way to client. I don't know if this is possible, but it might help to have a note box saying that we've entered a duplex pipe mode/phase for the duration of the connection.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -229,6 +232,11 @@ func (connector *SingleUseConnector) Connect(
...
+	// the rest of the auth negotiation takes place within the duplex pipe
+	return backendConn, err
```
diverdane: I think this return needs to be moved after the check for error being returned as part of the connection attempt.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -229,6 +232,11 @@ func (connector *SingleUseConnector) Connect(
...
+
 	if err != nil {
 		connector.writeErrorToClient(err)
 		return nil, err
```
diverdane: The call to connector.WriteLoginResponse on line 245 needs to be deleted, I believe?

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -285,7 +297,19 @@ func (connector *SingleUseConnector) waitForServerConnection(
 		// Since the communication between the client and Secretless must be unencrypted,
 		// we fool the client into thinking that it's talking to a server that does not
 		// support encryption.
-		preloginResponse[mssql.PreloginENCRYPTION] = []byte{mssql.EncryptNotSup}
+
+		// Conditionally carry out login encryption based on what the client says
```
diverdane: This comment says `Conditionally carry out login encryption`, but it seems the login encryption always happens (considering the variable name `happyWithOnlyLoginEncryption`, and the call to ServerUpgradeToTLSForLogin() when the client selects encryption off.

Issue #990, opened 2019-11-08
Convert integration tests to use configs.v2


sgnn7:
Currently our integration tests for PG and MySQL create generated configuration files for secretless that are designed for `config.v1` which has in at least one case caused unexpected problems in the product. These config generators need to be converted to `config.v2`.

AC:
- [ ] Broker's PG integration tests use `config.v2` configs
- [ ] Broker's MySQL integration tests use `config.v2` configs

vladdoster:
Hi @sgnn7, is this still open for a first time contributor? I enjoyed a [CyberArk article on this piece of software](https://www.conjur.org/blog/open-source-how-to-eat-your-own-dog-food/) and thought I'd contribute.

sgnn7:
@vladdoster Absolutely and we would love to have more contributors to the codebase! Just make sure to read through the [contribution](https://github.com/cyberark/secretless-broker/blob/master/CONTRIBUTING.md) doc first. There's a contributor agreement that you can just send us which is somewhat of a pain but you only need to do this one.

vladdoster:
Hi @sgnn7,

You'll have to bear with me as I am new to Go. I just have a few questions before I proceed further. 

So I was able to track down where the MySql and PG created there configs. It seemed to be in generate_secretless_yml.go. In my [branch](https://github.com/vladdoster/secretless-broker/blob/2968ad5451c0b3e183658e84642fc1fc2193803a/test/util/testutil/cmd/generate_secretless_yml.go#L14) I have added code which seems to take the generated v1 config and converts it to the v2 structure. I have commented it out because it was throwing an error. However, when I run the ./bin/test with my v2 conversion solution, it seems to write the correct structure. See attached screenshot.

![pic-selected-191113-2233-27](https://user-images.githubusercontent.com/10052309/68825554-7951f080-0668-11ea-93bc-5268a20aa432.png)

However it seems to error in MySQL test. Do I need to update the unit tests to take the updated config or is the way I tried converting the v1 config way off? Would love to work through this and fix the issue :smile:. 

btw - The shell scripts you all have written are top notch :+1:

sgnn7:
Hey @vladdoster!

> You'll have to bear with me as I am new to Go. I just have a few questions before I proceed further

Don't worry about asking questions and you're on the right track!

[...]

PR #1120, opened 2020-01-31
Convert integration tests to use configs.v2

doodlesbykumbi:
Resolves https://github.com/cyberark/secretless-broker/issues/990

### DIFF
--- a/pkg/secretless/config/v2/config.go
+++ b/pkg/secretless/config/v2/config.go
@@ -5,9 +5,10 @@ import (
 	"log"
 	"sort"
 
+	"gopkg.in/yaml.v2"
+
 	"github.com/cyberark/secretless-broker/pkg/secretless/plugin"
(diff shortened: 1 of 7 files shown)


pkg/secretless/config/v2/config.go
```diff
@@ -26,6 +27,38 @@ func (c Config) String() string {
 	return string(out)
 }
 
+// SecretlessYAML serializes Config to the secretless.yml format
+func (c Config) SecretlessYAML() ([]byte, error) {
```
sgnn7: This really should be `Marshal` to align with other golang constructs (https://godoc.org/gopkg.in/yaml.v2#Marshal). I kind of wonder if we can even just implement https://godoc.org/gopkg.in/yaml.v2#Marshaler and use generic YAML marshalling.

Issue #1125, opened 2020-02-05
Adds Juxtaposer persistent log and backdoor log read container


diverdane:
## Summary
NOTE: This issue (and its associated PR #1126) is being closed, since there are better ways to fix the logging disk space exhaustion and juxtaposer crashing:

- For long runs of Juxtaposer (e.g. 4-day runs), uncomment out the silent mode for the backend tester configurations in order to minimize the amount of logging being done in the Juxtaposer container.

- Add SIGPIPE to the list of signals being handled by the Secretless Juxtaposer tester to that the Juxtaposer container does not exit e.g. when there is a TCP socket read error that might happen if the Secretless sidecar container crashes. This will hopefully be enough to keep the Juxtaposer container running/resident so that we can exec into the Juxtaposer container and examine logs.

- Use a fix in the Conjur Follower infrastructure setup that uses a emptyDir volume to collect follower logs rather than using standard Docker logging. The emptyDir uses disk space directly on the host node, so this should provide more space for logging. (https://github.com/cyberark/kubernetes-conjur-deploy/pull/100)

======================================================
ORIGINAL DESCRIPTION:
--------------------------

While running Juxtaposer 4-day longevity tests with PostGres on OpenShift, there happened to be a failure in the Conjur follower that caused the both containers in the Juxtaposer pod (juxtaposer-pg and secretless containers) to crash and go into crash loop backoff.

There were two issues I had in this situation:
(1) Because the juxtaposer was in crash loop backoff, I couldn't kubectl exec into the container.
(2) I hadn't been running the test with the option to store logs in `/tmp/output.txt`, but even
      if I had, the `/tmp` directory wasn't being backed by an emptyDir for persistence across
      container failures (i.e. persistence throughout the lifetime of the pod).

Proposed fix:
* For (1), add a 'backdoor' container running Alpine that can always be kubectl exec'd into.
* For (2), add a 'persistent-log' emptyDir volume to which the juxtaposer container can optionally
   write.

## Actual Results (including error logs, if applicable)
A clear and concise description of what actually did happen.

diverdane:
Closing this issue for reasons noted above at the top of the description.

PR #1126, opened 2020-02-05
Adds juxtaposer persistent log and backdoor log read container

diverdane:
This change adds an emptyDir volume to the Juxtaposer container
so that Juxtaposer output can be concatenated for the lifetime
of the Juxtaposer pod.

This change also adds a "backdoor" Alpine container to the
Juxtaposer deployment so that the persistent log in the emptyDir
log volume can be read even when the other two containers
(juxtaposer and secretless) may be unavailable for exec'ing
e.g. when they are in (exponential) crash loop backoff.

Fixes Issue #1125

### DIFF
--- a/bin/juxtaposer/deploy/juxtaposer_deployment_template.yml
+++ b/bin/juxtaposer/deploy/juxtaposer_deployment_template.yml
@@ -36,6 +36,8 @@ spec:
         - name: ${APP_NAME}-sockets
           emptyDir:
             medium: Memory
+        - name: persistent-log
+          emptyDir: {}
 
       containers:
       - name: ${APP_NAME}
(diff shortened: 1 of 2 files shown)


bin/juxtaposer/deploy/juxtaposer_deployment_template.yml
```diff
@@ -53,6 +55,9 @@ spec:
           - mountPath: /sock
             name: ${APP_NAME}-sockets
             readOnly: false
+          - mountPath: /persistent-log
+            name: persistent-log
+            readOnly: false
```
BradleyBoutcher: Per our conversation on slack, it may be safer to set this to `readonly: true`

diverdane:
Closing this PR for now, since it *shouldn't be needed* if we do this instead:

- For long runs of Juxtaposer (e.g. 4-day runs), uncomment out the `silent` mode for the backend tester configurations in order to minimize the amount of logging being done in the Juxtaposer container.
- Add SIGPIPE to the list of signals being handled by the Secretless Juxtaposer tester to that the Juxtaposer container does not exit e.g. when there is a TCP socket read error that might happen if the Secretless sidecar container crashes. This will hopefully be enough to keep the Juxtaposer container running/resident so that we can exec into the Juxtaposer container and examine logs.
- Use a fix in the Conjur Follower infrastructure setup that uses a emptyDir volume to collect follower logs rather than using standard Docker logging. The emptyDir uses disk space directly on the host node, so this should provide more space for logging.

PR #1130, opened 2020-02-05
Add JDBC Integration tests for Postgres

BradleyBoutcher:
Utilizes the new JDBC Jar to test queries with postgres

- Runs several connection tests over TCP
- Created a new TestRunner that conforms to JDBC requirements

### DIFF
--- a/test/connector/tcp/pg/pkg/run_test_case.go
+++ b/test/connector/tcp/pg/pkg/run_test_case.go
@@ -10,6 +10,8 @@ import (
 	"github.com/smartystreets/goconvey/convey"
 )
 
+const jdbcJARPath = "/secretless/test/util/jdbc/jdbc.jar"
+
 // RunQuery runs a simply test query for the given client configuration and port.
 func RunQuery(
 	clientConfig testutil.ClientConfiguration,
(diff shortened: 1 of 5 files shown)


test/connector/tcp/pg/pkg/run_test_case.go
```diff
@@ -66,3 +66,43 @@ func RunQuery(
 
 	return string(cmdOut), err
 }
+
+// RunQuery runs a simply test query for the given client configuration and port.
+func RunJDBCQuery(
+	clientConfig testutil.ClientConfiguration,
+	connectPort testutil.ConnectionPort,
+) (string, error) {
+	const jdbcJARPath = "/secretless/test/util/jdbc/jdbc.jar"
```
sgnn7: This should be pulled out to the top of the file probably since it's a constant

test/connector/tcp/pg/pkg/run_test_case.go
```diff
@@ -66,3 +66,43 @@ func RunQuery(
...
+		"-P", clientConfig.Password,
+		"select count(*) from test.test",
+	}
+
+	// Pre command logs
+	convey.Println("")
+	convey.Println("---<< EXECUTED")
+
+	cmdOut, err := exec.Command("java", args...).CombinedOutput()
+	fmt.Println(args)
```
sgnn7: I would think that we want to show the args before things get executed otherwise if there a problem this line won't get reached.

test/connector/tcp/pg/tests/essentials_test.go
```diff
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+	"github.com/cyberark/secretless-broker/test/connector/tcp/pg/pkg"
```
sgnn7: Golang imports we have been generally splitting into sections:
```
<stdlib imports>

<third party imports>

<in-repo imports>
```

PR #1138, opened 2020-02-10
Add pipeline step to validate changelog

doodlesbykumbi:
1. Update CHANGELOG to pass validation
2. Update bin/prefill_changelog to generate valid CHANGELOG
3. Add pipeline step to validate changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -438,8 +422,6 @@ external plugins
 
 ## [0.1.0] - 2018-05-15
 
-The first tagged version.
```
izgeri: This should probably have some message - maybe it just needs `### Added` as the change type?

doodlesbykumbi:
@izgeri I agree with using bash-lib. I think that can be a follow up PR

PR #1141, opened 2020-02-11
Add openshift registry support for secretless-broker

BradleyBoutcher:
- Added an additional stage for building an openshift container
compliant image for secretless while making use of the build stage
- Added a stage to publishing for tagging and submitting the latest
version of the redhat compliant image
- Updated our build script to use stage targets, and since both
stages require different parameters, there would be a failure if
these target parameters would be left out, rather than overriding
an image with a different name

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -30,6 +30,7 @@ echo "Building secretless-broker:$FULL_VERSION_TAG Docker image"
 # shellcheck disable=SC2086
 docker build --tag "secretless-broker:${FULL_VERSION_TAG}" \
              --tag "secretless-broker:latest" \
+             --target "secretless-broker" \
              $DOCKER_FLAGS \
              --file "$TOPLEVEL_DIR/Dockerfile" \
              "$TOPLEVEL_DIR"
@@ -53,3 +54,13 @@ docker build --tag "secretless-broker-quickstart:${FULL_VERSION_TAG}" \
              $DOCKER_FLAGS \
              --file "$QUICK_START_DIR/Dockerfile" \
              "$QUICK_START_DIR"
+
+echo "Building secretless-broker-redhat:$FULL_VERSION_TAG Docker image"
+# (we want the flags to be word split here)
+# shellcheck disable=SC2086
+docker build --tag "secretless-broker-redhat:${FULL_VERSION_TAG}" \
+             --target "secretless-broker-redhat" \
+             --build-arg VERSION="${FULL_VERSION_TAG}" \
+             $DOCKER_FLAGS \
+              --file "$TOPLEVEL_DIR/Dockerfile" \
+             "$TOPLEVEL_DIR"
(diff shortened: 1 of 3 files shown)


bin/publish
```diff
@@ -42,5 +42,17 @@ for image_name in "${IMAGES[@]}"; do
       docker tag "$image_name:$FULL_VERSION_TAG" "$REGISTRY/$image_name:$tag"
       docker push "$REGISTRY/$image_name:$tag"
     done
+
+    for tag in "${TAGS[@]}"; do
+      # Publish only latest to Redhat Registries
```
izgeri: there is an image with a literal `latest` tag, so this comment could be clearer (I don't think you mean that we publish only the `latest` tag to redhat, since you're looping over the `TAGS` array)

izgeri: fwiw, I think it's fine if redhat only has the VERSION tag (and not the `latest`) - I think that's what we do for the authn-k8s client now, and it might be all we can do bc I'm pretty sure RH doesn't let us overwrite tags

BradleyBoutcher: Ok that definitely simplifies things! I'll just push the latest version, but not do the whole separate thing for the latest tag. I think openshift actually handles that on its own.

bin/build
```diff
@@ -53,3 +54,15 @@ docker build --tag "secretless-broker-quickstart:${FULL_VERSION_TAG}" \
              $DOCKER_FLAGS \
              --file "$QUICK_START_DIR/Dockerfile" \
              "$QUICK_START_DIR"
+
+echo "Building secretless-broker:4dfeef3-redhat Docker image"
+# NOTE: the latest tag is required by downstream pipeline stages
+# (we want the flags to be word split here)
+# shellcheck disable=SC2086
+docker build --tag "secretless-broker-redhat:${FULL_VERSION_TAG}" \
+             --tag "secretless-broker-redhat:latest" \
```
izgeri: Can this tag and the comment above be removed?

bin/publish
```diff
@@ -42,5 +42,15 @@ for image_name in "${IMAGES[@]}"; do
...
+
+    # Publish only latest to Redhat Registries
+    echo "Tagging and pushing ${REDHAT_IMAGE}"
+
+    docker tag "secretless-broker-redhat:${FULL_VERSION_TAG}" "${REDHAT_IMAGE}:${VERSION}"
+    # you can't push the same tag twice to redhat registry, so ignore errors
+    if ! docker push "${REDHAT_IMAGE}:${VERSION}"; then
+      echo 'RedHat push FAILED! (Maybe the image was pushed already?)'
+      exit 0
+    fi
```
izgeri: this section is going to get hit twice, once for image_name secretless-broker and once for image_name secretless-broker-quickstart

looking at it now I think this part needs to be broken out of the image loop and put in its own `if [ "$git_description" = "v${VERSION}" ]; then` loop

maybe there is a better way to organize it, but it doesn't seem right for it to get hit twice for no reason

BradleyBoutcher: good catch - fixed!

PR #1146, opened 2020-02-18
fix:conjur: set POSTGRES_HOST_AUTH_METHOD=trust 

doodlesbykumbi:
This is a new thing in `postgres:9.4` to allow all connections allow all connections without a password.

### DIFF
--- a/test/connector/http/conjur/docker-compose.yml
+++ b/test/connector/http/conjur/docker-compose.yml
@@ -2,6 +2,8 @@ version: '2'
 services:
   pg:
     image: postgres:9.4
+    environment:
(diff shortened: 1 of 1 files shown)


izgeri:
how does this not affect our pg tests? are they pinned at specific versions, so they haven't been upgraded to the point of it causing an issue yet?

maybe we need a card for bumping those at some point too?

doodlesbykumbi:
@izgeri yup, it's pinned to 9:3. i'm not sure what the motivation to upgrade is. this test could as well have been changed to use 9:3 instead of 9:4. but if the motivation becomes clear then we can do the bump.

Issue #1145, opened 2020-02-13
Secretless RH certified publish pipeline logs into RH first


izgeri:
AC:
- [x] Update the `bin/publish` script to log in to the RH container registry before calling the `push` command.

You can follow the example pull request opened at cyberark/conjur-authn-k8s-client#83 that has the changes for `conjur-authn-k8s-client`.  For `secretless-broker`, the key to use in secrets.yml is at `redhat/projects/secretless-broker/api-key`.

JakeQuilty:
Link to PR: https://github.com/cyberark/secretless-broker/pull/1149

PR #1149, opened 2020-02-20
Log into RH container registry before pushing

JakeQuilty:
Logs into RH container registry before pushing.

Connected to https://github.com/cyberark/secretless-broker/issues/1145

### DIFF
--- a/secrets.yml
+++ b/secrets.yml
@@ -1,3 +1,6 @@
+common:
+  REDHAT_API_KEY: !var redhat/projects/secretless-broker/api-key
+
 staging:
   S3_BUCKET: staging-secretless-io
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -123,7 +123,7 @@ pipeline {
       }
 
       steps {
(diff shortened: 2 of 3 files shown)


JakeQuilty:
This won't be able to be fully tested until a live attempt to push to RedHat is made

secrets.yml
```diff
@@ -9,3 +9,5 @@ production:
 
   AWS_ACCESS_KEY_ID: !var prod/aws/iam/users/secretless.io-publisher-prod/access_key_id
   AWS_SECRET_ACCESS_KEY: !var prod/aws/iam/users/secretless.io-publisher-prod/secret_access_key
+
+REDHAT_API_KEY: !var redhat/projects/secretless-broker/api-key
```
izgeri: can this be moved to the top in a `common` section, like we do [here](https://github.com/conjurdemos/kubernetes-conjur-demo/blob/master/ci/secrets.yml)? otherwise I think it's possible this might not work

JakeQuilty: Done

PR #1150, opened 2020-02-24
Bump version to v1.5.2

izgeri:
Bumps the version to v1.5.2

- [x] Manually tested [Keychain provider](https://github.com/cyberark/secretless-broker/tree/master/test/providers/keychain)
- [x] Manually run [Kubernetes-Conjur demo](https://github.com/conjurdemos/kubernetes-conjur-demo) with a local Secretless Broker image build of your branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [1.5.2] - 2020-02-24
+
+### Changed
+- Bump authn-k8s client to v0.16.1 (cyberark/conjur-authn-k8s-client#70)
+
+### Fixed
+- Updated RH image push to ensure we're logged into the RH container registry
+  appropriately before pushing (#1149)
+
 ## [1.5.1] - 2020-02-12
 
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 ## [Unreleased]
 
+## [1.5.2] - 2020-02-24
+
+### Changed
+- Bump authn-k8s client to v0.16.1 (cyberark/conjur-authn-k8s-client#70)
+
+### Fixed
+- Log into RH container registry before pushing (#1149)
+- Updated Conjur integration tests to set `POSTGRES_HOST_AUTH_METHOD=trust` to
```
sgnn7: This change is not worthy of an inclusion in the changelog tbh - it's an internal detail that shouldn't be externally visible I think.

CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [1.5.2] - 2020-02-24
+
+### Changed
+- Bump authn-k8s client to v0.16.1 (cyberark/conjur-authn-k8s-client#70)
+
+### Fixed
+- Log into RH container registry before pushing (#1149)
```
sgnn7: This probably should be changed to indicate that this fixes the push rather than implying that we fixed the logging in

CHANGELOG.md
izgeri: it is a change that was made in a PR in this version - I agree it's minor, and not relevant to end-users, but this is a log of all the changes.

I'm not sure whether we want to flag items like this (by prefacing them with `[test]` or something?) to preclude them from release notes, but I can see why we'd include them in the (contributor-facing) change logs

sgnn7: I don't believe that changelogs are intended to be a list of all changes but only the notable ones that may impact the users. Even the [keepachangelog defines](https://keepachangelog.com/en/1.1.0/) it as:
```markdown
### What is a changelog?

A changelog is a file which contains a curated, chronologically ordered
list of notable changes for each version of a project.
```

If I was a user of this repo, this change would not impact me unless I was a developer and as such is probably below the threshold of a curated + notable change.

PR #1155, opened 2020-03-06
Update README to use authenticateURLsMatching

izgeri:
The generic HTTP connector README has some sample configs for HTTP
connectors that don't work because they use the v1 config `match` syntax
instead of the v2 config `authenticateURLsMatching` syntax

See the original v2 config design document for more info:
https://github.com/cyberark/secretless-broker/blob/master/design/simpler-configuration.md


sgnn7:
@izgeri lol needs a rebase after I merged the other PR

izgeri:
rebased.

PR #1181, opened 2020-04-03
Update .codeclimate.yml

BradleyBoutcher:
- Increase method line cap
- Increase return statement cap


sgnn7:
CC: @andytinkham - we probably need this in the main template and other golang repos

PR #1183, opened 2020-04-03
Update .codeclimate.yml

BradleyBoutcher:
- Increase method line cap
- Increase return statement cap


izgeri:
I'm not sure how to validate the formatting, but fwiw this is ok with me

PR #1185, opened 2020-04-06
Add support for additional ssl options

BradleyBoutcher:
- Parse additional ssloptions if sslmode is enabled

- Check for option existence based
on length rather than being non-nil, following
the pattern set by our other connectors

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connection_details.go
+++ b/internal/plugin/connectors/tcp/mssql/connection_details.go
@@ -1,18 +1,24 @@
 package mssql
 
 import (
-	"bytes"
 	"strconv"
 )
 
 // ConnectionDetails stores the connection info to the real backend database.
 // These values are pulled from the SingleUseConnector credentials config
 type ConnectionDetails struct {
-	Host     string
-	Port     uint
-	Username string
-	Password string
-	SSLMode  string
+	Host       string
+	Port       uint
+	Username   string
+	Password   string
+	SSLMode    string
+	SSLOptions map[string]string
+}
+
+var sslOptions = []string{
+	"sslrootcert",
+	"sslkey",
+	"sslcert",
 }
 
 const defaultMSSQLPort = uint(1433)
@@ -23,36 +29,35 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
 
 	connDetails := &ConnectionDetails{}
 
-	if host := credentials["host"]; host != nil {
+	if len(credentials["host"]) > 0 {
 		connDetails.Host = string(credentials["host"])
 	}
 
 	connDetails.Port = defaultMSSQLPort
-	if credentials["port"] != nil {
+	if len(credentials["port"]) > 0 {
 		port64, _ := strconv.ParseUint(string(credentials["port"]), 10, 64)
 		connDetails.Port = uint(port64)
 	}
 
-	if credentials["username"] != nil {
+	if len(credentials["username"]) > 0 {
 		connDetails.Username = string(credentials["username"])
 	}
 
-	if credentials["password"] != nil {
+	if len(credentials["password"]) > 0 {
 		connDetails.Password = string(credentials["password"])
 	}
 
-	// More supported cases to be added
-	sslMode := credentials["sslmode"]
-	switch {
-	case bytes.Equal(sslMode, []byte("disable")):
-		connDetails.SSLMode = string(credentials["sslmode"])
-	case bytes.Equal(sslMode, []byte("enable")):
-		fallthrough
-	case bytes.Equal(sslMode, []byte("verify-ca")):
-		fallthrough
-	default:
-		connDetails.SSLMode = "disable"
+	sslMode := string(credentials["sslmode"])
+	if sslMode != "disable" {
+		// Currently, we only support disable
+		sslMode = "disable"
+		// In the event that the user does not choose 'disable', i.e. when
+		// ssl is 'required', we want to parse the additional ssl credentials
+		// that are are needed
+		connDetails.SSLOptions = newSSLOptions(credentials)
 	}
+
+	connDetails.SSLMode = sslMode
 
 	return connDetails
 }
@@ -62,4 +67,17 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
 // connect to the database -- eg, in cmd line tools.
 func (cd *ConnectionDetails) Address() string {
 	return cd.Host + ":" + strconv.FormatUint(uint64(cd.Port), 10)
+}
+
+func newSSLOptions(credentials map[string][]byte) map[string]string {
+	SSLOptions := make(map[string]string)
+
+	for _, sslOption := range sslOptions {
+		value := string(credentials[sslOption])
+		if len(value) > 0 {
+			SSLOptions[sslOption] = value
+		}
+	}
+
+	return SSLOptions
 }
--- a/internal/plugin/connectors/tcp/mssql/connection_details_test.go
+++ b/internal/plugin/connectors/tcp/mssql/connection_details_test.go
@@ -5,29 +5,39 @@ import (
 	"testing"
 )
 
+type args struct {
+	credentials map[string][]byte
+}
+
+var defaultConnectionDetails = &ConnectionDetails{
+	Username: "herp",
+	Password: "derp",
+	Host:     "0.0.0.0",
+	Port:     1234,
+	SSLMode:  "disable",
+	SSLOptions: map[string]string{
+		"sslrootcert": "foo",
+		"sslkey":      "bar",
+		"sslcert":     "foobar",
+	},
+}
+
+var emptyConnectionDetails = &ConnectionDetails{
+	Port:       defaultMSSQLPort,
+	SSLMode:    "disable",
+	SSLOptions: map[string]string{},
+}
+
 func TestConnectionDetails_Address(t *testing.T) {
-	type fields struct {
-		Host     string
-		Port     uint
-		Username string
-		Password string
-		SSLMode  string
-	}
 	tests := []struct {
 		description string
-		fields      fields
+		fields      *ConnectionDetails
 		expected    string
 	}{
 		{
 			description: "default address format",
-			fields: fields{
-				Host:     "hostname",
-				Port:     5555,
-				Username: "foo",
-				Password: "bar",
-				SSLMode:  "xyz",
-			},
-			expected: "hostname:5555",
+			fields:      defaultConnectionDetails,
+			expected:    "0.0.0.0:1234",
 		},
 	}
 	for _, tt := range tests {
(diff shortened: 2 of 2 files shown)


internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -46,12 +54,15 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
 	switch {
 	case bytes.Equal(sslMode, []byte("disable")):
 		connDetails.SSLMode = string(credentials["sslmode"])
+		break
 	case bytes.Equal(sslMode, []byte("enable")):
+		connDetails.parseSSLOptions(credentials)
```
doodlesbykumbi, line 59: I think we can refactor this so that the `parseSSLOptions` is separated out and the switch can be concerned wholly with setting SSLMode. This can be done by moving the invocation of `parseSSLOptions` to after the switch:

```go
if connDetails.SSLMode != "disable" {
	connDetails.parseSSLOptions(credentials)
}
```

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -8,36 +8,44 @@ import (
...
+const defaultSSLMode = "disable"
 
 // NewConnectionDetails is a constructor of ConnectionDetails structure from a
 // map of credentials.
 func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
 
 	connDetails := &ConnectionDetails{}
 
-	if host := credentials["host"]; host != nil {
+	if len(credentials["host"]) > 0 {
```
sgnn7, line 34: This new type of check (`len`) you have in here will panic if "host" is not defined in the config file I think

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -46,12 +54,15 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
 	switch {
 	case bytes.Equal(sslMode, []byte("disable")):
 		connDetails.SSLMode = string(credentials["sslmode"])
+		break
 	case bytes.Equal(sslMode, []byte("enable")):
+		connDetails.parseSSLOptions(credentials)
 		fallthrough
 	case bytes.Equal(sslMode, []byte("verify-ca")):
+		connDetails.parseSSLOptions(credentials)
 		fallthrough
```
sgnn7, line 63: I think you can stack two case statements together: https://www.golangprograms.com/golang-switch-case-statements.html under "Switch multiple cases Statement"

internal/plugin/connectors/tcp/mssql/connection_details.go
doodlesbykumbi: re: the `enable` case, it isn't a valid sslmode, I think that should be `require`.

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -46,12 +54,15 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
 	switch {
 	case bytes.Equal(sslMode, []byte("disable")):
 		connDetails.SSLMode = string(credentials["sslmode"])
+		break
```
sgnn7, line 57: Since this logic path will just go directly to the return on the bottom,I think this can be simplified a lot. Some pseudocode:
```golang
...
sslMode := string(credentials["sslmode"])
if sslMode == "disable" || sslMode == "" {
    connDetails.SSLMode = defaultSSLMode
    return connDetails
}

connDetails.parseSSLOptions(credentials)

return connDetails
```

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -63,3 +74,17 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
 func (cd *ConnectionDetails) Address() string {
 	return cd.Host + ":" + strconv.FormatUint(uint64(cd.Port), 10)
 }
+
+func (cd *ConnectionDetails) parseSSLOptions(credentials map[string][]byte) {
+	cd.SSLOptions = make(map[string]string)
+
+	for _, sslOption := range sslOptions {
+		if len(credentials[sslOption]) > 0 {
```
sgnn7, line 82: Ditto: this might panic if that field isn't present in the config (which ends up in the credentials)

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -63,3 +74,17 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
...
+			value := string(credentials[sslOption])
+			if value != "" {
+				cd.SSLOptions[sslOption] = value
+			}
+		}
```
sgnn7, line 87: I think we do need to handle cases where:
- a needed ssl var is not set
- a needed ssl var is empty

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -8,50 +8,64 @@ import (
...
+		break
 	case bytes.Equal(sslMode, []byte("verify-ca")):
 		fallthrough
+	case bytes.Equal(sslMode, []byte("disable")):
+		fallthrough
 	default:
-		connDetails.SSLMode = "disable"
+		connDetails.SSLMode = defaultSSLMode
+		return connDetails
+	}
```
doodlesbykumbi, line 65: After some thought, the switch here seems unnecessary. I think our main goal right now is to force sslMode to disable always, and in the hypothetical (but not currently possible) case when sslMode is not disable we invoke parseSSLOptions. That can be achieved with:
```
sslMode := string(credentials["sslmode"])
if sslMode != "disable" {
	sslMode = "disable"
	connDetails.parseSSLOptions(credentials)
}

connDetails.SSLMode = sslMode
```

The switch seems to be anticipate something to come. I think the confusion/complexity that arises is more costly than the value we'll realise by preempting the code now. The YAGNI principle comes to mind here.

izgeri: also, can we have a debug message if we override what's written in the config with `disable`? that seems like something an end user might want to track

BradleyBoutcher: The idea for the switch was to prevent values besides disable, moreso than preempting functionality. However this is a definitely better way to do this, so I'm happy to make the change.

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -63,3 +74,17 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
...
+	cd.SSLOptions = make(map[string]string)
+
+	for _, sslOption := range sslOptions {
+		if len(credentials[sslOption]) > 0 {
+			value := string(credentials[sslOption])
+			if value != "" {
+				cd.SSLOptions[sslOption] = value
+			}
+		}
+		delete(credentials, sslOption)
```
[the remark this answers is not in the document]
jonahx: To elaborate on this, you should have a method:

```
func NewSSLOptions(credentials map[string][]byte) (SSLOptions, error) {
  // do the parsing here
}
```
Which gives a proper type to SSLOptions, too.

Also, I looked at the connector code and wasn't able to see any reason for the key deletions anyway.

internal/plugin/connectors/tcp/mssql/connection_details_test.go
```diff
@@ -108,3 +102,56 @@ func TestNewConnectionDetails(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.description, func(t *testing.T) {
+			actualConnDetails := NewConnectionDetails(tt.args.credentials)
+			actualConnDetails.parseSSLOptions(tt.args.credentials)
+
+			assert.Equal(t, tt.expected, actualConnDetails)
+
+			assert.Empty(t, tt.args.credentials)
```
doodlesbykumbi, line 154: It's worth adding a comment above this to inform that this is validating zeroization.

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -63,3 +66,17 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
...
+	}
```
doodlesbykumbi, line 81: I think zeroization should take place in a defer at the top of NewConnectionDetails, so that we get all the key-value pairs. That way we don't need zeroization in multiple places (i.e. here and https://github.com/cyberark/secretless-broker/pull/1184)

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -1,59 +1,62 @@
...
+}
+
+var sslOptions = []string{
+	"sslrootcert",
+	"sslkey",
+	"sslcert",
 }
 
 const defaultMSSQLPort = uint(1433)
+const defaultSSLMode = "disable"
```
doodlesbykumbi, line 25: `defaultSSLMode` is only being used in the tests. I think on line 52 and 54, instead of the current string literal.

internal/plugin/connectors/tcp/mssql/connection_details.go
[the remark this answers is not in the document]
BradleyBoutcher: This exact block of code is used in the PG connector. Credential deletion from the map also happens in the MySQL connector. I'm following that pattern here to remove credentials from being duplicated elsewhere in memory.

internal/plugin/connectors/tcp/mssql/connection_details.go
sgnn7: Sounds good

internal/plugin/connectors/tcp/mssql/connection_details.go
jonahx: @doodlesbykumbi I don't think this is even zeroizing.  It's just removing the key.  You'd be relying on GC to remove it, even assuming it has no other refs.

Also, didn't we have a generic solution for zeroizing at the framework level above this?

internal/plugin/connectors/tcp/mssql/connection_details.go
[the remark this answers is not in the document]
jonahx: Per my other comment below, this won't do that.   We have a framework solution in place for zeroizing already:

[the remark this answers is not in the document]
jonahx: Also re: your pg comment, if you're referring to this:

https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/tcp/pg/connect_details.go#L92

It's not being done to zeroize.  It's being done so we can bulk add "all the remaining options" in the line I linked.  

As a sidenote, I never liked that code and always thought  it was a confusing pattern, which is looking like a correct prediction :)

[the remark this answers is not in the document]
doodlesbykumbi: That is true. We do have a framework solution. The connectors don't need to zeroize credentials

Issue #1193, opened 2020-04-09
Tests exist for verifying that sslmode and certificates are propogated to the server


BradleyBoutcher:
Testing must be added to verify that each sslmode (disable, require, and verify-ca), as well as the root certificate, are propagated correctly to the server.

- [ ] Integration / unit tests exist for checking parameters have been propagated from secretless to the driver
- [ ] Integration / unit tests exist for verifying that the certificate has been processed as-needed by the driver and received / handled by the server

izgeri:
@BradleyBoutcher @doodlesbykumbi do we have a mechanism for testing this now? we've left this out of test plans for other connectors due to its complexity (and coverage with unit tests)

wondering if this card can be dropped

doodlesbykumbi:
I think...

The integration tests side of things on this card is covered by the contents of this other [card](https://github.com/cyberark/secretless-broker/issues/1167) (and the associated [PR](https://github.com/cyberark/secretless-broker/pull/1208)). 

I think what is missing and would be useful is a unit test for the [URL method](https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/tcp/mssql/connection_details.go#L111) in connection details. We need to ensure it generates the DSN that the driver expects.

doodlesbykumbi:
The integration tests side of things on this card is covered by the contents of this other [card](https://github.com/cyberark/secretless-broker/issues/1167) (and the associated [PR](https://github.com/cyberark/secretless-broker/pull/1208)). 

What is missing and would be useful is a unit test for the [URL method](https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/tcp/mssql/connection_details.go#L111) in connection details. We need to ensure it generates the DSN that the driver expects.

izgeri:
Filed #1224, closing this one now.

PR #1194, opened 2020-04-09
1165 - MSSQL support exists for sslmodes verify-ca & require

BradleyBoutcher:
sslmodes require and verify-ca dictate that TLS must
be used between secretless and the target mssql
service.

- Added support for verify-ca and building
the associated DSN
- Added support for require and building the
associated DSN

- created a map of sslmodes and their associated
params for building DSNs easily and consistently
- Modified and added tests, as needed, for the
changes and additions to connection_details.go

Todo: Add more thorough testing to verify 
parameter propagation per story #1193

### DIFF
--- a/internal/plugin/connectors/tcp/mssql/connection_details.go
+++ b/internal/plugin/connectors/tcp/mssql/connection_details.go
@@ -1,25 +1,41 @@
 package mssql
 
 import (
+	"net/url"
 	"strconv"
 )
 
 // ConnectionDetails stores the connection info to the real backend database.
 // These values are pulled from the SingleUseConnector credentials config
 type ConnectionDetails struct {
-	Host       string
-	Port       uint
-	Username   string
-	Password   string
-	SSLMode    string
-	SSLOptions map[string]string
+	Host      string
+	Port      uint
+	Username  string
+	Password  string
+	SSLParams map[string]string
 }
 
-var sslOptions = []string{
-	"sslrootcert",
-	"sslkey",
-	"sslcert",
+var sslModeToBaseParams = map[string]map[string]string{
+	sslModeDisable: {
+		"encrypt": "disable",
+	},
+	sslModeRequire: {
+		"encrypt":                "true",
+		"trustservercertificate": "true",
+	},
+	sslModeVerifyCA: {
+		"encrypt":                "true",
+		"trustservercertificate": "false",
+	},
 }
+
+const (
+	sslModeDisable  = "disable"
+	sslModeRequire  = "require"
+	sslModeVerifyCA = "verify-ca"
+)
+
+var defaultSSLMode = []byte(sslModeRequire)
 
 const defaultMSSQLPort = uint(1433)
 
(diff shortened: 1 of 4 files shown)


internal/plugin/connectors/tcp/mssql/connection_details_test.go
```diff
@@ -14,18 +14,18 @@ var defaultConnectionDetails = &ConnectionDetails{
...
 	Host:     "0.0.0.0",
 	Port:     1234,
-	SSLMode:  "disable",
-	SSLOptions: map[string]string{
-		"sslrootcert": "foo",
-		"sslkey":      "bar",
-		"sslcert":     "foobar",
+	SSLParams: map[string]string{
+		"encrypt":                "true",
+		"trustServerCertificate": "true",
```
doodlesbykumbi, line 19: `s/trustServerCertificate /trustservercertificate`, it must be all lowercase

internal/plugin/connectors/tcp/mssql/connection_details_test.go
```diff
@@ -114,6 +117,10 @@ func TestNewConnectionDetails(t *testing.T) {
 	}
 }
 
+func TestDefaultSSLModeExists(t *testing.T) {
+	assert.NotNil(t, sslModeToBaseParams[string(defaultSSLMode)])
```
doodlesbykumbi, line 121: `assert.NotNil` only works for value that can be nil. This test won't fail for an empty string. In fact, I just spotted that there's a few places in the codebase where `assert.NotNil` is being used for string values :'(. I'll create a card to fix those tests

internal/plugin/connectors/tcp/mssql/connection_details_test.go
```diff
@@ -125,28 +132,39 @@ func TestConnectionDetails_NewSSLOptions(t *testing.T) {
 			args: args{
 				credentials: map[string][]byte{},
 			},
-			expected: map[string]string{},
+			expected: sslModeToBaseParams[sslModeRequire],
+		},
+		{
+			description: "sslmode: disable",
```
doodlesbykumbi, line 138: a thought: I know this concern hasn't been introduced by this particular PR but I'm thinking we should be descriptive of our test cases. Going to look into some best practices for this.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -3,10 +3,10 @@ package mssql
 import (
 	"context"
 	"fmt"
+	"github.com/pkg/errors"
```
doodlesbykumbi, line 6: nit: maintain the order of imports (standard lib => external => internal).

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -69,15 +74,18 @@ func (cd *ConnectionDetails) Address() string {
...
+		return newSSLParams(credentials)
+	}
 
-	for _, sslOption := range sslOptions {
-		value := string(credentials[sslOption])
-		if len(value) > 0 {
-			SSLOptions[sslOption] = value
-		}
+	if sslMode == "verify-ca" {
+		params["rawcertificate"] = string(credentials["sslrootcert"])
```
doodlesbykumbi, line 87: Let's confirm that passing in a raw certificate is URL-safe, though I suspect not. This might mean that we should explore using a driver constructor that takes in the params map instead of the DSN string.

internal/plugin/connectors/tcp/mssql/connection_details_test.go
doodlesbykumbi: I think we need to use `assert.NotEmpty`

internal/plugin/connectors/tcp/mssql/connection_details_test.go
BradleyBoutcher: good catch - corrected

internal/plugin/connectors/tcp/mssql/connection_details_test.go
BradleyBoutcher: Good catch - I'll create an issue for that and correct it here

BradleyBoutcher: Corrected here, and new issue created - #1196

internal/plugin/connectors/tcp/mssql/connection_details_test.go
BradleyBoutcher: Agreed, I went through and tried to make them more descriptive, but I'm interested in learning the best practice for this as well. We should bring the team together for a small chat on that topic and document something.

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -321,15 +322,22 @@ func (connector *SingleUseConnector) writeErrorToClient(err error) {
...
 		// it'll be used between Secretless and the server.
 		// To disable TLS altogether we must change to
 		// "sqlserver://%s:%s@%s?encrypt=disable",
-		"sqlserver://%s:%s@%s?encrypt=%s",
+		dsn,
 		connDetails.Username,
 		connDetails.Password,
 		connDetails.Address(),
-		connDetails.SSLMode,
+		strings.Join(params, "&"),
```
doodlesbykumbi, line 341: I think we should refactor the code here to something like below. By using the url package to generate the url string we get encoding for free. This is mostly for the `rawcertificate` which will get damaged in transit if we don't encode.
```
query := url.Values{}
var params []string
for key, value := range connDetails.SSLParams {
	query.Add(key, value)
}

u := &url.URL{
	Scheme:   "sqlserver",
	User:     url.UserPassword(connDetails.Username, connDetails.Password),
	Host:     connDetails.Address(),
	RawQuery: query.Encode(),
}

return u.String()
```

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -318,18 +316,4 @@ func (connector *SingleUseConnector) waitForServerConnection(
...
-		// To disable TLS altogether we must change to
-		// "sqlserver://%s:%s@%s?encrypt=disable",
-		"sqlserver://%s:%s@%s?encrypt=%s",
-		connDetails.Username,
-		connDetails.Password,
-		connDetails.Address(),
-		connDetails.SSLMode,
-	)
-}
+}
```
doodlesbykumbi, line 319: missing newline at EOF

internal/plugin/connectors/tcp/mssql/connector.go
```diff
@@ -2,12 +2,10 @@ package mssql
 
 import (
 	"context"
-	"fmt"
+	"github.com/pkg/errors"
```
doodlesbykumbi, line 5: nit: maintain the order of imports (standard lib => external => internal).

internal/plugin/connectors/tcp/mssql/connection_details.go
```diff
@@ -47,37 +63,49 @@ func NewConnectionDetails(credentials map[string][]byte) *ConnectionDetails {
...
 
-	for _, sslOption := range sslOptions {
-		value := string(credentials[sslOption])
-		if len(value) > 0 {
-			SSLOptions[sslOption] = value
-		}
+	u := &url.URL{
+		Scheme:   "sqlserver",
+		User:     url.UserPassword(cd.Username, cd.Password),
+		Host:     cd.Host,
```
doodlesbykumbi, line 106: This should be `cd.Address()`. The Host field is in URL takes host and port.

internal/plugin/connectors/tcp/mssql/connector.go
doodlesbykumbi: There also needs to be whitespace between each type. So:
```
	"context"
	"fmt"

	"github.com/pkg/errors"

        ...
```

PR #1197, opened 2020-04-13
WIP - 1176 - Lower trivy threshold to detect 'high' level issues

BradleyBoutcher:
- Introduces fixes for any 'high' level trivy issues in the redhat and quickstart containers
- Lowers the trivy issue threshold to 'high'

### DIFF
--- a/demos/quick-start/docker/Dockerfile
+++ b/demos/quick-start/docker/Dockerfile
@@ -20,26 +20,28 @@ COPY etc/nginx.conf                 /etc/nginx/
 
 COPY etc/secretless.yml etc/motd    /etc/
 
+RUN apk update && apk upgrade
+
 RUN apk add --no-cache apache2-utils \
                        libc6-compat \
                        nginx \
                        tini \
                        openssh \
-                       openssl \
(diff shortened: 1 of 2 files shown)


demos/quick-start/docker/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN apk add --no-cache apache2-utils \
                        tini \
                        openssh \
                        openssl \
+    && apk upgrade --no-cache libxslt \
```
sgnn7, line 29: Can we do [this](https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Upgrade_a_Running_System) instead. We should also probably do this before any custom package install (between line 21 and 23) to take advantage of caching.

izgeri:
@BradleyBoutcher FYI I just updated the PR description to remove the WIP and to be clear about which container we updated this for

PRs are a public record of the work we do, so they should be succinct and clear about the changes within without having to click into the PR, if possible. Also, when a PR is ready to be reviewed, the WIP should be removed from the title. Please try to remember to do this on future PRs!

PR #1200, opened 2020-04-16
1188 - Zeroize credentials immediately after connecting

BradleyBoutcher:
In addition to a defer statement, called if we return
an error or panic after retreiving credentials, we
also call the zeroizing function immediately after
the connection is made and the credentials are no
longer needed.

### DIFF
--- a/internal/plugin/connectors/tcp/proxy_service.go
+++ b/internal/plugin/connectors/tcp/proxy_service.go
@@ -103,6 +103,7 @@ func (proxy *proxyService) handleConnection(clientConn net.Conn) error {
 	}()
 
 	backendCredentials, err := proxy.retrieveCredentials()
+	// zeroize credentials if we exit early due to an error
 	defer internal.ZeroizeCredentials(backendCredentials)
 	if err != nil {
 		return errors.Wrap(err, "failed on retrieve credentials")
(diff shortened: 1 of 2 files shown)


internal/plugin/connectors/tcp/proxy_service.go
```diff
@@ -103,6 +103,7 @@ func (proxy *proxyService) handleConnection(clientConn net.Conn) error {
 	}()
 
 	backendCredentials, err := proxy.retrieveCredentials()
+	// zeroize credentials if we exit early due to an error
 	defer internal.ZeroizeCredentials(backendCredentials)
```
sgnn7, line 107: not a blocker but this line should probably move below the error (or move into the cred retrieval) unless we return non-nil backendCredentials when error is set

internal/plugin/connectors/tcp/proxy_service.go
```diff
@@ -114,7 +115,8 @@ func (proxy *proxyService) handleConnection(clientConn net.Conn) error {
 	if err != nil {
 		return errors.Wrap(err, "failed on connect")
 	}
-
+	// immediately zeroize credentials after connecting
```
sgnn7, line 118: nit: Use spacing liberally. I think this line needs a newline above it.

BradleyBoutcher: Reinstated the newline above ✨

internal/plugin/connectors/tcp/proxy_service.go
BradleyBoutcher: Without the `defer` statement located before the error returns, it is possible to still access the connection details, which may not happen outside of a unit test, but it's probably better to be safe and keep the defer statement where it is.

PR #1206, opened 2020-04-20
Add MSSQL TLS test environments

doodlesbykumbi:
The only degree of freedom for MSSQL TLS test environment is a toggle on forcing TLS. As such, there are 2 environments to reflect the possibilities.


test/connector/tcp/mssql/certs/server-key.pem
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
sgnn7, line 1: does this need a gitleaks exception?

doodlesbykumbi: it has it

PR #1215, opened 2020-04-28
1212 - Generic HTTP Config Example for Elasticsearch

BradleyBoutcher:
- secretless yml configuration for use with elasticsearch api

- Supports both access token and api key based accesses

### DIFF
--- a/examples/generic_connector_configs/elasticsearch_secretless.yml
+++ b/examples/generic_connector_configs/elasticsearch_secretless.yml
@@ -1,0 +1,31 @@
+version: 2
+services:
+  token-service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:9030
+    credentials:
+      token:
+        from: keychain
+        get: service#elasticsearch/access_token
+    config:
+      headers:
+        Authorization: Bearer {{ .token }}
+      forceSSL: true
+      authenticateURLsMatching:
+        - ^http[s]*\:\/\/
+  api-key-service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:9020
+    credentials:
+      id:
+        from: keychain
+        get: service#elasticsearch/api_key_id
+      api_key:
+        from: keychain
+        get: service#elasticsearch/api_key
+    config:
+      headers:
+        Authorization: ApiKey {{ printf "%s:%s" .id .api_key | base64 }}
+      forceSSL: true
+      authenticateURLsMatching:
+        - ^http[s]*\:\/\/
--- a/examples/generic_connector_configs/README.md
+++ b/examples/generic_connector_configs/README.md
@@ -8,9 +8,14 @@ requires to authenticate.
 
 ## Sample Configurations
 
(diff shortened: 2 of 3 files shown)


examples/generic_connector_configs/elasticsearch_secretless.yml
```diff
@@ -0,0 +1,31 @@
...
+  token-service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:9030
+    credentials:
+      token:
+        from: keychain
+        get: summon#elasticsearch/access_token
+    config:
+      headers:
+        Authorization: Bearer {{ .token }}
```
doodlesbykumbi, line 12: Shouldn't the token value also be base64 encoded?

examples/generic_connector_configs/elasticsearch_secretless.yml
```diff
@@ -0,0 +1,31 @@
+version: 2
+services:
+  token-service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:9030
+    credentials:
+      token:
+        from: keychain
+        get: summon#elasticsearch/access_token
```
doodlesbykumbi, line 9: I wonder if something like` path/to/elasticsearch/access/token` would be more informative for the example.

examples/generic_connector_configs/elasticsearch_secretless.yml
```diff
@@ -0,0 +1,31 @@
...
+      forceSSL: true
+      authenticateURLsMatching:
+        - ^http[s]*\:\/\/
+  api-key-service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:9020
+    credentials:
+      id:
+        from: keychain
+        get: summon#elasticsearch/api_key_id
```
doodlesbykumbi, line 22: Same here. I wonder if something like `path/to/elasticsearch/api/key/id` and `path/to/elasticsearch/api/key` would be more informative for the example.

examples/generic_connector_configs/elasticsearch_secretless.yml
BradleyBoutcher: According to their documentation, it shouldn't be

examples/generic_connector_configs/elasticsearch_secretless.yml
BradleyBoutcher: Agreed - this may be a change we should make in all the examples. @izgeri , what do you think?

izgeri: doesn't much matter to me, but this is the syntax for the keychain provider so we may consider switching the examples to a different provider in that case

examples/generic_connector_configs/elasticsearch_secretless.yml
doodlesbykumbi: Ah OK. Based on [this documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-token.html) the Bearer token must be base64 encoded. I guess in this case the user will have stored the token base64 encoded already.

I think it would be worth stating this clearly in a comment to avoid confusion e.g. having no bas64 encoding or double encoding.

BradleyBoutcher: According to that documentation, it seems that only kerberos tickets as a parameter involve base 64. It doesn't mention using base 64 for bearer tokens as far as I can tell.

PR #1220, opened 2020-05-05
Go fmt has been ran on secretless project dir

BradleyBoutcher:
There were many outstanding style issues, as go
fmt has not been ran on the entire project until now.
This change was caused by running `go fmt ./...` from the
top level directory of the project.


BradleyBoutcher:
> @BradleyBoutcher Can you split out into a different commit (or remove) the changes to go.sum?

Done!

izgeri:
it's nice when things move in this direction on a PR! 🍾 

Screen Shot 2020-05-05 at 3 44 16 PM

Issue #1233, opened 2020-05-21
Secretless publishes images on tag-triggered builds


izgeri:
At current, Secretless master branch Jenkins pipelines run whenever there is a change to the master branch, and we determine whether to publish based on whether the most recent tag matches the current commit.

https://github.com/cyberark/secretless-broker/blob/8d7f757f7bcfb2cb26283e685da06e0cc8121d84/bin/publish#L36

We now have support in Jenkins for tag-triggered builds (docs link TBA), so we'd like to update the project to publish when it's a tag build rather than publishing based on a logical question that could fail if another commit is pushed to the master branch before the build completes.

hughsaunders:
* [Docs on Jenkins triggers](https://github.com/conjurinc/docs/blob/master/reference/jenkins.md#jenkins-pipeline-github-triggers)
* [Example conditional stage for tag jobs only](https://github.com/cyberark/parse-a-changelog/blob/master/Jenkinsfile#L43)

PR #1236, opened 2020-05-22
Secretless publishes images on tag-triggered builds

doodlesbykumbi:
### What does this PR do?
Non-internal images are published on tag-triggered builds
The Jenkinsfile is the entrypoint to the change. The publishing steps have been split in 2. 
  1. An internal publish which happens only on master.
  2. A non-internal publish which happens only on tags.

  ~~**TODO** The non-internal publish happens to also publish latest. This isn't great because if there's a retagging then latest will represent some older version.~~
  ~~**TODO** Should we use the TAG_NAME env var from Jenkins or should we use the VERSION in the Go file. Most of the time they should be the same. What do we do when they're not!?~~

Connected to https://github.com/cyberark/secretless-broker/issues/1233

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -169,15 +169,7 @@ pipeline {
       when { tag "v*" }
       steps {
         // The tag trigger sets TAG_NAME as an environment variable
-
-        // Clean up first
-        sh 'docker run -i --rm -v $PWD:/src -w /src alpine/git clean -fxd'
-
         sh 'summon -e production ./bin/publish'
-
-        // Clean up again...
-        sh 'docker run -i --rm -v $PWD:/src -w /src alpine/git clean -fxd'
-        deleteDir()
       }
     }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -153,13 +153,31 @@ pipeline {
       }
     }
 
-    stage('Push Images') {
+    stage('Push Images Internally') {
       when {
         branch 'master'
       }
 
       steps {
+        sh './bin/publish_internal'
+      }
+    }
+
+    stage('Push Images') {
+      agent { label 'releaser-v2' }
+      // Only run this stage when triggered by a tag
+      when { tag "v*" }
+      steps {
+        // The tag trigger sets TAG_NAME as an environment variable
+
+        // Clean up first
+        sh 'docker run -i --rm -v $PWD:/src -w /src alpine/git clean -fxd'
+
         sh 'summon -e production ./bin/publish'
+
+        // Clean up again...
+        sh 'docker run -i --rm -v $PWD:/src -w /src alpine/git clean -fxd'
+        deleteDir()
       }
     }
 
--- a/bin/publish
+++ b/bin/publish
@@ -7,45 +7,29 @@ set -e
 readonly REGISTRY="cyberark"
 readonly VERSION="$(short_version_tag)"
 readonly FULL_VERSION_TAG="$(full_version_tag)"
-readonly INTERNAL_REGISTRY="registry.tld"
 readonly REDHAT_IMAGE="scan.connect.redhat.com/ospid-18d9f51d-9c0c-4031-9f9e-ef08aa2ff409/secretless-broker"
 readonly IMAGES=(
   "secretless-broker"
   "secretless-broker-quickstart"
 )
-
 readonly TAGS=(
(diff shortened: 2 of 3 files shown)


Jenkinsfile
```diff
@@ -153,13 +153,31 @@ pipeline {
...
+
+    stage('Push Images') {
+      agent { label 'releaser-v2' }
+      // Only run this stage when triggered by a tag
+      when { tag "v*" }
+      steps {
+        // The tag trigger sets TAG_NAME as an environment variable
+
+        // Clean up first
+        sh 'docker run -i --rm -v $PWD:/src -w /src alpine/git clean -fxd'
```
izgeri, line 174: Are these required? I think those might have been oddities of the parse-a-changelog pipeline

I'm not sure the `deleteDir` is needed below either


If these are needed, can we please add a comment explaining why we're adding them to this PR?

bin/publish
```diff
@@ -7,45 +7,59 @@ set -e
...
   "secretless-broker"
   "secretless-broker-quickstart"
 )
 
-readonly TAGS=(
-  "$VERSION"
-  "latest"
-)
+function main() {
+  TAG_NAME="${TAG_NAME:-}"
```
izgeri, line 18: should we also have a check here for if TAG_NAME is not set, and have the script fail in this case? this script should only be run from tag-triggered builds, I think

similar to here: https://github.com/cyberark/parse-a-changelog/blob/69f941471eca1a24081f573959875ce4ff54bbd3/publish.sh#L5:L10

bin/publish_internal
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -e
+
+. bin/build_utils
+
+readonly REGISTRY="cyberark"
```
izgeri, line 7: I don't think this is used here
```suggestion
```

Jenkinsfile
doodlesbykumbi: Yeah maybe this isn't necessary

doodlesbykumbi: I'm taking it out until I find out if there's a good reason for it

bin/publish
```diff
@@ -1,51 +1,70 @@
...
-    echo "Revision $git_description matches version $VERSION exactly. Pushing to Dockerhub..."
+  # Add less specific version tags, eg. given 1.2.3, add 1.2 and 1
+  for tag in $(gen_versions "$VERSION"); do
+    TAGS+=("${tag}")
+  done
+
+  # Add 'latest' tag only when TAG_NAME is the highest semantic tag
+  local LATEST_TAG
+  LATEST_TAG="$(
+    git for-each-ref refs/tags --sort=-refname --format='%(refname:short)' --count=1
```
sgnn7, line 46: This doesn't do correct ordering since it's lexical:
```
~/checkout/conjur $ git tag v1.10.5
~/checkout/conjur $  git for-each-ref refs/tags --sort=-refname --format='%(refname:short)' 
v1.6.0
v1.5.1
v1.5.0
v1.4.7
v1.4.6
v1.4.4
v1.4.3
v1.4.2
v1.4.1
v1.4.0
v1.3.7
v1.3.6
v1.3.5
v1.3.4
v1.3.3
v1.3.2
v1.3.1
v1.3.0
v1.2.0
v1.10.5 <----
v1.1.2
v1.1.1
v1.1.0
v1.0.1
v1.0.0
v0.9.0
v0.8.1
v0.8.0
v0.7.0
v0.6.0
v0.3.0
v0.2.0
v0.1.1
v0.1.0
delete
1.2.0

File: bin/publish [hunk 5]
```diff
@@ -1,51 +1,70 @@
 #!/bin/bash
 
+if [[ -z "${TAG_NAME:-}" ]]; then
+  echo "Please supply environment variable TAG_NAME."
+  echo "If you see this error in Jenkins it means the publish script was run"
+  echo "for a build that wasn't triggered by a tag - please check publish stage conditions."
+  exit 1
+fi
+
 set -e
```
sgnn7, line 10: Is there a reason that `-e` is below the usage? can't hurt to have that run there as well.

bin/publish
doodlesbykumbi: Good catch. Thanks.`git tag | sort -r --version-sort` does it right.

Issue #1235, opened 2020-05-21
Jenkins pipeline auto-creates GitHub releases on tags


izgeri:
Assuming tag-triggered builds have been added to the Secretless pipeline in #1233 and Goreleaser runs as part of the pipeline and archives artifacts as in #1234, then the Secretless pipeline may be updated to auto-publish on tag-triggered builds a GitHub pre-release with:
- Attached artifacts from Goreleaser and the tagged branch
- A copy of the changelog for the tag version

Note that the GitHub release should be marked as a _prerelease_ until additional stability tests have been run.

AC:
- [ ] Secretless auto-publishes a GitHub prerelease on tag-triggered builds with auto-copied change log and attached artifacts from the pipeline

Depends on #1233 
Depends on #1234

Issue #1234, opened 2020-05-21
Goreleaser runs in pipeline and saves artifacts in Jenkins


izgeri:
At current, we run Goreleaser manually and upload artifacts to GitHub releases manually with each Secretless tag/release. This has caused problems in the past where the Goreleaser config may be out of date and the goreleaser script fails, which can result in a tag having to be deleted until the goreleaser script is fixed.

In addition, when running the script this way, the output artifacts are build on a local development machine instead of the remote Jenkins environment where all of the tests are run. It would be much cleaner to run the script in Jenkins, and if we run it regularly we also would get early notice of changes in Goreleaser that break our config.

In this effort, we should update the pipeline to run Goreleaser (always, to get alerted early on failures) and archive the output artifacts. To keep the pipeline runtime low the Goreleaser script may be run in the same parallel stage as the image build.

AC:
- [ ] Goreleaser runs in pipeline and archives output artifacts in Jenkins

PR #1239, opened 2020-05-29
Goreleaser runs in pipeline and saves artifacts in Jenkins

doodlesbykumbi:
### What does this PR do?
  + Goreleaser runs in pipeline and saves artifacts in Jenkins
  + Goreleaser docker uses latest Go releaser version
The root of the changes are:
  + Jenkinsfile
  + Dockerfile.relaser

Connected to #1234 and #1235 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -118,7 +118,7 @@ nfpms:
     homepage: https://secretless.io
     license: "Apache 2.0"
     maintainer: CyberArk Maintainers <conj_maintainers@cyberark.com>
-    name_template: "{{.ProjectName}}_{{.Version}}_{{.Arch}}"
+    file_name_template: "{{.ProjectName}}_{{.Version}}_{{.Arch}}"
     vendor: CyberArk
 
 snapshot:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -61,6 +61,11 @@ builds:
     post:
       # Copy the binary out into the <dist> path, and give the copy the name we want
       # in the release <extra_files>.
+      # e.g. Suppose a windows amd64 build generates a binary at
+      # path/to/secretless-broker.exe. This will be copied to
+      # path/to/../secretless-broker-windows_amd64.exe. The copy path can then be added to
+      # the release <extra_files> and will result in a release artifact with the name
+      # secretless-broker-windows_amd64.exe.
       - cp "{{ .Path }}" "{{ dir .Path }}/../secretless-broker-{{.Target}}{{.Ext}}"
 
 - <<: *secretless-broker-build
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -135,9 +135,6 @@ snapshot:
   name_template: "{{ .Tag }}-next"
 
 release:
-  github:
-    owner: conjurinc
-    name: playroom
   disable: false
   draft: true
   extra_files:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -140,6 +140,6 @@ release:
   extra_files:
     - glob: NOTICES.txt
     - glob: LICENSE
-    - glob: dist/goreleaser/CHANGELOG.md
+    - glob: CHANGELOG.md
     - glob: dist/goreleaser/secretless-broker-linux_amd64
     - glob: dist/goreleaser/secretless-broker-darwin_amd64
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,7 +21,7 @@ builds:
   ldflags: -s -w -linkmode external -extldflags "-static"
 
 - <<: *summon-2-build
-  id: summon2-osx
+  id: summon2-darwin
   env:
   - CGO_ENABLED=1
   - CC=o64-clang
@@ -57,9 +57,14 @@ builds:
   # final binary. See `Static long version tags` in the `Building` section
   # of `CONTRIBUTING.md` for more information.
   ldflags: -s -w -linkmode external -X "github.com/cyberark/secretless-broker/pkg/secretless.Tag={{ .ShortCommit }}" -extldflags "-static"
+  hooks:
+    post:
+      # Copy the binary out into the <dist> path, and give the copy the name we want
+      # in the release <extra_files>.
+      - cp "{{ .Path }}" "{{ dir .Path }}/../secretless-broker-{{.Target}}{{.Ext}}"
 
 - <<: *secretless-broker-build
-  id: secretless-broker-osx
+  id: secretless-broker-darwin
   env:
   - CGO_ENABLED=1
   - CC=o64-clang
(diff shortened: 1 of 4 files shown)


bin/build_release
```diff
@@ -5,11 +5,6 @@ CURRENT_DIR=$("$(dirname "$0")/abspath")
 # xgo because it allows cross-compilation
 GORELEASER_IMAGE="cyberark/goreleaser:latest-xgo"
```
izgeri, line 6: whoa! are we still not able to use the official goreleaser image in our project?

doodlesbykumbi: goreleaser doesn't provide an image with cross compilation capabilities :(

.goreleaser.yml
```diff
@@ -118,12 +123,21 @@ nfpms:
...
 
 snapshot:
   name_template: "{{ .Tag }}-next"
 
 release:
-  disable: true
-  prerelease: auto
+  github:
+    owner: conjurinc
+    name: playroom
```
sgnn7, line 135: Both of these GH params are wrong. Also, I think this will try to auto-release the broker on GH too when we run goreleaser.

sgnn7: Actually, this GH release should probably be completely removed afaict

.goreleaser.yml
```diff
@@ -57,9 +57,14 @@ builds:
   # final binary. See `Static long version tags` in the `Building` section
   # of `CONTRIBUTING.md` for more information.
   ldflags: -s -w -linkmode external -X "github.com/cyberark/secretless-broker/pkg/secretless.Tag={{ .ShortCommit }}" -extldflags "-static"
+  hooks:
+    post:
+      # Copy the binary out into the <dist> path, and give the copy the name we want
+      # in the release <extra_files>.
+      - cp "{{ .Path }}" "{{ dir .Path }}/../secretless-broker-{{.Target}}{{.Ext}}"
```
sgnn7, line 64: The comment should reflect what the final name should look like (e.g. `secretless-broker-foo-blah.bar`)

Jenkinsfile
```diff
@@ -163,13 +178,23 @@ pipeline {
...
+        stage('Push Images') {
+          steps {
+            // The tag trigger sets TAG_NAME as an environment variable
+            sh 'summon -e production ./bin/publish'
+          }
+        }
+        stage('Create draft release') {
+          steps {
+            sh "summon --provider summon-conjur --yaml 'GITHUB_TOKEN: !var github/users/conjur-jenkins/api-token' ./bin/build_release"
+            archiveArtifacts 'dist/goreleaser/'
```
sgnn7, line 195: Doesn't build already archive the artifacts?

Jenkinsfile
```diff
@@ -27,9 +27,24 @@ pipeline {
...
+        stage('Build Release Artifacts') {
+          when {
+            branch 'master'
+          }
+
+          steps {
+            sh './bin/build_release --snapshot'
+            archiveArtifacts 'dist/goreleaser/'
+          }
+        }
```
sgnn7, line 47: Building release artifacts on each build step before even running tests seems like an increase to the build time with no benefits to the developer for being done this early. If we really want the artifacts saved, we should probably do it when the tests all pass.

bin/build_release
sgnn7: The goreleaser main contributor is pretty responsive - maybe we can provide a PR?

Jenkinsfile
doodlesbykumbi: Build only runs on master

bin/build_release
doodlesbykumbi: Might do this, yup

.goreleaser.yml
doodlesbykumbi: The params are wrong on purpose, for testing. The params won't be necessary on merge because it'll infer from the repo.

Jenkinsfile
doodlesbykumbi: Good point. I can move this particular stage to after the tests

Jenkinsfile
doodlesbykumbi: Whereas the complete draft release is on tag

.goreleaser.yml
doodlesbykumbi: Addressed

.goreleaser.yml
izgeri: maybe it's worth leaving this unresolved until these are removed prior to merging? otherwise we may accidentally merge with these in (which we don't want)

bin/Dockerfile.releaser
```diff
@@ -9,15 +9,6 @@ RUN apt-get update && \
...
-    GORELEASER_SHA=34b7e3b843158bd0714d1be996951685496491adab4524fb1198ae144ab06ba3 \
-    GORELEASER_DOWNLOAD_FILE=goreleaser_Linux_x86_64.tar.gz
-
-ENV GORELEASER_DOWNLOAD_URL=https://github.com/goreleaser/goreleaser/releases/download/v${GORELEASER_VERSION}/${GORELEASER_DOWNLOAD_FILE}
-
-RUN wget "${GORELEASER_DOWNLOAD_URL}" && \
-    echo "$GORELEASER_SHA $GORELEASER_DOWNLOAD_FILE" | sha256sum -c - && \
-    tar -xzf "$GORELEASER_DOWNLOAD_FILE" -C /usr/bin/ goreleaser && \
-    rm "$GORELEASER_DOWNLOAD_FILE"
+RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh
```
izgeri, line 13: I just commented on this dockerfile in your cyberark/sidecar-injector#36 PR - and it occurs that I was commenting below on the _tag_ `cyberark/goreleaser` and you may not remember that we had an actual project to build a custom goreleaser image: https://github.com/conjurinc/goreleaser-docker/blob/master/Dockerfile

now we have this in two projects - we should definitely try to contribute the fix upstream sooner rather than later, and file tech debt issues in these two repos to update the image used. if we can't fix upstream, I suppose we'll have to resurrect our custom image :/

doodlesbykumbi: I get 404 on https://github.com/conjurinc/goreleaser-docker/blob/master/Dockerfile. We should definitely have a custom image to avoid the duplication (both repos are building the same image). Contributing to goreleaser would be nice bonus.

sgnn7:
@doodlesbykumbi LGTM but can we get a green build and fixup those playroom vars before I do the approve?

bin/Dockerfile.releaser
izgeri: ah, it's private archived and I can't change the access rights. it does exist!!

I was hoping to move away from bespoke goreleaser images - I thought we had made good progress, then these PRs brought to light for me that there were still issues. sounds like you're taking a good approach moving forward, though :)

doodlesbykumbi:
@sgnn7 All set

Issue #548, opened 2018-12-03
PostgreSQL and MySQL connectors support sslmode=verify-full


izgeri:
`verify-full` is equivalent to `VERIFY_IDENTITY` as stated in the [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_ssl-mode):
> VERIFY_IDENTITY: Like VERIFY_CA, but additionally perform host name identity verification by checking the host name the client uses for connecting to the server against the identity in the certificate that the server sends to the client:

AC:
- [ ] perform host name identity verification by checking the host name the client uses for connecting to the server against the identity in the certificate that the server sends to the client

izgeri:
Note - there was a request to get this fixed in https://github.com/cyberark/secretless-broker/issues/922

izgeri:
Today we briefly looked into what it would take to get this implemented. We noticed the two commented lines:


and wanted to validate whether simply uncommenting them would get this working e2e. It doesn't appear to have worked.

To try to get it working, I cloned the `latest` branch, uncommented these two lines, and ran `./bin/build_darwin`.

I spun up a MySQL instance in a Docker container with:
```
docker run -d -p 3307:3306 --name mysql -e MYSQL_ROOT_PASSWORD=myrootpassword -e MYSQL_DATABASE=mydb -e MYSQL_USER=myuser -e MYSQL_PASSWORD=mypassword mysql:5
```
and copied the `ca.pem`, `server-cert.pem`, and `server-key.pem` files from `/var/lib/mysql` into my local dir.

Then I ran `./dist/darwin/amd64/secretless-broker -f secretless.yml` where my secretless config was
```
version: "2"
services:
  test-verify-full:
    protocol: mysql
    listenOn: tcp://localhost:3306
    credentials:
      host: localhost
      port: 3307
      username: myuser
      password: mypassword
      sslmode: verify-full
      sslrootcert:
        from: file
        get: ca.pem
      sslcert:
        from: file
        get: server-cert.pem
      sslkey:
        from: file
        get: server-key.pem
```

I tried to run `mysql -h 127.0.0.1 -P 3306 --ssl-mode=DISABLED`, but I got the error
```
ERROR: 2000 (HY000): tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
```

Replacing `verify-full` with `verify-ca` got me into the DB.

porjo:
[...]

PR #1249, opened 2020-06-11
Add support for sslmode=verify-full for mysql and pg

doodlesbykumbi:
This means full verification of rootCA and hostname. The value for the hostname to check against is either 'sslhost' or 'host', 'sslhost' takes precedence when it is not empty.

### What does this PR do?

  + Adds support for  sslmode=verify-full for mysql and pg
  + Adds `sslhost` option that can override the `host` for certificate hostname verification
  + Add integration tests

  The entrypoint to the feature is `internal/plugin/connectors/tcp/ssl/ssl.go`. This file contains the TLS implementation for both mysql and pg.

  N/A
Connected to https://github.com/cyberark/secretless-broker/issues/548

#### Change log

#### Test coverage

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs. https://github.com/cyberark/secretless-docs/issues/255


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- MySQL and PostgreSQL connectors support SSL host name verification with `verify-full`
+  SSL mode. Also adds optional `sslhost` configuration parameter that is compared to the
+  server's certificate SAN. [#1249](https://github.com/cyberark/secretless-broker/pull/1249)
```
izgeri, line 12: can we please refer to the issue it's resolving rather than the PR? usually the PRs have less contextual info

doodlesbykumbi: Done

internal/plugin/connectors/tcp/ssl/ssl.go
```diff
@@ -51,18 +51,29 @@ func NewDbSSLMode(o options, requireCanVerifyCAOnly bool) (DbSSLMode, error) {
...
-	//case "verify-full":
-	//	sslMode.ServerName = o["host"]
+
+	case "verify-full":
+		// Use stdlib's verification
+		sslMode.InsecureSkipVerify = false
+		sslMode.VerifyCaOnly = false
+
+		// 'sslhost', when not empty, takes precedence over 'host'
+		if o["sslhost"] != "" {
```
BradleyBoutcher, line 69: I'm not sure if this should be an empty string check or a length check. In either case, my follow-up PR will use a structured representation instead, so this shouldn't be an issue.

internal/plugin/connectors/tcp/mysql/connection_details.go
```diff
@@ -2,57 +2,73 @@ package mysql
...
-	if credentials["password"] != nil {
+	if len(credentials["password"]) > 0 {
 		connDetails.Password = string(credentials["password"])
 	}
 
-	// Make sure that we process the SSL mode arg even if it's not specified
-	// otherwise it will get ignored
-	if _, ok := credentials["sslmode"]; !ok {
-		credentials["sslmode"] = []byte("")
+	for _, sslOption := range sslOptions {
```
BradleyBoutcher, line 57: Below you use a shorter method for this:
`connectionDetails.SSLOptions[sslOption] = string(options[sslOption])`
I think you can use that here too

PR #1253, opened 2020-06-16
Add PG int test exception to gitleaks

BradleyBoutcher:
Adds gitleaks exception to a PG
docker-compose file that uses a dummy
password

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -198,7 +198,8 @@ files = [
 ]
 regexes = [
   "AKIAJC5FABNOFVBKRWHA", # sample AWS key in AWS HTTP connector comments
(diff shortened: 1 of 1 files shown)


.gitleaks.toml
```diff
@@ -195,6 +195,7 @@ files = [
   "test/ssh/id_(.*)", # since-removed ssh test certs
   "test/util/ssl/(.*)", # test ssl certs
   "internal/plugin/connectors/tcp/mssql/connection_details_test.go", # fake cert string
+  "test/connector/http/conjur/docker-compose.yml",  # fake password
```
izgeri, line 198: would it be better to have this as a regex as `SuperSecure`?

adding the file exception leaves open the possibility that at some point in the future we'll add a real secret to this file and gitleaks will ignore it

PR #1256, opened 2020-06-16
Update gitleaks.toml to support gitleaks 4

ismarc:
### What does this PR do?
This updates `.gitleaks.toml` to support the new config changes in gitleaks v4.  It is done in a backwards compatible way with v3 and v2 until everyone has upgraded to v4.  This will require whitelisted values or files to be update in two locations, but is the only way to support both versions.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -150,6 +150,12 @@ tags = ["key", "Mailgun"]
 description = "Password in URL"
 regex = '''[a-zA-Z]{3,10}:\/\/[^\/\s:@]{3,20}:[^\/\s:@]{3,20}@.{1,100}\/?.?'''
 tags = ["key", "URL", "generic"]
+[[rules.whitelist]]
+description = "Testing/example url for postgresql containers"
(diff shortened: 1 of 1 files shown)


.gitleaks.toml
```diff
@@ -196,6 +214,25 @@ files = [
...
+  "test/ssh_agent_handler",
+  "test/connector/http/generic/certs",
+  "test/connector/ssh",
+  "test/connector/ssh_agent",
+  "test/connector/tcp/mssql/certs",
+  "test/ssh",
+  "test/util/ssl",
+  "internal/plugin/connectors/tcp/mssql"
+]
 regexes = [
```
izgeri, line 236: this section should be removed, right?

does the v4 gitleaks parse the commit exclusions too?

ismarc: We need to preserve any existing values for backwards compatibility until everyone has upgraded to v4.  v4 does parse the commit exclusions, I just translated existing file based exclusions rather than interpreting intention.

sgnn7:
@ismarc Do you mind rebasing this branch - seems like a good addition but it went stale vs master.

Issue #1290, opened 2020-06-26
HTTP Generic Connector supports URL parameter injection


BradleyBoutcher:
## Is your feature request related to a problem? Please describe.

Certain APIs require the ability to pass authentication data through methods other than `Headers`, which is currently the only supported injection capability of the Generic HTTP connector. To support, more use cases, we would need to allow query params to be passed in by the user's secretless configuration.

## Describe the solution you would like

The Generic HTTP Connector will create an encoded query, attached to the URL passed through Secretless. For example, the user configuration would look like this:
```
services:
  # Injects basic auth credentials and uses http on upstream connection
  basic_auth_http:
    connector: generic_http
    listenOn: tcp://0.0.0.0:8080
    credentials:
      username: someuser
      password: testpassword
      key: somekey
    config:
      credentialPatterns:
        username: '[^:]+'
      headers:
        Authorization: ' Basic {{ printf "%s:%s" .username .password | base64 }}'
      params:
        origin: "foo"
        key: "{{ .key }}"
      authenticateURLsMatching:
        - ^http
```

The keys and values within the `params` section would then be passed into a function for constructing a URL, such as this:
> Note: This example would exist within the `Connect` method of `internal/plugins/http/generic/connector.go`

> Note: `renderedParams` is a placeholder, modeled after `renderedHeaders` in the above file, which uses templating for filling values
```
	// Add configured params to request
	params, err := c.config.renderedParams(credentialsByID)
	if err != nil {
		return fmt.Errorf("failed to render headers: %s", err)
	}
	var query url.Values
	if len(r.URL.RawQuery) > 0 {
		query = r.URL.Query()
	} else {
		query = url.Values{}
	}
	for key, value := range params {
		query.Add(key, value)
	}
	r.URL.RawQuery = query.Encode()
```

So, if the user passed a request such as this through their shell:

```
http_proxy=localhost:9020 curl -X GET "http://httpbin.org/anything" -H "accept: application/json"
```

The final url would look like this, after Secretless does its thing:

```
http://httpbin.org/anything?origin=foo&key=somekey
```

## Additional context

Wikipedia describes the [query string](https://en.wikipedia.org/wiki/Query_string) in more detail.

Postman uses the category `params` for adding or inspecting URL query params, which inspired our naming convention.

sgnn7:
The only suggestion I would make is that `params` changes into `queryParams` just to make it clear as to what it's doing

sgnn7:
[...]

PR #1291, opened 2020-06-26
HTTP Generic Connector supports URL parameter injection

BradleyBoutcher:
## What does this PR do?
This PR provides support for `query params` in the config for the generic HTTP connector. This will allow the building of a query string, with any necessary credentials injected into it.

- modified 'renderHeader' function to allow rendering of multiple input fields of the same type
- Moved the validation and saving of template strings to its own function to reduce code reuse
- Added support for passing query parameters through a secretless configuration file, and appending them to the url in the http request

To manually test this, add the `queryParam` section to the `config` subsection of a `secretless.yml` file. Add a dummy key and value (ex. foo: "bar"), and then use secretless as a proxy for an http request. The key and value should be appended to your request, and your response will reflect this.

Connected to #1290 

#### Change log
- [X] The CHANGELOG has been updated, or

#### Test coverage
- [X] This PR includes new unit and integration tests to go with the code changes, or
  - Added unit tests for parsing of query params

#### Documentation
- [X] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs

### DIFF
--- a/internal/plugin/connectors/http/generic/README.md
+++ b/internal/plugin/connectors/http/generic/README.md
@@ -67,11 +67,16 @@ services:
       password:
         from: conjur
         get: somepassword
+      address:
+        from: conjur
+        get: address
     config:
       credentialValidations:
         username: '[^:]+'    # username cannot contain a colon
       headers:
         Authorization: "Basic {{ printf \"%s:%s\" .username .password | base64 }}"
+      queryParams:
+        location: "{{ .address }}"
       forceSSL: true
       authenticateURLsMatching:
         - ^http
@@ -96,6 +101,41 @@ template](https://golang.org/pkg/text/template/), as defined in the
 You can refer to your credentials in this template using the credential name
 preceded by a `.` (eg, `.username` and `.password` refer to the credentials
 `username` and `password`). At runtime, Secretless will replace these
+credential references with your real credentials.
+
+As you can see in the Basic auth example, the `text/template` package has
+powerful transformation features.  You can use `printf` for formatting and
+compose functions using pipes `|`.  See the text template package docs linked
+above for detailed information on these and other features.
+
+### `queryParams`
+
+Like `headers`, this is another key section. The `queryParams` section
+is used to generate a query string, which is appended to your existing URL
+without replacing any existing query parameters.
+
+The _keys_ of the queryParams are defined by the yaml keys.  In the examples
+above, these query parameter key is `location`.
+
+The query parameter _values_ are defined using a [Go text
+template](https://golang.org/pkg/text/template/), as defined in the
+`text/template` package.
+
+In the above example, let us say that your request URL looks like the following,
+
+```
+http://anything.com/foo?fruit=apple
+```
+
+After proxying through secretless, your request URL would look like the following,
+
+```
+http://anything.com/foo?location=valueofaddress&fruit=apple
+```
+
+You can refer to your credentials in this template using the credential name
+preceded by a `.` (eg, `.address` will refer to the credential
+`address`). At runtime, Secretless will replace these
 credential references with your real credentials.
 
 As you can see in the Basic auth example, the `text/template` package has
--- a/internal/plugin/connectors/http/generic/README.md
+++ b/internal/plugin/connectors/http/generic/README.md
@@ -140,7 +140,7 @@ credential references with your real credentials.
 
 As you can see in the Basic auth example, the `text/template` package has
 powerful transformation features.  You can use `printf` for formatting and
-compose functions using pipes `|`.  See the text template package docs linked
+you can compose functions using pipes `|`.  See the text template package docs linked
 above for detailed information on these and other features.
 
 #### `credentialValidations`
--- a/internal/plugin/connectors/http/generic/README.md
+++ b/internal/plugin/connectors/http/generic/README.md
@@ -115,7 +115,7 @@ is used to generate a query string, which is appended to your existing URL
 without replacing any existing query parameters.
 
 The _keys_ of the queryParams are defined by the yaml keys.  In the examples
-above, these query parameter key is `location`.
+above, the query parameter key is `location`.
 
 The query parameter _values_ are defined using a [Go text
 template](https://golang.org/pkg/text/template/), as defined in the
--- a/internal/plugin/connectors/http/generic/config.go
+++ b/internal/plugin/connectors/http/generic/config.go
@@ -3,6 +3,7 @@ package generic
 import (
 	"encoding/base64"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strings"
 	"text/template"
@@ -14,6 +15,7 @@ import (
 type config struct {
 	CredentialPatterns map[string]*regexp.Regexp
 	Headers            map[string]*template.Template
+	QueryParams        map[string]*template.Template
 	ForceSSL           bool
 }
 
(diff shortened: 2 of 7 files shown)


internal/plugin/connectors/http/generic/config_test.go
```diff
@@ -59,6 +64,32 @@ func Test_newConfig(t *testing.T) {
...
+		if err != nil {
+			return
+		}
+
+		// A simple test case
+		assert.Equal(t, "Foo Bar", params["NameWithSpaces"])
+
+		// Assert against value calculated with independent base64 encoder.
+		assert.Equal(t, "c29tZUtleQ==", params["Key"])
+	})
```
sgnn7, line 92: I'd like to see a couple of more tests here:
- One with special/utf chars so that we know encode is working properly
- One with existing url params so that we can be sure that appending works

internal/plugin/connectors/http/generic/config.go
```diff
@@ -34,34 +35,35 @@ func (c *config) validate(credsByID connector.CredentialValuesByID) error {
 	return nil
 }
 
-// renderedHeaders returns the config's header templates filled in with the
+// Render returns the config's `config` section templates filled in with the
 // given credentialValues.
-func (c *config) renderedHeaders(
+func render(
```
sgnn7, line 40: Hmm... why was this changed from being a method on config object?

sgnn7: I think we need a better name than just `render` - maybe we should add one more word to describe what it's actually rendering.

BradleyBoutcher: - Before, the only thing it needed from `config` was access to the Header object. Since we now pass those objects in as a parameter, it seemed unnecessary to keep it as an object specific method. I don't mind either way.
- Agreed on the naming, I can make it more specific

internal/plugin/connectors/http/generic/README.md
```diff
@@ -67,11 +67,16 @@ services:
       password:
         from: conjur
         get: somepassword
+      location:
```
JakeQuilty, line 70: Should this be "address" since line 79 uses `.address` or vice versa?

internal/plugin/connectors/http/generic/README.md
```diff
@@ -67,11 +67,16 @@ services:
...
+        from: conjur
+        get: address
     config:
       credentialValidations:
         username: '[^:]+'    # username cannot contain a colon
       headers:
         Authorization: "Basic {{ printf \"%s:%s\" .username .password | base64 }}"
+      queryParams:
+        location: "{{ .address }}"
```
JakeQuilty, line 79: See above comment

BradleyBoutcher: fixed!

izgeri:
@BradleyBoutcher please use the PR template next time, it's really handy to make sure we're adding adequate tests, docs, and change log messages with each PR.

internal/plugin/connectors/http/generic/config.go
```diff
@@ -85,23 +87,45 @@ func newConfig(cfgYAML *ConfigYAML) (*config, error) {
...
+	errs validation.Errors) (map[string]*template.Template, validation.Errors) {
+	parsedTemplates := make(map[string]*template.Template)
+	// Validate and save template strings
+	for tmplName, tmplStr := range templates {
+		tmpl := newHTTPTemplate(tmplName)
 		// Ignore pointer to receiver returned by Parse(): it's just "tmpl".
 		_, err := tmpl.Parse(tmplStr)
 		if err != nil {
-			errs[header] = fmt.Errorf("invalid header template: %q", err)
+			errs[tmplName] = fmt.Errorf("invalid query param template: %q", err)
```
sgnn7, line 109: This function is run for both headers and query params so this error shouldn't be used for both.

internal/plugin/connectors/http/generic/config.go
```diff
@@ -85,23 +87,45 @@ func newConfig(cfgYAML *ConfigYAML) (*config, error) {
...
+
+	if err := errs.Filter(); err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+func validateAndSaveTemplateString(templates map[string]string,
+	errs validation.Errors) (map[string]*template.Template, validation.Errors) {
```
sgnn7, line 101: This is an odd indentation style. Maybe something like this for readability would be better:
```golang
func foo(
    blah int,
    baz int,
)(blahblah int) {

    logicstuff()
}
```

internal/plugin/connectors/http/generic/config.go
```diff
@@ -85,23 +87,45 @@ func newConfig(cfgYAML *ConfigYAML) (*config, error) {
...
+	cfg.QueryParams, errs = validateAndSaveTemplateString(cfgYAML.QueryParams, errs)
+
+	if err := errs.Filter(); err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+func validateAndSaveTemplateString(templates map[string]string,
```
sgnn7, line 100: I think something shorter may be needed. How about `unmarshalYAMLParams`? Something else?

internal/plugin/connectors/http/generic/config.go
```diff
@@ -85,23 +87,45 @@ func newConfig(cfgYAML *ConfigYAML) (*config, error) {
...
 
-	if err := errs.Filter(); err != nil {
-		return nil, err
+func appendQueryParams(URL url.URL, params map[string]string) string {
+	var query url.Values
+	if len(URL.RawQuery) > 0 {
+		query = URL.Query()
+	} else {
+		query = url.Values{}
+	}
```
sgnn7, line 123: Can this be flattened somehow? Eg:
```golang
query := url.Values{}
if len(URL.RawQuery) > 0  {
    query = URL.Query()
}
```

internal/plugin/connectors/http/generic/config_test.go
```diff
@@ -59,6 +65,32 @@ func Test_newConfig(t *testing.T) {
...
```
sgnn7, line 93: I still don't see a test here that includes special/utf chars. The one that uses `==` isn't adequate enough. See if you can add a test using chars like [these](https://github.com/cyberark/conjur-api-python3/blob/master/test/test_config/policies/variables.yml#L6).

internal/plugin/connectors/http/generic/config.go
BradleyBoutcher: Goooood catch

internal/plugin/connectors/http/generic/config.go
BradleyBoutcher: Totally agree, fixed it up 👍

internal/plugin/connectors/http/generic/config.go
BradleyBoutcher: I'm not sure that's accurate, as we've already performed our unmarshalling at this point. Here we're converting the strings to templates, so I think something like "stringToTemplate" would be best.

internal/plugin/connectors/http/generic/README.md
```diff
@@ -103,6 +108,41 @@ powerful transformation features.  You can use `printf` for formatting and
...
 above for detailed information on these and other features.
 
+### `queryParams`
+
+Like `headers`, this is another key section. The `queryParams` section
+is used to generate a query string, which is appended to your existing URL
+without replacing any existing query parameters.
+
+The _keys_ of the queryParams are defined by the yaml keys.  In the examples
+above, these query parameter key is `location`.
```
izgeri, line 118:
```suggestion
above, the query parameter key is `location`.
```

internal/plugin/connectors/http/generic/README.md
```diff
@@ -103,6 +108,41 @@ powerful transformation features.  You can use `printf` for formatting and
...
+```
+
+You can refer to your credentials in this template using the credential name
+preceded by a `.` (eg, `.address` will refer to the credential
+`address`). At runtime, Secretless will replace these
+credential references with your real credentials.
+
+As you can see in the Basic auth example, the `text/template` package has
+powerful transformation features.  You can use `printf` for formatting and
+compose functions using pipes `|`.  See the text template package docs linked
```
izgeri, line 143:
```suggestion
you can compose functions using pipes `|`.  See the text template package docs linked
```
not having the "you can" at the start here led me to try to understand "formatting and compose functions" as a concept

PR #1299, opened 2020-07-07
Add Docker Registry generic connector

JakeQuilty:
Adds a generic connector for connecting to a Docker Registry and updates the README with details

### DIFF
--- a/examples/generic_connector_configs/README.md
+++ b/examples/generic_connector_configs/README.md
@@ -2,6 +2,7 @@
 
 ## Table of Contents
 * [Sample Configurations](#sample-configurations)
+  * [Docker Registry API](#docker-registry-api)
   * [Elasticsearch API](#elasticsearch-api)
   * [GitHub API](#github-api)
   * [OAuth 2.0 API](#oauth-20-api)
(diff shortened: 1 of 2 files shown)


examples/generic_connector_configs/README.md
```diff
@@ -29,9 +30,47 @@ target’s CA to Secretless’ trusted certificate pool.
...
+
+#### Example Usage
+<details>
+  <summary><b>How to use this connector locally</b></summary>
+  <ol>
+    <li>Set up a <a href="https://docs.docker.com/registry/deploying/">
+    local Registry</a> or use one from <a href="https://hub.docker.com">Dockerhub</a></li>
+    <li>Make a request to the Registry API to get the <a href="https://docs.docker.com/registry/spec/auth/oauth/">
+    OAuth2 Token</a>.</li>
+    <li>Run Secretless locally</li>
```
mdodell, line 60: You should add a step to add a token to the local keychain. If you pull master, you can copy the exact language/HTML from the other config connectors.

izgeri:
@JakeQuilty please be sure to use the "Connected to #1271" syntax to make sure this PR is connected to the issue in GitHub - I'd also be glad if you could use the PR template, instead of deleting it

@mdodell please delete branches after merging so that we don't keep accumulating branches in github :)

Issue #1265, opened 2020-06-22
There is an example generic HTTP connector config for Facebook


izgeri:
## Use case
I am a developer building an app that has to communicate with the Facebook API

## Additional details
TODO: before starting work, add details on how authentication works for this webservice, to determine whether implementing it as a generic HTTP connector is feasible.

## Acceptance criteria
- [ ] There is a generic HTTP connector config available in the [example generic HTTP connector configs](https://github.com/cyberark/secretless-broker/tree/master/examples/generic_connector_configs) for connecting to the Facebook API via Secretless

JakeQuilty:
## Multiple API's
- [ ] 
- [ ] 
- [ ] 

Live Video API requires [App Review](https://developers.facebook.com/docs/app-review)

JakeQuilty:
All of the API's authenticate in the same way, but the set up of the demo app sides are different. I focused on an example with Instagram, and mentioned that the rest of Facebook's tutorials could be used instead.

PR #1301, opened 2020-07-13
Add Facebook API example connector

JakeQuilty:
### What does this PR do?
This PR adds a generic connector for the Facebook API, with an example for demoing with Instagram API. All of the API's authenticate in the same way, but the set up of the demo app sides are different. I focused on an example with Instagram, and mentioned that the rest of Facebook's tutorials could be used instead.

Connected to #https://github.com/cyberark/secretless-broker/issues/1265

#### Change log

#### Test coverage

#### Documentation

#### ~~(For releases only) Manual tests~~

### DIFF
--- a/examples/generic_connector_configs/README.md
+++ b/examples/generic_connector_configs/README.md
@@ -6,6 +6,7 @@
   * [Docker Registry API](#docker-registry-api)
   * [Dropbox API](#dropbox-api)
   * [Elasticsearch API](#elasticsearch-api)
+  * [Facebook API](#facebook-api)
   * [GitHub API](#github-api)
   * [Mailchimp API](#mailchimp-api)
   * [OAuth 2.0 API](#oauth-20-api)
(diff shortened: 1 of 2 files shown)


JakeQuilty:
All of the API's authenticate in the same way, but the set up of the demo app sides are different. I focused on an example with Instagram, and mentioned that the rest of Facebook's tutorials could be used instead.

examples/generic_connector_configs/README.md
```diff
@@ -5,6 +5,7 @@
   * [Docker Registry API](#docker-registry-api)
   * [Elasticsearch API](#elasticsearch-api)
   * [GitHub API](#github-api)
+  * [Facebook API](#facebook-api)
```
mdodell, line 8: Should we call this the `Instagram API?`

examples/generic_connector_configs/README.md
```diff
@@ -150,6 +151,49 @@ The configuration file for the GitHub API can be found at [github_secretless.yml
 
 ___
 
+### Facebook API
```
mdodell, line 154: See Instagram comment.

examples/generic_connector_configs/README.md
```diff
@@ -150,6 +151,49 @@ The configuration file for the GitHub API can be found at [github_secretless.yml
...
+This example focuses on [Instagram](https://developers.facebook.com/docs/instagram), but can be
+ used to interact with [Facebook's many API's](https://developers.facebook.com/docs/instagram).
```
mdodell, line 156: This is the same link as the one above. Can we be more explicit about what APIs from Facebook this supports?

examples/generic_connector_configs/README.md
JakeQuilty: I thought about calling it that, but I double checked and this connector will work with all of the Facebook API's listed [here](https://developers.facebook.com/docs/). I made sure to point out in the description that it could be adjusted for all of them, but I was focussing on Instagram for the example. I think it keeping it as Facebook, will help people recognize that without having to dig too deep

examples/generic_connector_configs/README.md
JakeQuilty: Attached the wrong link, good catch!

Issue #1297, opened 2020-07-06
Generic HTTP Config supports OAuth1


JakeQuilty:
## Is your feature request related to a problem? Please describe.

Currently Secretless Broker does not support any of the 3 methods of authenticating with OAuth1

## Problems with OAuth 1.0a and Secretless

OAuth 1.0a requires 7 different parameters to authenticate:
```
Authorization:  oauth_consumer_key="0685bd9184jfhq22",
                oauth_token="ad180jjd733klru7",
                oauth_signature_method="HMAC-SHA1",
                oauth_signature="wOJIO9A2W5mFwDgiDvZbTSMK%2FPY%3D",
                oauth_timestamp="137131200",
                oauth_nonce="4572616e48616d6d65724c61686176",
                oauth_version="1.0"
```

`consumer_key` - Constant secret
` nonce` - Dynamic, created specifically for request
`signature` - Dynamic, [generated](https://developer.twitter.com/en/docs/basics/authentication/oauth-1-0a/creating-a-signature) from all other parameters
`timestamp` - Dynamic, timestamp of request and will be denied if too old when request is sent
`token` - Constant secret
`signature_method` - Dynamic
`version` - Dynamic

There are [3 ways](https://oauth.net/core/1.0/#consumer_req_param) to pass these parameters in the HTTP request:
```
1. In the HTTP Authorization header as defined in OAuth HTTP Authorization Scheme.
2. As the HTTP POST request body with a content-type of application/x-www-form-urlencoded.
3. Added to the URLs in the query part (as defined by RFC3986 section 3).
```

### Method 1
Only 2 of the 7 parameters are constant secrets for Secretless to inject, the rest need to be created by the application when making the request. 
The main problem with OAuth 1.0a and Secretless is that Secretless currently replaces the whole `Authorization` line in the header, with no way to "save" the other parameters that are passed.

Example:

Request:
```
http_proxy=localhost:8071 curl -k --request POST \
  --url 'test:4443' \
  --header 'authorization: OAuth oauth_consumer_key="CONSUMER_API_KEY", oauth_nonce="OAUTH_NONCE", oauth_signature="OAUTH_SIGNATURE", oauth_signature_method="HMAC-SHA1", oauth_timestamp="OAUTH_TIMESTAMP", oauth_token="ACCESS_TOKEN", oauth_version="1.0"' \
```

Secretless generic config:
```
...
config:
      headers:
        Authorization: ApiKey test
      forceSSL: true
[...]

PR #1321, opened 2020-07-30
Added OAuth1 functionality to the generic HTTP connector

mdodell:
### What does this PR do?
- Added OAuth1 functionaly to the generic HTTP connector

Connected to #1297 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   in the secretless configuration file, under `config`. This allows the
   construction of a query string which can have credentials injected
   as needed. [#1290](https://github.com/cyberark/secretless-broker/issues/1290)
+- Generic HTTP connector now supports `oauth1` as a configurable section in the
+  secretless configuration file, under `config`. This allows the construction of
+  a header for an OAuth 1.0 request. The OAuth 1.0 feature currently only supports
+  HMAC-SHA1, but there is an [issue](https://github.com/cyberark/secretless-broker/issues/1324)
+  logged to support other hashing methods. [#1297](https://github.com/cyberark/secretless-broker/issues/1297)
 
 ## [1.6.0] - 2020-05-04
 
--- a/examples/generic_connector_configs/twitter_secretless.yml
+++ b/examples/generic_connector_configs/twitter_secretless.yml
@@ -14,3 +14,28 @@ services:
       forceSSL: true
       authenticateURLsMatching:
         - ^http[s]*\:\/\/api\.twitter\.com*
+  twitter-oauth1:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:8061
+    credentials:
+      consumer_key:
+        from: keychain
+        get: service#twitter/consumer-key
+      consumer_secret:
+        from: keychain
+        get: service#twitter/consumer-secret
+      token:
+        from: keychain
+        get: service#twitter/token
+      token_secret:
+        from: keychain
+        get: service#twitter/token-secret
+    config:
+      forceSSL: true
+      oauth1:
+        consumer_key: "{{ .consumer_key }}"
+        consumer_secret: "{{ .consumer_secret }}"
+        token: "{{ .token }}"
+        token_secret: "{{ .token_secret }}"
+      authenticateURLsMatching:
+        - ^http[s]*\:\/\/api\.twitter\.com*
--- a/internal/plugin/connectors/http/generic/README.md
+++ b/internal/plugin/connectors/http/generic/README.md
@@ -34,7 +34,7 @@ services:
     connector: generic_http
     listenOn: tcp://0.0.0.0:8080
     credentials:
-      apikey: 
+      apikey:
         from: conjur
         get: my-services-api-key  # the id of your API key within conjur
     config:
@@ -61,7 +61,7 @@ services:
     connector: generic_http
     listenOn: tcp://0.0.0.0:8080
     credentials:
-      username: 
+      username:
         from: conjur
         get: someuser
       password:
@@ -123,13 +123,13 @@ template](https://golang.org/pkg/text/template/), as defined in the
 
 In the above example, let us say that your request URL looks like the following,
 
-```
+```http
 http://anything.com/foo?fruit=apple
 ```
 
 After proxying through secretless, your request URL would look like the following,
 
-```
+```http
 http://anything.com/foo?location=valueofaddress&fruit=apple
 ```
 
@@ -143,6 +143,69 @@ powerful transformation features.  You can use `printf` for formatting and
 you can compose functions using pipes `|`.  See the text template package docs linked
 above for detailed information on these and other features.
 
+### `oauth1`
+
+Like `headers` this is another key section. The `oauth1` section is used to generate
+an OAuth1 `Authorization` header.
+
+**Note: Declaring an `Authorization` header in the `config`
+ while `oauth1` is present will throw an error:
+ `authorization header already exists, cannot override header`**
+
+There are four required _keys_ for the `oauth1` section which are as follows:
+
+1. `consumer_key` - A value used by the consumer to identify itself to
+   the service provider.
+1. `consumer_secret` - A secret used by the consumer to establish ownership
+   of the consumer key.
+1. `token` - A value used by the consumer to gain access to the protected
+   resources on behalf of the user, instead of using the user’s service
+   provider credentials.
+1. `token_secret` - A secret used by the consumer to establish ownership of
+   a given token.
+
+The oauth1 _values_ are defined using a [Go text
+template](https://golang.org/pkg/text/template/), as defined in the
+`text/template` package.
+
+You can refer to your credentials in this template using the credential name
+preceded by a `.` (e.g. `.consumer_key` and `.token` refer to the credentials
+`consumer_key` and `token`). At runtime, Secretless will replace these
+credential references with your real credentials.
+
+For instance:
+
+```yaml
+
+version: 2
+services:
+  oauth1-service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:8080
+    credentials:
+      consumer_key:
+        from: conjur
+        get: somekey
+      consumer_secret:
+        from: conjur
+        get: someconsumersecret
+      token:
+        from: conjur
+        get: sometoken
+      token_secret:
+        from: conjur
+        get: sometokensecret
+    config:
+      oauth1:
+        consumer_key: "{{ .consumer_key }}"
+        consumer_secret: "{{ .consumer_secret }}"
+        token: "{{ .token }}"
+        token_secret: "{{ .token_secret }}"
+      forceSSL: true
+      authenticateURLsMatching:
+        - ^http
+```
+
 #### `credentialValidations`
 
 This section lets you use regular expressions to define validations for the
@@ -150,7 +213,7 @@ header values.
 
 For example, in the first example above, the line:
 
-```
+```yaml
 apikey: '^[A-Z0-9]+$'
 ```
 
@@ -200,7 +263,7 @@ Let's take a look at the Basic auth connector to see how this works:
 // Taken from:
 // internal/plugin/connectors/http/basicauth/plugin.go
 //
-// Also note: 
+// Also note:
 // "NewConnectorConstructor" and "ConfigYAML" are defined in:
 // /internal/plugin/connectors/http/generic/external_api.go
 
@@ -214,7 +277,7 @@ newConnector, err := generichttp.NewConnectorConstructor(
     },
   },
 )
-``` 
+```
 
 Note how this code exactly mirrors the yaml from the end-user example.  Please
 refer to those docs above for an explanation of the `CredentialValidations` and
--- a/internal/plugin/connectors/http/generic/config.go
+++ b/internal/plugin/connectors/http/generic/config.go
@@ -15,6 +15,7 @@ import (
 type config struct {
 	CredentialPatterns map[string]*regexp.Regexp
 	Headers            map[string]*template.Template
+	OAuth1Secrets      map[string]*template.Template
 	QueryParams        map[string]*template.Template
 	ForceSSL           bool
 }
@@ -89,6 +90,7 @@ func newConfig(cfgYAML *ConfigYAML) (*config, error) {
 
 	cfg.Headers, errs = stringsToTemplates(cfgYAML.Headers, errs)
 	cfg.QueryParams, errs = stringsToTemplates(cfgYAML.QueryParams, errs)
+	cfg.OAuth1Secrets, errs = stringsToTemplates(cfgYAML.OAuth1Secrets, errs)
 
 	if err := errs.Filter(); err != nil {
 		return nil, err
--- a/internal/plugin/connectors/http/generic/connector.go
+++ b/internal/plugin/connectors/http/generic/connector.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	gohttp "net/http"
 
+	oauth1 "github.com/cyberark/secretless-broker/internal/plugin/connectors/http/generic/oauth/v1"
 	"github.com/cyberark/secretless-broker/pkg/secretless/log"
 	"github.com/cyberark/secretless-broker/pkg/secretless/plugin/connector"
 )
@@ -12,6 +13,49 @@ import (
 type Connector struct {
 	logger log.Logger
 	config *config
+}
+
+func addOAuth1Header(c *Connector, credentialsByID connector.CredentialValuesByID, r *gohttp.Request) error {
+	// Doesn't check for error. Instead we check to see if
+	// any values exist in the oauth1 key in the config.
+	oAuth1Params, err := renderTemplates(c.config.OAuth1Secrets, credentialsByID)
+	if err != nil {
+		return fmt.Errorf("failed to render oauth1 params: %s", err)
+	}
+
+	if len(oAuth1Params) > 0 {
+		oauthHeader, err := oauth1.CreateOAuth1Header(oAuth1Params, r)
+		if err != nil {
+			return fmt.Errorf("failed to create oAuth1 'Authorization' header: %s", err)
+		}
+
+		if len(r.Header.Get("Authorization")) > 0 {
+			return fmt.Errorf("authorization header already exists, cannot override header")
+		}
+
+		r.Header.Set("Authorization", oauthHeader)
+	}
+	return nil
+}
+
+func addQueryParams(c *Connector, credentialsByID connector.CredentialValuesByID, r *gohttp.Request) error {
+	queryParams, err := renderTemplates(c.config.QueryParams, credentialsByID)
+	if err != nil {
+		return fmt.Errorf("failed to render query params: %s", err)
+	}
+	r.URL.RawQuery = appendQueryParams(*r.URL, queryParams)
+	return nil
+}
+
+func addHeaders(c *Connector, credentialsByID connector.CredentialValuesByID, r *gohttp.Request) error {
+	headers, err := renderTemplates(c.config.Headers, credentialsByID)
+	if err != nil {
+		return fmt.Errorf("failed to render headers: %s", err)
+	}
+	for headerName, headerVal := range headers {
+		r.Header.Set(headerName, headerVal)
+	}
+	return nil
 }
 
 // Connect implements the http.Connector func signature.
@@ -30,20 +74,19 @@ func (c *Connector) Connect(
 	}
 
 	// Add configured headers to request
-	headers, err := renderTemplates(c.config.Headers, credentialsByID)
-	if err != nil {
-		return fmt.Errorf("failed to render headers: %s", err)
-	}
-	for headerName, headerVal := range headers {
-		r.Header.Set(headerName, headerVal)
+	if err := addHeaders(c, credentialsByID, r); err != nil {
+		return err
 	}
 
 	// Add configured params to request
-	params, err := renderTemplates(c.config.QueryParams, credentialsByID)
-	if err != nil {
-		return fmt.Errorf("failed to render query params: %s", err)
+	if err := addQueryParams(c, credentialsByID, r); err != nil {
+		return err
 	}
-	r.URL.RawQuery = appendQueryParams(*r.URL, params)
+
+	// Add oAuth1 Authorization Header to the request
+	if err := addOAuth1Header(c, credentialsByID, r); err != nil {
+		return err
+	}
 
 	return nil
 }
--- a/internal/plugin/connectors/http/generic/external_api.go
+++ b/internal/plugin/connectors/http/generic/external_api.go
@@ -19,9 +19,10 @@ import (
 //         .yml file.  That is, we convert from .yml --> ConfigYAML --> config.
 type ConfigYAML struct {
 	CredentialValidations map[string]string `yaml:"credentialValidations"`
-	Headers               map[string]string `yaml:"headers"`
-	QueryParams           map[string]string `yaml:"queryParams"`
 	ForceSSL              bool              `yaml:"forceSSL"`
+	Headers               map[string]string `yaml:"headers"`
+	OAuth1Secrets         map[string]string `yaml:"oauth1"`
+	QueryParams           map[string]string `yaml:"queryParams"`
 }
 
 // NewConfigYAML takes the raw cfgBytes and unmarshals them into a ConfigYAML
--- a/internal/plugin/connectors/http/generic/oauth/v1/protocol.go
+++ b/internal/plugin/connectors/http/generic/oauth/v1/protocol.go
@@ -1,0 +1,262 @@
+package oauth1protocol
+
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha1"
+	"encoding/base64"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	gohttp "net/http"
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+	"text/template"
+	"time"
+)
+
+// OAuth1 is a struct of all OAuth1 specific parameters
+type OAuth1 struct {
+	ConsumerKey     string
+	ConsumerSecret  string
+	Nonce           string
+	Signature       string
+	SignatureMethod string
+	TimeStamp       string
+	Token           string
+	TokenSecret     string
+	Version         string
+}
+
+// OAuth Header Key Names
+const (
+	oauthConsumerKey     = "oauth_consumer_key"
+	oauthNonce           = "oauth_nonce"
+	oauthSignatureMethod = "oauth_signature_method"
+	oauthTimestamp       = "oauth_timestamp"
+	oauthToken           = "oauth_token"
+	oauthVersion         = "oauth_version"
+)
+
+// OAuth Header Variables
+const (
+	// Nonce doesn't have a specified charset or length
+	// Reference: https://tools.ietf.org/html/rfc5849#section-3.3
+	nonceCharset = "abcdefghijklmnopqrstuvwxyz" +
+		"ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"0123456789"
+	nonceLength = 16
+	// Currently, this feature supports HMAC-SHA1, but there is an issue logged to
+	// support more methods.
+	// Reference: https://github.com/cyberark/secretless-broker/issues/1324
+	oauthSignatureMethodType = "HMAC-SHA1"
+	oauthVersionNumber       = "1.0"
+)
+
+// Required Config YAML Variables
+const (
+	consumerKey    = "consumer_key"
+	consumerSecret = "consumer_secret"
+	token          = "token"
+	tokenSecret    = "token_secret"
+)
+
+var requiredConfigParams = []string{
+	consumerKey,
+	consumerSecret,
+	token,
+	tokenSecret,
+}
+
+func generateNonce(length int, charset string) string {
+	seededRand := rand.New(
+		rand.NewSource(time.Now().UnixNano()))
+
+	randomChars := make([]byte, length)
+	for index := range randomChars {
+		randomChars[index] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(randomChars)
+}
+
+// checkRequiredOAuthParams returns an error if a key from
+// "requiredConfigParams" was not included in the supplied config
+// returns the first missing key found
+func checkRequiredOAuthParams(params map[string]string) error {
+	for _, requiredParam := range requiredConfigParams {
+		if len(params[requiredParam]) < 1 {
+			return fmt.Errorf("required oAuth1 parameter '%s' not found", requiredParam)
+		}
+	}
+	return nil
+}
+
+// extractOAuthParams extracts the required oAuth1 parameters from
+// the config and assigns the other required paramters
+func extractOAuthParams(params map[string]string) OAuth1 {
+	oauth := OAuth1{
+		ConsumerKey:     params[consumerKey],
+		ConsumerSecret:  params[consumerSecret],
+		Nonce:           generateNonce(nonceLength, nonceCharset),
+		SignatureMethod: oauthSignatureMethodType,
+		Token:           params[token],
+		TokenSecret:     params[tokenSecret],
+		TimeStamp:       strconv.Itoa(int(time.Now().Unix())),
+		Version:         oauthVersionNumber,
+	}
+
+	return oauth
+}
+
+// encodeURI percent encodes a string and changes encoding of <SPACE> from "+" to "%20"
+// Reference: https://tools.ietf.org/html/rfc5849#section-3.6
+func encodeURI(stringToEncode string) string {
+	return strings.ReplaceAll(url.QueryEscape(stringToEncode), "+", "%20")
+}
+
+// collectParameters creates a map of body and query parameters by percent encoding all
+// keys and values and sorting values post-encoding
+func collectParameters(oauth1 OAuth1, r *gohttp.Request) map[string][]string {
+	// Create two copies of the request body
+	// Since r.Body is io.ReadCloser the body is empty after reading.
+	// Need to replace r.Body after r.ParseForm() dumps the body contents
+	buf, _ := ioutil.ReadAll(r.Body)
+	originalBody := ioutil.NopCloser(bytes.NewBuffer(buf))
+	copyBody := ioutil.NopCloser(bytes.NewBuffer(buf))
+	r.Body = originalBody
+
+	paramMap := make(map[string][]string, 0)
+	// get query and body params and URL encode keys and
+	// values sort values after encoding
+	r.ParseForm()
+	for paramKey, paramValues := range r.Form {
+		encodedKey := encodeURI(paramKey)
+		paramMap[encodedKey] = paramValues
+
+		// Keys with multiple values require the values
+		// to be sorted post-percent encoding
+		for valueIndex, value := range paramMap[encodedKey] {
+			encodedValue := encodeURI(value)
+			paramMap[encodedKey][valueIndex] = encodedValue
+		}
+		sort.Strings(paramMap[encodedKey])
+	}
+	r.Body = copyBody
+
+	// these keys are already URL safe and don't need sorted(1 value)
+	paramMap[oauthConsumerKey] = []string{encodeURI(oauth1.ConsumerKey)}
+	paramMap[oauthNonce] = []string{encodeURI(oauth1.Nonce)}
+	paramMap[oauthSignatureMethod] = []string{encodeURI(oauth1.SignatureMethod)}
+	paramMap[oauthTimestamp] = []string{encodeURI(oauth1.TimeStamp)}
+	paramMap[oauthToken] = []string{encodeURI(oauth1.Token)}
+	paramMap[oauthVersion] = []string{encodeURI(oauth1.Version)}
+
+	return paramMap
+}
+
+// generateParameterString creates a Parameter String from
+// sorting an encoded map of parameters and values
+// Reference: https://tools.ietf.org/html/rfc5849#section-3.5
+func generateParameterString(paramMap map[string][]string) string {
+	var parameterString strings.Builder
+
+	// make map of keys and sort to append lexicographically later
+	keys := make([]string, 0, len(paramMap))
+	for key := range paramMap {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	// create string
+	for keyIndex, key := range keys {
+		// Write keys and value pairs to `parameterString`
+		// ex: key0=value&key1=value&key2=value
+		for valueIndex, value := range paramMap[key] {
+			// Do not add '&' on first value
+			if keyIndex > 0 || valueIndex > 0 {
+				parameterString.WriteString("&")
+			}
+			parameterString.WriteString(key + "=" + value)
+		}
+	}
+
+	return parameterString.String()
+}
+
+// generateBaseString combines parameters to create the
+// pre-hash base string for the signature
+// Reference: https://tools.ietf.org/html/rfc5849#section-3.4.1
+func generateBaseString(r *gohttp.Request, paramString string) string {
+	method := strings.ToUpper(r.Method)
+	urlString := fmt.Sprintf("%s://%s%s", r.URL.Scheme, r.URL.Host, r.URL.Path)
+	return fmt.Sprintf("%s&%s&%s", method, encodeURI(urlString), encodeURI(paramString))
+}
+
+// generateSigningKey concats the consumer secret and
+// token secret with an ampersand, and encodes them
+// Reference: https://tools.ietf.org/html/rfc5849#section-3.4
+func generateSigningKey(oauth1 OAuth1) string {
+	return fmt.Sprintf("%s&%s", encodeURI(oauth1.ConsumerSecret), encodeURI(oauth1.TokenSecret))
+}
+
+// hashSignature creates a HMAC-SHA1 hash
+// Reference: https://tools.ietf.org/html/rfc5849#section-3.4.2
+func hashSignature(signingKey string, baseString string) string {
+	h := hmac.New(sha1.New, []byte(signingKey))
+	h.Write([]byte(baseString))
+	return encodeURI(base64.StdEncoding.EncodeToString(h.Sum(nil)))
+}
+
+// constructOAuthString creates the Authorization header
+// that is needed for the final request
+// Reference: https://tools.ietf.org/html/rfc5849#section-3.5.1
+func constructOAuthString(oauth1 OAuth1) (string, error) {
+	tmpl, err := template.New("oauth").Parse("OAuth " +
+		"oauth_consumer_key=\"{{.ConsumerKey}}\", " +
+		"oauth_nonce=\"{{.Nonce}}\", " +
+		"oauth_signature=\"{{.Signature}}\", " +
+		"oauth_signature_method=\"{{.SignatureMethod}}\", " +
+		"oauth_timestamp=\"{{.TimeStamp}}\", " +
+		"oauth_token=\"{{.Token}}\", " +
+		"oauth_version=\"{{.Version}}\"")
+	if err != nil {
+		return "", fmt.Errorf("could not create the OAuth template: %s", err)
+	}
+
+	var tpl bytes.Buffer
+	err = tmpl.Execute(&tpl, oauth1)
+	if err != nil {
+		return "", fmt.Errorf("could not turn the OAuth template into a string: %s", err)
+	}
+
+	return tpl.String(), nil
+}
+
+// CreateOAuth1Header calls a series of methods to construct
+// the oAuth1 'Authorization' Header String
+func CreateOAuth1Header(params map[string]string, r *gohttp.Request) (string, error) {
+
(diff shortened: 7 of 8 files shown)


examples/generic_connector_configs/twitter_secretless.yml
```diff
@@ -14,3 +14,28 @@ services:
...
+        get: service#twitter/consumer-secret
+      token:
+        from: keychain
+        get: service#twitter/token
+      token_secret:
+        from: keychain
+        get: service#twitter/token-secret
+    config:
+      forceSSL: true
+      oauth1Params:
```
sgnn7, line 35: `Params` seems superfluous since we're in `config` section already

examples/generic_connector_configs/twitter_secretless.yml
```diff
@@ -14,3 +14,28 @@ services:
...
+        oauth_consumer_key: "{{ .consumer_key }}"
+        oauth_consumer_secret: "{{ .consumer_secret }}"
+        oauth_token: "{{ .token }}"
+        oauth_token_secret: "{{ .token_secret }}"
```
sgnn7, line 39: `oauth_` prefix seems superfluous since we're already in oauth1 section of config

internal/plugin/connectors/http/generic/config.go
```diff
@@ -15,6 +15,7 @@ import (
 type config struct {
 	CredentialPatterns map[string]*regexp.Regexp
 	Headers            map[string]*template.Template
+	OAuth1             map[string]*template.Template
```
sgnn7, line 18: This needs a touch better name. what is `Oauth1`? Object? Auth info? Auth data? etc

internal/plugin/connectors/http/generic/connector.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	gohttp "net/http"
 
+	oauth "github.com/cyberark/secretless-broker/internal/plugin/connectors/http/generic/oauthprotocol"
```
sgnn7, line 7: this line should be together with other packages from secretless broker below

internal/plugin/connectors/http/generic/connector.go
```diff
@@ -39,11 +41,22 @@ func (c *Connector) Connect(
...
-	params, err := renderTemplates(c.config.QueryParams, credentialsByID)
+	queryParams, err := renderTemplates(c.config.QueryParams, credentialsByID)
 	if err != nil {
 		return fmt.Errorf("failed to render query params: %s", err)
 	}
-	r.URL.RawQuery = appendQueryParams(*r.URL, params)
+	r.URL.RawQuery = appendQueryParams(*r.URL, queryParams)
+
+	// Add OAuth1 Authorization Header to the request
+	oauth1params, err := renderTemplates(c.config.OAuth1, credentialsByID)
```
sgnn7, line 51:
```suggestion
	oauth1Params, err := renderTemplates(c.config.OAuth1, credentialsByID)
```

sgnn7: This whole block should probably only happen if oauth info is specified but we run it every time

internal/plugin/connectors/http/generic/connector.go
```diff
@@ -39,11 +41,22 @@ func (c *Connector) Connect(
...
+	if err != nil {
+		return fmt.Errorf("failed to render OAuth1 params: %s", err)
+	}
+	oauthHeader, err := oauth.CreateOAuth1Header(oauth1params, r)
+	if err != nil {
+		return fmt.Errorf("failed to create OAuth1 'Authorization' header: %s", err)
+	}
```
sgnn7, line 58: It almost seems like that `renderTemplates` and `CreateOAuth1Header` should be in a separate smaller function

internal/plugin/connectors/http/generic/external_api.go
```diff
@@ -22,6 +22,7 @@ type ConfigYAML struct {
 	Headers               map[string]string `yaml:"headers"`
 	QueryParams           map[string]string `yaml:"queryParams"`
 	ForceSSL              bool              `yaml:"forceSSL"`
+	Oauth1Params          map[string]string `yaml:"oauth1Params"`
```
sgnn7, line 25: This struct should have the fields ordered alphabetically

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
+package oauth1protocol
```
sgnn7, line 1: I think maybe just `oauth/v1/protocol.go` might be a better layout of this. Thoughts?

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	oauth_signature_method = "oauth_signature_method"
+	oauth_version          = "oauth_version"
+	oauth_timestamp        = "oauth_timestamp"
+
+	oauth_version_number        = "1.0"
+	oauth_signature_method_type = "HMAC-SHA1"
+	nonce_length                = 16
+)
+
+const charset = "abcdefghijklmnopqrstuvwxyz" +
```
sgnn7, line 49: What is this charset? Probably the name should reflect that.

test/connector/http/generic/secretless.yml
```diff
@@ -8,6 +8,8 @@ services:
     credentials:
       username: someuser
       password: testpassword
+    oauthparams:
```
sgnn7, line 11: isn't this `oauth1Params`?

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	"ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+	"0123456789"
+
+func stringWithCharset(length int, charset string) string {
```
sgnn7, line 53: This method is not intuitive as to what it's doing. we should have both a comment and a more descriptive name on it I think.

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	return stringWithCharset(length, charset)
+}
+
+func checkRequiredOAuthParams(params map[string]string) error {
+	var requiredParams = []string{
+		oauth_consumer_key,
+		oauth_consumer_secret,
+		oauth_token,
+		oauth_token_secret,
+	}
```
sgnn7, line 74: Can this static var be a const or moved up top?

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	var oauth OAuth1
+
+	oauth.ConsumerKey = params[oauth_consumer_key]
+	oauth.ConsumerSecret = params[oauth_consumer_secret]
+	oauth.Token = params[oauth_token]
+	oauth.TokenSecret = params[oauth_token_secret]
+	oauth.TimeStamp = strconv.Itoa(int(time.Now().Unix()))
+	oauth.Version = oauth_version_number
+	oauth.SignatureMethod = oauth_signature_method_type
+	oauth.Nonce = generateNonce(nonce_length)
```
sgnn7, line 97: This seems like a place where a class constructor would fit better:
```go
oauth := OAuth1{
  ConsumerKey: ...,
 ...
}
```

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return oauth
+}
+
+func encodeURI(stringToEncode string) string {
```
sgnn7, line 102: This needs a comment as to why we're encoding an escaped string

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	return strings.ReplaceAll(url.QueryEscape(stringToEncode), "+", "%20")
+}
+
+// Create a Parameter String from the required OAuth params and query & body params
+// https://tools.ietf.org/html/rfc5849#section-3.5
+func generateParameterString(oauth1 OAuth1, r *gohttp.Request) string {
+	// create map of all parameters
+	buf, _ := ioutil.ReadAll(r.Body)
+	originalBody := ioutil.NopCloser(bytes.NewBuffer(buf))
+	copyBody := ioutil.NopCloser(bytes.NewBuffer(buf))
```
sgnn7, line 112: This needs at least a comment as to why we have a copy and original

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	r.Body = originalBody
```
sgnn7, line 113: Not sure what's happening here - you read the body then reassign it back.. and then later we overwrite it with the copy?

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+
+	var parameterString bytes.Buffer
+	// create map of all parameters
+	paramMap := make(map[string][]string, 0)
```
sgnn7, line 117: Why do we need a `string -> []string` for params? Wouldn't a `string -> string` map be enough?

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	r.Body = copyBody
+
+	// these keys are already URL safe
+	paramMap[oauth_consumer_key] = []string{oauth1.ConsumerKey}
+	paramMap[oauth_nonce] = []string{oauth1.Nonce}
+	paramMap[oauth_signature_method] = []string{oauth1.SignatureMethod}
+	paramMap[oauth_timestamp] = []string{oauth1.TimeStamp}
+	paramMap[oauth_token] = []string{oauth1.Token}
+	paramMap[oauth_version] = []string{oauth1.Version}
```
sgnn7, line 132: Ditto not sure why this needs to be an array of strings

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// make map of keys and sort to append lexicographically later
+	keys := make([]string, 0, len(paramMap))
+	for k := range paramMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	// create string
+	for indK, key := range keys {
```
sgnn7, line 142:
```suggestion
	for keyIndex, key := range keys {
```

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+		// URL encode all values in all keys (most should be 1, but this double for loop is still gross)
+		for ind, value := range paramMap[key] {
```
sgnn7, line 144:
```suggestion
		for valueIndex, value := range paramMap[key] {
```

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+			encodedValue := strings.ReplaceAll(url.QueryEscape(value), "+", "%20")
```
sgnn7, line 145: You already have a dedicated function for this

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+			paramMap[key][ind] = encodedValue
+		}
+
+		// sort the array of URL encoded values in each key
+		sort.Strings(paramMap[key])
+
+		for indV, value := range paramMap[key] {
```
sgnn7, line 152: This loop needs at least some comments - not sure why this is needed.

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+			keyValuePair := fmt.Sprintf("%s=%s", key, value)
+			parameterString.WriteString(keyValuePair)
+		}
+	}
+
+	return parameterString.String()
+}
+
+func generateBaseString(r *gohttp.Request, paramString string) string {
```
sgnn7, line 165: This needs a comment description at the very least

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+	return encodeURI(base64.StdEncoding.EncodeToString(h.Sum(nil)))
+}
+
+func constructOAuthString(oauth1 OAuth1) string {
+	return fmt.Sprintf("OAuth oauth_consumer_key=\"%s\", oauth_nonce=\"%s\", "+
+		"oauth_signature=\"%s\", oauth_signature_method=\"%s\", "+
+		"oauth_timestamp=\"%s\", oauth_token=\"%s\", oauth_version=\"%s\"",
+		oauth1.ConsumerKey, oauth1.Nonce, oauth1.Signature, oauth_signature_method_type,
+		oauth1.TimeStamp, oauth1.Token, oauth1.Version)
+}
```
sgnn7, line 191: Can we use a template for this? This seems really brittle as-is.

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
```diff
@@ -0,0 +1,214 @@
...
+
+	oauth1 := extractOAuthParams(params)
+
+	paramString := generateParameterString(oauth1, r)
+	baseString := generateBaseString(r, paramString)
+	signingKey := generateSigningKey(oauth1)
+
+	oauth1.Signature = hashSignature(signingKey, baseString)
+
+	oauthString := constructOAuthString(oauth1)
```
sgnn7, line 211: Something tells me there's more ways this can error out outside of param extraction but for now not a big deal

internal/plugin/connectors/http/generic/oauthprotocol/protocol_test.go
```diff
@@ -0,0 +1,394 @@
...
+				Signature:       "oauth_signature",
+				SignatureMethod: "HMAC-SHA1",
+				TimeStamp:       "",
+				Token:           "oauth_token",
+				TokenSecret:     "oauth_token_secret",
+				Version:         "1.0",
+			},
+		},
+	}
+	for _, tt := range tests {
```
sgnn7, line 48: You are running a single test using a matrix test setup. I think this can just be a single inline test.

internal/plugin/connectors/http/generic/oauthprotocol/protocol_test.go
```diff
@@ -0,0 +1,394 @@
...
+	}{
+		{
+			name: "Generates a 16 length nonce",
+			args: args{
+				length: 16,
+			},
+			want: 16,
+		},
+	}
+	for _, tt := range tests {
```
sgnn7, line 252: Ditto - single test using matrix test pattern

internal/plugin/connectors/http/generic/oauthprotocol/protocol_test.go
```diff
@@ -0,0 +1,394 @@
...
+					Signature:       "tnnArxj06cWHq44gCs1OSKk%2FjLY%3D",
+					SignatureMethod: "HMAC-SHA1",
+					TimeStamp:       "1318622958",
+					Token:           "370773112-GmHxMAgYyLbNEtIKZeRNFsMKPR9EyMZeS9weJAEb",
+					TokenSecret:     "LswwdoUaIvS8ltyTt5jkRh4J50vUPVVHtR2YPi5kE",
+					Version:         "1.0",
+				},
+			},
+			want: "OAuth oauth_consumer_key=\"xvz1evFS4wEEPTGEFPHBog\", oauth_nonce=\"kYjzVBB8Y0ZFabxSWbWovY3uYSQ2pTgmZeNu2VS4cg\", oauth_signature=\"tnnArxj06cWHq44gCs1OSKk%2FjLY%3D\", oauth_signature_method=\"HMAC-SHA1\", oauth_timestamp=\"1318622958\", oauth_token=\"370773112-GmHxMAgYyLbNEtIKZeRNFsMKPR9EyMZeS9weJAEb\", oauth_version=\"1.0\"",
+		},
```
sgnn7, line 338: Ditto - matrix test pattern for a single test

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
JakeQuilty: This was a big problem we ran into and saw other people have online too. When you read out the `Body` of a Request, since it's an `io.ReadCloser`, it dumps the `r.Body` contents; leaving it empty. Even if you `r.Clone()` the body, it will still refer to the same `io.ReadCloser` object and when `r.Body` is read it will dump the contents.

When `r.ParseForm()` is called, the `r.Body` contents are dumped, which causes the overall HTTP request is sent on to the destination without a body. This also causes another error, because the Header `Content-Length: X` is still present, but the actual request body length is now 0.

To fix this we dump out the contents of `r.Body`, put the contents into two separate `io.ReaderCloser` objects, assign one back to `r.Body`(since it is now empty again from being dumped), run through the method(`r.ParseForm()` dumps the body again >:0), and finally assign the second copy to `r.Body` again, so the HTTP Request can be sent on like nothing ever happened.

If you have another idea, please let us know. We worked on this for a while, and it seems like it's a fairly niche, but common issue when ran into.

References:
https://github.com/golang/go/issues/36095
https://stackoverflow.com/questions/23070876/reading-body-of-http-request-without-modifying-request-state
https://stackoverflow.com/questions/28158990/golang-io-ioutil-nopcloser

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
JakeQuilty: `r.ParseForm()` returns a `map[string][]string`. It returns this, because some query or body parameters may have multiple values. In a `map[string]string` it would have been impossible to have two keys with the same name that each have different values. 

This was the route that made the most sense to us after some thought

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
JakeQuilty: These need to be a string array to match the `map[string][]string` that we get from `r.Form`. It made a lot more sense logically to lump both of these parameter maps together, and use the same logic that follows to extract, and append the values together to make the string

internal/plugin/connectors/http/generic/config.go
JakeQuilty: `OAuth1Secrets`?

JakeQuilty: It's for the 4 secrets passed in in the config

internal/plugin/connectors/http/generic/connector.go
JakeQuilty: We need to run `renderTemplates()` to check for the `oauth1` section in config, like `queryParams` does. This looks for `oauth1` and returns an `err` if it isn't there. I agree with moving the if statement after that to only be used if `renderTemplates()` doesn't return an `err`. But how would you suggest checking for this instead?

internal/plugin/connectors/http/generic/oauthprotocol/protocol.go
JakeQuilty: Do you want the template in it's own file?

internal/plugin/connectors/http/generic/README.md
```diff
@@ -143,6 +143,26 @@ powerful transformation features.  You can use `printf` for formatting and
 you can compose functions using pipes `|`.  See the text template package docs linked
 above for detailed information on these and other features.
 
+### `oauth1`
+Like `headers` this is another key section. The `oauth1` section is used to generate
+an OAuth1 `Authorization` header. This will override any `Authorization` header that
+you have previously specified in your configuration.
```
sgnn7, line 149: We should make this last line bold since it's important for the user to see

internal/plugin/connectors/http/generic/README.md
```diff
@@ -143,6 +143,26 @@ powerful transformation features.  You can use `printf` for formatting and
...
+
+There are four required _keys_ for the `oauth1` section which are as follows:
+1. `consumer_key`
+1. `consumer_secret`
+1. `token`
+1. `token_secret`
```
sgnn7, line 155: These fields should have descriptions attached here as to what they are (or links to where you can find out more about them)

internal/plugin/connectors/http/generic/README.md
```diff
@@ -143,6 +143,26 @@ powerful transformation features.  You can use `printf` for formatting and
...
+
+The oauth1 _values_ are defined using a [Go text
+template](https://golang.org/pkg/text/template/), as defined in the
+`text/template` package.
+
+You can refer to your credentials in this template using the credential name
+preceded by a `.` (eg, `.consumer_key` and `.token` refer to the credentials
```
sgnn7, line 162:
```suggestion
preceded by a `.` (e.g. `.consumer_key` and `.token` refer to the credentials
```

internal/plugin/connectors/http/generic/README.md
```diff
@@ -143,6 +143,26 @@ powerful transformation features.  You can use `printf` for formatting and
...
+`consumer_key` and `token`). At runtime, Secretless will replace these
+credential references with your real credentials.
```
sgnn7, line 164: An example here would be good

internal/plugin/connectors/http/generic/connector.go
```diff
@@ -39,11 +40,23 @@ func (c *Connector) Connect(
...
-	params, err := renderTemplates(c.config.QueryParams, credentialsByID)
+	queryParams, err := renderTemplates(c.config.QueryParams, credentialsByID)
 	if err != nil {
 		return fmt.Errorf("failed to render query params: %s", err)
 	}
-	r.URL.RawQuery = appendQueryParams(*r.URL, params)
+	r.URL.RawQuery = appendQueryParams(*r.URL, queryParams)
+
+	// Add OAuth1 Authorization Header to the request
+	oauth1params, err := renderTemplates(c.config.OAuth1Secrets, credentialsByID)
```
sgnn7, line 50: As talked about before, this should be a conditional check and only done if oath creds are specified

sgnn7: Also `oauth1params` -> `oauth1Params` (or even `oAuth1Params`)

internal/plugin/connectors/http/generic/connector.go
```diff
@@ -39,11 +40,23 @@ func (c *Connector) Connect(
...
+	if err != nil {
+		return fmt.Errorf("failed to render OAuth1 secrets: %s", err)
+	} else {
```
sgnn7, line 53: If you return from the function, the `else` is implied already
```suggestion
	if err != nil {
		return fmt.Errorf("failed to render OAuth1 secrets: %s", err)
	}
	
	 <other code>
```

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+	"math/rand"
+	gohttp "net/http"
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type OAuth1 struct {
```
sgnn7, line 19: I see a lot of uncommented public fields here. Run the code through `go lint` and/or `go fmt`

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+const (
+	oauth_consumer_key     = "oauth_consumer_key"
+	oauth_consumer_secret  = "oauth_consumer_secret"
+	oauth_token            = "oauth_token"
+	oauth_token_secret     = "oauth_token_secret"
+	oauth_nonce            = "oauth_nonce"
+	oauth_signature        = "oauth_signature"
+	oauth_signature_method = "oauth_signature_method"
+	oauth_version          = "oauth_version"
+	oauth_timestamp        = "oauth_timestamp"
```
sgnn7, line 42: These should be sorted alphabetically

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+
+	oauth_version_number        = "1.0"
+	oauth_signature_method_type = "HMAC-SHA1"
+	nonce_length                = 16
+	nonce_charset               = "abcdefghijklmnopqrstuvwxyz" +
```
sgnn7, line 47: This should have a link to the RFC where the charset is defined

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+	oauth_token            = "oauth_token"
+	oauth_token_secret     = "oauth_token_secret"
+	oauth_nonce            = "oauth_nonce"
+	oauth_signature        = "oauth_signature"
+	oauth_signature_method = "oauth_signature_method"
+	oauth_version          = "oauth_version"
+	oauth_timestamp        = "oauth_timestamp"
+
+	oauth_version_number        = "1.0"
+	oauth_signature_method_type = "HMAC-SHA1"
```
sgnn7, line 45: I think you should put that comment in here that other modes may need to be supported in the future

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+	nonce_length                = 16
+	nonce_charset               = "abcdefghijklmnopqrstuvwxyz" +
+		"ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"0123456789"
+)
+
+var required_config_params = [...]string{
```
sgnn7, line 52: Golang variable naming uses camelCase

sgnn7: I'm not 100% sure but I think you don't need the dots in the `[]` part

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+	"consumer_key",
+	"consumer_secret",
+	"token",
+	"token_secret",
+}
+
+func generateNonce(length int, charset string) string {
+	var seededRand *rand.Rand = rand.New(
```
sgnn7, line 60: `var foo =` pattern is usually less favorable vs the `foo :=` pattern

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+		randomChars[index] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(randomChars)
+}
+
+// checkRequiredOAuthParams returns an error if a key from "required_config_params"
+// was not included in the supplied config
+func checkRequiredOAuthParams(params map[string]string) error {
+	for _, requiredParam := range required_config_params {
+		if param := params[requiredParam]; len(param) <= 0 {
```
sgnn7, line 74: This can be a single statement. Also the `<=` should be just equals or be `< 1`

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+// encodeURI percent encodes a string and changes encoding of
+// <SPACE> from "+" to "%20"
+// https://tools.ietf.org/html/rfc5849#section-3.6
+func encodeURI(stringToEncode string) string {
+	return strings.ReplaceAll(url.QueryEscape(stringToEncode), "+", "%20")
+}
+
+// collectParameters creates a map of body and query parameters by
+// percent encoding all keys and values
+// and sorting values post-encoding
```
sgnn7, line 107: Inconsistent comment word wrap length

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+	// Need to replace r.Body after r.ParseForm() dumps the body contents
+	buf, _ := ioutil.ReadAll(r.Body)
+	originalBody := ioutil.NopCloser(bytes.NewBuffer(buf))
+	copyBody := ioutil.NopCloser(bytes.NewBuffer(buf))
+	r.Body = originalBody
+
+	paramMap := make(map[string][]string, 0)
+	// get query and body params and URL
+	// encode keys and values
+	// sort values after encoding
```
sgnn7, line 120: Inconsistent word wrap length

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+	r.ParseForm()
+	for paramKey, paramValues := range r.Form {
+		encodedKey := encodeURI(paramKey)
+		paramMap[encodedKey] = paramValues
+
+		// URL encode all values in all keys
+		// Keys with multiple values require the values to be sorted post-percent encoding
```
sgnn7, line 127: Inconsistent word wrap length

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	sort.Strings(keys)
+
+	// create string
+	for keyIndex, key := range keys {
+		// Write keys and value pairs to `parameterString`
+		// ex: key0=value&key1=value&key2=value
+		for valueIndex, value := range paramMap[key] {
+			// Do not add '&' on first value
+			if keyIndex != 0 || valueIndex > 0 {
```
sgnn7, line 165: These two conditions seem the same but are written differently

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,222 @@
...
+				parameterString.WriteString("&")
+			}
+			keyValuePair := fmt.Sprintf("%s=%s", key, value)
+			parameterString.WriteString(keyValuePair)
```
sgnn7, line 169:
```suggestion
			parameterString.WriteString(key+"="+value)
```

internal/plugin/connectors/http/generic/README.md
JakeQuilty: We still need to add a panic for this scenario

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
JakeQuilty: Nonce doesn't have a specified charset or length. It just needs to be a random string of characters to make duplicate request attacks harder.
https://tools.ietf.org/html/rfc5849#section-3.3

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
mdodell: If the first key has multiple values, we want to append the ampersand for the second value.

mdodell: And we never want to add the ampersand before the first key/we want to append the ampersand for multiple keys

So this takes care of this case:
```
key[]=value1&key[]=value2
```
where `valueIndex` is 1 for `value2`

This also takes care of this case:
```
key1=value1&key2=value2
```
we only want to add the ampersand after `key1`

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
mdodell: Yup! Added that to the Changelog, Jake is writing up those issues now.

JakeQuilty:
Last 2 force pushes were fixing markdown and go lint tests

internal/plugin/connectors/http/generic/oauth/v1/protocol_test.go
```diff
@@ -0,0 +1,352 @@
...
+		ConsumerSecret:  "consumerSecret",
+		Nonce:           "",
+		SignatureMethod: "HMAC-SHA1",
+		TimeStamp:       "",
+		Token:           "token",
+		TokenSecret:     "tokenSecret",
+		Version:         "1.0",
+	}
+	if reflect.DeepEqual(result, want) {
+		t.Log("extractOAuthParams() Passed!")
```
BradleyBoutcher, line 40: Throughout this file, you use standard logging and error throwing, which can be difficult to read when running tests, and isn't the standard for testing in secretless. IIRC the logging / error style is suggested by Goland, but you can ignore that.

I'd refer to `generic/config-test.go` to see how we use the `assert` collection. It's very similar to what you have here, and easy to implement. You shouldn't have to write any custom logging formats, just provide the expected and actual values.

internal/plugin/connectors/http/generic/oauth/v1/protocol_test.go
```diff
@@ -0,0 +1,352 @@
...
+		paramString string
+		r           *gohttp.Request
+	}
+	tests := []struct {
+		name string
+		args args
+		want string
+	}{
+		{
+			name: "Can generate a valid Base string: All components supplied",
```
BradleyBoutcher, line 160: When you switch to using assert, it'll print what the name of the test is (i.e. Test_generateBaseString) which means you can strike "Can generate a valid Base string", as the context will be readable by just the method name. Which is a sign of good naming conventions! 👍

internal/plugin/connectors/http/generic/oauth/v1/protocol_test.go
```diff
@@ -0,0 +1,352 @@
...
+				Scheme:   "http",
+				Host:     "example.com",
+				Path:     "/wp-json/wp/v2/posts",
+				RawQuery: "include_entities=true",
+			},
+			Header: map[string][]string{
+				"Authorization": {"doesn't matter"},
+				"Content-Type":  {"application/x-www-form-urlencoded"},
+			},
+			Body: createRequestBody("status=Hello%20Ladies%20%2B%20Gentlemen%2C%20a%20signed%20OAuth%20request%21"),
```
BradleyBoutcher, line 77: Nit: I'd change this (ladies and gentlemen) to be something more neutral, like "foo bar hello world".

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,275 @@
...
+	var tpl bytes.Buffer
+	err = tmpl.Execute(&tpl, oauth1)
+	if err != nil {
+		return "", fmt.Errorf("could not turn the OAuth template into a string")
+	}
+	return tpl.String(), nil
+}
+
+/*
+CreateOAuth1Header calls a series of methods to construct the oAuth1 'Authorization' Header String
```
BradleyBoutcher, line 251: A few comments go past our standard line length, I'd break this (and others of a similar length) into two lines

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,275 @@
...
+		Version:         oauthVersionNumber,
+	}
+
+	return oauth
+}
+
+/*
+encodeURI percent encodes a string and changes encoding of <SPACE> from "+" to "%20"
+
+Reference: https://tools.ietf.org/html/rfc5849#section-3.6
```
BradleyBoutcher, line 118: Wow, TIL

internal/plugin/connectors/http/generic/oauth/v1/protocol_test.go
```diff
@@ -0,0 +1,352 @@
...
+	type args struct {
+		stringToEncode string
+	}
+	tests := []struct {
+		name string
+		args args
+		want string
+	}{
+		{
+			name: "Validate encodeURI(): 1",
```
BradleyBoutcher, line 317: So I have a few thoughts on this test:
1. Let's change the naming of these tests to describe the situations we're covering. The context of a test can be hard to determine from the values being passed in and out.
2. On [this page](https://tools.ietf.org/html/rfc5849#section-3.6) you shared, we can see a few cases we should test specifically.
- Characters in the unreserved character set as defined by
          [RFC3986], Section 2.3 (ALPHA, DIGIT, "-", ".", "_", "~") MUST
          NOT be encoded.
- All other characters MUST be encoded.
- The two hexadecimal characters used to represent encoded
          characters MUST be uppercase. 

Let's make sure we test these cases as well as add names that make it clear that we're testing 'em

internal/plugin/connectors/http/generic/oauth/v1/protocol_test.go
JakeQuilty: We pulled these examples straight from the Twitter examples, but that's a good point.

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
sgnn7: I'm not sure I understand - the comment was just about making the condition look the same:
```go
if keyIndex > 0 || valueIndex > 0 {
```

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,275 @@
...
+	return paramMap
+}
+
+/*
+generateParameterString creates a Parameter String from sorting an encoded map of parameters and values
+
+Reference: https://tools.ietf.org/html/rfc5849#section-3.5
+*/
+func generateParameterString(paramMap map[string][]string) string {
+	var parameterString bytes.Buffer
```
sgnn7, line 170: I wonder if maybe https://golang.org/pkg/strings/#Builder would be better here. https://medium.com/@felipedutratine/string-concatenation-in-golang-since-1-10-bytes-buffer-vs-strings-builder-2b3081848c45.

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		return "", fmt.Errorf("could not create the OAuth template")
+	}
+	var tpl bytes.Buffer
+	err = tmpl.Execute(&tpl, oauth1)
+	if err != nil {
+		return "", fmt.Errorf("could not turn the OAuth template into a string")
+	}
+	return tpl.String(), nil
+}
```
sgnn7, line 248: Lots of squished code here making it hard to read. Add newlines after each `if err != nil {}` block.

internal/plugin/connectors/http/generic/connector.go
```diff
@@ -14,6 +15,40 @@ type Connector struct {
 	config *config
 }
 
+func addOAuth1Header(c *Connector, credentialsByID connector.CredentialValuesByID, r *gohttp.Request) error {
+	oAuth1Params, _ := renderTemplates(c.config.OAuth1Secrets, credentialsByID)
```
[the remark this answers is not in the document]
JakeQuilty: I know we did this and tested the length of `oAuth1Params` for a reason, I'm just having troubling remembering exactly why(Which is exactly why we should've made a comment).I'm 99% sure it's because renderTemplates will throw an err if `oauth1` is not declared in the config. So we check the length of the returned map instead. I'm going to double check this and make a comment explaining why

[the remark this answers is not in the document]
JakeQuilty: This came from a comment by @sgnn7 [here](https://github.com/cyberark/secretless-broker/pull/1321#discussion_r465341047.

We only want to run our code if the rendered template comes back with a length > 0. If we didn't do this, we'd have to declare `oauth1` in the config every time. Do you think this is the right solution to this problem?

cc: @mdodell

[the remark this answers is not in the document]
JakeQuilty: We have error handling for someone declaring `oauth1` in the config, but not providing the necessary values [here](https://github.com/cyberark/secretless-broker/blob/deda81e5e06862d8dbbfdca06eea298c6da42b0f/internal/plugin/connectors/http/generic/oauth/v1/protocol.go#L88)

internal/plugin/connectors/http/generic/connector.go
```diff
@@ -14,6 +15,40 @@ type Connector struct {
...
+	if len(oAuth1Params) > 0 {
+		oauthHeader, err := oauth1.CreateOAuth1Header(oAuth1Params, r)
+		if err != nil {
+			return fmt.Errorf("failed to create oAuth1 'Authorization' header: %s", err)
+		} else if len(r.Header.Get("Authorization")) > 0 {
```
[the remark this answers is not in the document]
JakeQuilty: We changed how this is handled to kill Secretless if someone declares an `Authorization` header and `oauth1` in the config since the header will be overwritten anyways. Must've forgot to change the README, fixing now

internal/plugin/connectors/http/generic/README.md
```diff
@@ -143,14 +143,76 @@ powerful transformation features.  You can use `printf` for formatting and
 you can compose functions using pipes `|`.  See the text template package docs linked
 above for detailed information on these and other features.
 
+### `oauth1`
+
+Like `headers` this is another key section. The `oauth1` section is used to generate
+an OAuth1 `Authorization` header.
+
+**Note: This will override any `Authorization` header that
```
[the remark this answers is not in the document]
JakeQuilty: See https://github.com/cyberark/secretless-broker/pull/1321#discussion_r468078916

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
```diff
@@ -0,0 +1,275 @@
...
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+	"text/template"
+	"time"
+)
+
+// oAuth1 is a struct of all OAuth1 specific parameters
+type oAuth1 struct {
```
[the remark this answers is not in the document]
JakeQuilty: Switched all params to private too

CHANGELOG.md
```diff
@@ -16,6 +16,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   in the secretless configuration file, under `config`. This allows the
   construction of a query string which can have credentials injected
   as needed. [#1290](https://github.com/cyberark/secretless-broker/issues/1290)
+- Generic HTTP connector now supports `oauth1` as a configurable section in the
+  secretless configuration file, under `config`. This allows the construction of
+  a header for an OAuth 1.0 request. The oauth1 feature currently only supports
```
izgeri, line 21:
```suggestion
  a header for an OAuth 1.0 request. The OAuth1 feature currently only supports
```
or maybe this should be `oauth1` in a code block?

izgeri:
I know I'm coming to this late in the game

But why _didn't_ we just make a standalone `oauth1` connector vs adding an `oauth1` "generic" config option that overrides any other `headers` (or just `Authorization` headers?) provided to the generic HTTP connector?

I have this sense that that's effectively what we did, but we put it in the `connector/http/generic/` folder instead of in `connector/http/` directly. The `oauth1` key that can be provided in the generic HTTP connector config is not useful for any other HTTP connector besides `oauth1`.

It's possible I'm reading too much into this and this is the right call, but I would really like to question this - cc @sgnn7 @jonahx @BradleyBoutcher @doodlesbykumbi @diverdane

internal/plugin/connectors/http/generic/oauth/v1/protocol.go
JakeQuilty: lgtm! I hadn't heard of `strings.Builder` but it makes sense to use that instead

BradleyBoutcher:
@izgeri I'm copying this convo here for future reference

There's a few reasons we didn't got that route. TLDR; We explored this possibility, but the issue we ran into was how much duplicate code was involved. Essentially, we'd have to import the entire generic connector or rewrite the existing code used for parsing templates, which led us to implement it as a protocol within the generic connector. The generic connector is meant to be consumed exclusively, and doesn't allow for logic to be performed on the ejected configuration after its been instantiated.

First, there are values that are required by oauth, and they aren't distinguishable without specific keys. So, we decided to keep those keys seperate in the oauth section, which allows us to further validate and access them using existing generic connector functionality. However, that isn't the issue. We have to perform logic on the values provided by the user to generate params required by oauth1, and that logic can only take place before the generic connector is instantiated, in the same way we parse Headers and QueryParams

JakeQuilty:
Rebased with master

internal/plugin/connectors/http/generic/connector.go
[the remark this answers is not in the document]
BradleyBoutcher: My concern is if there is an issue rendering the templates, before we check what the length is, because the rendering itself may have failed. We don't need to call `addOAuth1Header` every time, but if we do, we should confirm that there were no errors when calling `renderTemplates` inside of it. 

`renderTemplates` shouldn't throw an error if `c.config.OAuth1Secrets` is empty. Further, `c.config.OAuth1Secrets` will never be `nil` because `1stringsToTemplates` sets it to an empty map if no values are given.

@sgnn7 what do you think?

Issue #1331, opened 2020-08-29
Support KV v2 backend in Vault provider


michael2m:
The current provider for secrets from HashiCorp Vault does not work for secrets stored in the KV version 2 backend. It leads to errors retrieving secrets. This is due to differences in the JSON structure of the response (wrapping in an additional "data" object). Note that KV v2 has the path to the secret include an additional "data" segment as well, e.g. "secret/data/path/to/it" for a secret "path/to/it". However this is compatible with current implementation of the secret provider in secretless-broker (ensuring the ID covers the proper path). 

Proposal to support KV v2 backend: extend the current Vault provider in secretless-broker to parse an optional version indicator in the ID of the secret being retrieved. Currently it accepts: "path/to/secret" (with implied field "value") and "path/to/secret#somefield" with explicit field "somefield". This could be extended to support an optional version and delimiter prepended to the ID, e.g. "v2:path/to/secret" or "v2:path/to/secret#somefield". When omitted it should imply "v1", to not break current current behavior and expect the KV v1 backend.

izgeri:
@michael2m thanks for filing this request!

I'm not sure what our timeline for making this change would be, but if you're interested in contributing you can find more info in our [Secretless Contributing Guide](https://github.com/cyberark/secretless-broker/blob/master/CONTRIBUTING.md) and in our general [Conjur Contributing Guide](https://github.com/cyberark/community/tree/master/Conjur/). You can also contact us on [Discourse](https://discuss.cyberarkcommons.org/c/secretless-broker/development/27) - we have a channel specifically devoted to Secretless development, and we'd be happy to help you get up and running.

izgeri:
Per @michael2m's email request, I've marked this as `implementing` for him 🎉

sgnn7:
Fixed via #1345 and queued for next release (`>= 1.7.1`). Thanks @michael2m - much appreciated!

PR #1334, opened 2020-09-04
Support Vault KV v2 backend. Connected to #1331.

michael2m:
Signed-off-by: Meijer <michaelmeijer@outlook.com>

### What does this PR do?
- Support secrets of Vault's KV v2 backend which is incompatible with the current implementation in secretless-broker.
- New behavior is implemented with backwards compatability to current (old) behavior.

Connected to #1331 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- The `vault` provider now supports loading secrets from the KV Version 2 secret
+  engine. Reference a secret in Vault using the right path and a field
+  navigation in the Secretless configuration.
+  [#1331](https://github.com/cyberark/secretless-broker/issues/1331)
+
 ## [1.7.0] - 2020-09-11
 
 ### Added
--- a/internal/providers/vault/README.md
+++ b/internal/providers/vault/README.md
@@ -1,0 +1,98 @@
+# Vault Provider
+
+The Vault provider for Secretless can fetch secrets from configured secret
+engines in [HashiCorp Vault](https://www.vaultproject.io). The provider is based
+on the [Vault API client](https://pkg.go.dev/github.com/hashicorp/vault/api) in
+Go. It reads the secret object from the configured path and returns the value
+navigated to by the configured fields (or default field otherwise).
+
+## Usage Documentation
+
+The Vault provider is configured in the `secretless.yml` using:
+
+```yaml
+from: vault
+get: /path/to/secret/in/vault
+```
+
+Or with explicit fields navigating to the value in the secret returned at path:
+
+```yaml
+from: vault
+get: /path/to/secret/in/vault#navigate.to.this.field
+```
+
+The provider will read a secret (object) at a given path and returns the value
+of field `value` (by default). By appending `#data.fieldName` to the path, the
+provider will instead read the value at the field `fieldName` in the object
+`data` in the secret (object) instead.
+
+Below are some examples showing how to configure the provider for secrets.
+
+### Example: API key from KV backends (v1 and v2)
+
+Below is an excerpt of an example configuration for a fictional "Example
+Service" that requires an API key, e.g. used in a request header. It gets the
+API key from Vault's KV version 1 backend at path `kv/example-service` under the
+secret's `value` field.
+
+```yaml
+version: 2
+services:
+  my_example_service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:8080
+    credentials:
+      apikey:
+        from: vault
+        get: kv/example-service
+        # gets path to API key in Vault, field 'value' holds the API key
+    ...
+```
+
+A slightly different configuration explicitly sets the field `api-key` (instead
+of the default `value`) to hold the API key.
+
+```yaml
+version: 2
+services:
+  my_example_service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:8080
+    credentials:
+      apikey:
+        from: vault
+        get: kv/example-service#api-key
+        # gets path to API key in Vault, field 'api-key' holds the API key
+    ...
+```
+
+If the secret is stored in a KV v2 backend (mounted at `secret` by default), the
+configuration must use the use the `data` segment in the path and the
+`#data.api-key` suffix. This is behavior specific to KV v2 in Vault, see Vault
+API docs.
+
+```yaml
+version: 2
+services:
+  my_example_service:
+    connector: generic_http
+    listenOn: tcp://0.0.0.0:8080
+    credentials:
+      apikey:
+        from: vault
+        get: secret/data/example-service#data.api-key
+        # gets path to API key in Vault stored in the KV v2 secret engine
+    ...
+```
+
+## Limitations
+
+- Only token-based login to Vault supported at the moment.
+- Only secrets that are "read" in Vault are supported at the moment. Backends
+  that require "writes" to obtain the secret (e.g. PKI, dynamic database
+  credentials) are not supported at the moment.
+- Backends that have multiple values change simultaneously (e.g. client id and
+  secret, database username and password) are not supported at the moment.
+- Limited support for KV v2 secret engine, only latest version of a secret can
+  be retrieved.
--- a/internal/providers/vault/provider.go
+++ b/internal/providers/vault/provider.go
@@ -16,7 +16,10 @@ type Provider struct {
 }
 
 // ProviderFactory constructs a Provider. The API client is configured from
-// environment variables.
+// environment variables. Underlying Vault API client by default uses:
+// - VAULT_ADDR: endpoint of Vault, e.g. http://vault:8200/
+// - VAULT_TOKEN: token to login to Vault
+// See Vault API docs at https://godoc.org/github.com/hashicorp/vault/api
 func ProviderFactory(options plugin_v1.ProviderOptions) (plugin_v1.Provider, error) {
 	config := vault.DefaultConfig()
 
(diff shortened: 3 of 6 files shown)


izgeri:
Thanks for submitting this @michael2m! I'll have the team review the code within the next couple of business days.

Given that this increases the complexity of the `vault` provider configuration, would you be able to help us in writing a little documentation for this provider? We could add a README.md file to the `/internal/providers/vault/` folder that has some info about how to configure Secretless with vault. That would include something like:
```markdown
# HashiCorp Vault Provider

## Configuration
<Required steps to configure Secretless to authenticate with HCV>

## Using the `vault` provider
<Instructions for referencing credentials from vault using this provider, including examples when using `kv` or `kv-v2`>
```

Even if we just added a godoc-style comment to the top of the `provider.go` file so that it would have some info here: https://godoc.org/github.com/cyberark/secretless-broker/internal/providers/vault, that would be good - especially since there's now more than one way to use this provider.

I would like to add a CHANGELOG entry for this, too - we provide [some guidelines](https://github.com/cyberark/community/blob/master/Conjur/CONTRIBUTING.md#changelog-guidelines) for writing good change log messages. If you'd prefer, we can add a commit later that adds a message for your changes. I might just write something like this under `Added`:
```
- The `vault` provider now supports loading secrets from the KV Secrets Engine - Version 2. To reference secrets using this version, preface the secret path with `kv-v2:` in the Secretless configuration. [cyberark/secretless-broker#1331](<link>)
```

michael2m:
You're totally right @izgeri. I will add some godoc comments. Just wondering about the docs though. It seems there was a documentation page once for the Vault provider, however it links to a non-existing page in the Conjur docs (secretless-broker/docs/docs/reference/providers/vault.md). I would prefer proper documentation, rather than a README somewhere in code.

izgeri:
@michael2m this provider is still considered [community level](https://github.com/cyberark/community/blob/master/Conjur/conventions/certification-levels.md), which we document in-repo. You can see an example [here](https://github.com/cyberark/secretless-broker/tree/master/internal/plugin/connectors/http/generic) for the generic HTTP connector.

I do have it as a TODO to make sure all community providers / connectors have this minimal documentation, and that they're linked from the project README so they're easier to find. Whether we choose to augment the godocs or add a README here, I'm fine either way - either one is linkable from the README.

Hope that makes sense, but if you have any additional questions please lmk.

internal/providers/vault/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+The Vault provider for Secretless can fetch secrets from configured backends 
+in [HashiCorp Vault](https://www.vaultproject.io). The provider is based on 
+the [Vault API client](https://pkg.go.dev/github.com/hashicorp/vault/api) in Go. 
+It reads values from the configured paths.
+
+## Usage Documentation
+
+The Vault provider is configured in the `secretless.yml`. Using
+`from: vault` and `get: /path/to/secret/in/vault` for the credentials in the
```
sgnn7, line 11: You should use a code block for the config here probably

internal/providers/vault/README.md
```diff
@@ -0,0 +1,88 @@
...
+the [Vault API client](https://pkg.go.dev/github.com/hashicorp/vault/api) in Go. 
+It reads values from the configured paths.
+
+## Usage Documentation
+
+The Vault provider is configured in the `secretless.yml`. Using
+`from: vault` and `get: /path/to/secret/in/vault` for the credentials in the 
+service connector configuration. Note that Vault supports various backends 
+that store different kinds of secrets. There is the cubbyhole, KV v1 and v2, 
+database, etc.
```
sgnn7, line 14: Here we should mention which ones we support.

internal/providers/vault/provider.go
```diff
@@ -43,6 +46,20 @@ func (p *Provider) GetName() string {
...
+// parseVaultBackend returns the (optional) backend of the backend from the id.
+func parseVaultBackend(id string) (string, string) {
+	tokens := strings.SplitN(id, ":", 2)
+	switch len(tokens) {
+	case 1:
+		return "", id
+	default:
+		return tokens[0], tokens[1]
+	}
+}
```
sgnn7, line 61: We should raise an explicit error here if the backend ID is not supported

sgnn7 (changes requested):
@michael2m Great work and thank you for your contribution! There's just a handful of things that may need a bit fixing. Also, your branch contains 2 signed commits (wanted) and an unwanted merge commit which will need a fix before we get this approved.

internal/providers/vault/provider.go
```diff
@@ -89,3 +115,41 @@ func (p *Provider) GetValue(id string) (value []byte, err error) {
...
+//   - `kv-v2:secret/data/database` returns the value of field `value` in secret object at path
+//     `secret/data/database`, unwrapping the secret data as required for KV v2 backend.
+//   - `kv-v2:secret/data/database#password` returns the value of field `password` in secret object at path
+//     `secret/data/database`, unwrapping the secret data as required for KV v2 backend.
+// Secrets in Vault are stored as objects in the shape of map[string]interface{}. In case of the KV v2 backend it is
+// wrapped in another object map[string]interface{} with a key/field `data` referencing the secret.
+// Note that in the Vault API the path to a KV v2 secret includes the `data` segment, while this is not required in
+// Vault CLI. Due to this provider using the Vault API client, the given id to GetValue must follow the conventions
+// of the Vault API path. Please see documentation of Vault for details.
+func (p *Provider) GetValue(id string) (value []byte, err error) {
```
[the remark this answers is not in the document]
michael2m: Will fix

izgeri:
@michael2m I see that you're waiting on feedback on this PR - apologies for the delay, we've been pushing hard to get a new release out for another project. I'll make sure we get back to you on this by tomorrow. (cc @sgnn7)

internal/providers/vault/provider.go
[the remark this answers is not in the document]
sgnn7: > Aren't named return values a common practice ?

Yes, a very common one but the problem is that for maintenance, they make things more difficult than the alternative. The problem is that as someone is going down the flow of the method reading the code, they will inevitably either miss the fact that some variable is a side-effect return value and/or they will need to go back up to the function definition to see its type. This kind of implicit return effect as well as back-and-forth reading has a real penalty for maintenance which we would like to avoid.

internal/providers/vault/provider.go
```diff
@@ -43,6 +46,20 @@ func (p *Provider) GetName() string {
...
 
+// VaultBackendKVv2 is the Vault backend KV version 2.
+const VaultBackendKVv2 = "kv-v2"
+
+// parseVaultBackend returns the (optional) backend of the backend from the id.
+func parseVaultBackend(id string) (string, string) {
+	tokens := strings.SplitN(id, ":", 2)
+	switch len(tokens) {
+	case 1:
+		return "", id
```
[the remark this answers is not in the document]
sgnn7: > Using kv: prefix would break existing behavior. I advice against it.

Can you elaborate on this (I'm a bit newish on the Vault API specifics)? How is a generic secret fetch different from KVv1?

> Keep in mind that 99% of the code is equal. It's just the unwrapping of the KV v2 response that is different. I don't think extensibility is improved by factoring out into kv and kv-v2 variants.

I'd disagree here to an extent. Even if we ignore the extensibility part which I for sure should have elaborated on more, there's a bit of a problem. Whenever multiple logic paths are embedded this way in a codebase, the mental overhead increases for the maintainer to know how they intertwine in the flow as they are parsing the code and the test matrix grows by orders of magnitude for each branch in the code. I think some of this pattern is generally described by [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) and more often than not it has held accurate in my experience.

internal/providers/vault/README.md
sgnn7: The Vault provider is configured in the `secretless.yml`. Using
    ```yaml
    from: vault
    get: /path/to/secret/in/vault
    ```
    for the credentials in the ...

internal/providers/vault/README.md
sgnn7: Oh I'm not suggesting adding additional support for the others, just that we mention which ones are supporting (or have confidence that they work)

internal/providers/vault/provider.go
sgnn7: This ties into my other comment about `kv:` prefix but at the very least yes we should trow an error if a backend is provided and we don't support it.

sgnn7:
@michael2m Sorry for the delay on this and thank you for your patience! As @izgeri mentioned we've been really busy on getting our redesigned [Puppet integration](https://forge.puppet.com/cyberark/conjur) project released and we have more to work on that but I've had some time today to reply to your comments. Let me know if you have any questions!

PS: I think the PR will need a rebase and the git history will need a touch of cleanup.

internal/providers/vault/provider.go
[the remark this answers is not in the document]
michael2m: Clear, will fix.

internal/providers/vault/provider.go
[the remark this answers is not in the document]
michael2m: > Can you elaborate on this (I'm a bit newish on the Vault API specifics)? How is a generic secret fetch different from KVv1?

Vault looks like a file system where you read from / write to, with contents made up of JSON. The "old" implementation the Vault provider for Conjur expects the secret value as JSON field in the returned object. Which works for various secret engines supported by Vault, hence the generic secret fetch. A notable exception is the KVv2 secret engine, due to its additional wrapping of the secret value in another object within the response object. That's why it needs to be handled differently.

> ... single-responsibility principle ...
I will look into a better code design, will get back to you on this.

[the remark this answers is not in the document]
michael2m: I opted for an alternative design, in which the distinction between various Vault secret engines/backend doesn't appear as much as it did in my previous solution. I redesigned to have more expressive field navigation. It uses `path/to/secert#navigate.to.field` instead of prefixing with `kv-v2:`. Hence the field name has become field navigation, as if traversing through an object structure. This is both compatible to the current behavior as well as extends naturally to the KV v2 case.

internal/providers/vault/provider.go
michael2m: Given new design, this should no longer be relevant.

michael2m:
Redesigned the support for KV v2 in an alternative way (see review thread above). Also cleaned up commit history rebasing the pull request off master.

internal/providers/vault/provider.go
```diff
@@ -39,53 +42,71 @@ func (p *Provider) GetName() string {
...
+//  	at path `secret/data/database`.
+// Secrets in Vault are stored as (JSON) objects in the shape of map[string]interface{}. Both path to the secret and
+// fields to the value in the secret must follow Vault API client conventions. Please see documentation of Vault for
+// details.
+func (p *Provider) GetValue(id string) ([]byte, error) {
+	path, fields := parseVaultID(id)
+
+	var err error
+	var secret *vault.Secret
+	if secret, err = p.Client.Logical().Read(path); err != nil {
```
sgnn7, line 102: You might be able to make this a bit cleaner with use of `:=`
```suggestion
	secret, err := p.Client.Logical().Read(path)
	if err != nil {
```

internal/providers/vault/provider.go
```diff
@@ -39,53 +42,71 @@ func (p *Provider) GetName() string {
...
-	if secret == nil {
-		err = fmt.Errorf("HashiCorp Vault provider could not find a secret called '%s'", id)
-		return
+// parseSecret returns value navigated by given fields on secret object.
+// Note that a secret returned from Vault is effectively a JSON object.
+func parseSecret(secret *vault.Secret, path string, fields []string) ([]byte, error) {
+	// navigate to secret value
+	var obj, ok = secret.Data, false
+	for _, field := range fields[:len(fields)-1] {
+		obj, ok = obj[field].(map[string]interface{})
```
sgnn7, line 65: I'm a bit confused why this value is declared outside of the `for` block if the check of it's value is only done immediately after this line here.

internal/providers/vault/provider.go
```diff
@@ -39,53 +42,71 @@ func (p *Provider) GetName() string {
...
+		if !ok {
+			err := fmt.Errorf("HashiCorp Vault provider expects the secret '%s' to contain field '%s' along '%s'", path, field, strings.Join(fields, "."))
```
sgnn7, line 67: This line needs word wrapping

internal/providers/vault/provider.go
```diff
@@ -39,53 +42,71 @@ func (p *Provider) GetName() string {
...
 
-	var ok bool
-	var valueObj interface{}
-	valueObj, ok = secret.Data[fieldName]
-	if !ok {
-		err = fmt.Errorf("HashiCorp Vault provider expects the secret '%s' to contain field '%s'", id, fieldName)
-		return
-	}
+	var field = fields[len(fields)-1]
+	var value = obj[field]
```
sgnn7, line 73:
```suggestion
	field := fields[len(fields)-1]
	value := obj[field]
```

I would also suggest that you name these vars a bit better. What field is this? What value is here?

internal/providers/vault/provider.go
sgnn7: I think this loop also needs a bit clearer logic about what it does either with more obvious code (preferred) or some really good comments

internal/providers/vault/provider.go
```diff
@@ -39,53 +42,71 @@ func (p *Provider) GetName() string {
...
 
-	switch v := valueObj.(type) {
+	// secret value must be either string or bytes
```
sgnn7, line 75: This (and other) comments should be capitalized

sgnn7 (changes requested):
@michael2m Hey Michael, great work on this! I like the cleaner look of this quite a bit! I left some comments on it for you but it's mostly tiny nits.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- The `vault` provider now supports loading secrets from the KV Version 2 secret
+  engine. To reference secrets using this version, preface the secret path with
+  `kv-v2:` in the Secretless configuration.
```
sgnn7, line 13: Oops - forgot that this may need updating too since the implementation changed

internal/providers/vault/provider.go
michael2m: Relates to your other remark on the same section of code.

internal/providers/vault/provider.go
michael2m: I rewrote / refactored. Effectively it takes an object and a path traversing through the object structure. Suppose obj = { "foo": { "bar": "qux" } } and fields = "foo.bar", then returned value is "qux". It navigates from outer to inner object and last field's value.

michael2m:
@sgnn7 I hope the rewrite/refactor with comments helps you understand the solution.

internal/providers/vault/provider.go
sgnn7: Oh for sure - I wasn't confused about what you were trying to do. My comment was more to ensure others can easily understand what is done in the flow if they are reading this for the first time.

sgnn7:
/approve 8d56e6354aa91bead3f4c3fc18a30be4d9ddb220

@izgeri Can you take a look too?

sgnn7:
@michael2m This looks good to me - I just need one more approval to move the branch to the repo itself so that we can verify that tests pass. I'll keep this thread updated though.

sgnn7:
Closing this branch as it is superseded by https://github.com/cyberark/secretless-broker/pull/1345

PR #1344, opened 2020-09-24
Stale: Change GetValue to GetValues for Providers to allow batch retrieval

doodlesbykumbi:
GetValues works exactly the same as GetValue, however GetValues gives providers the ability to implement arbitrary batch retrieval mechanisms.

Succeeded by https://github.com/cyberark/secretless-broker/pull/1356 (Cobertura got in the way)

### What does this PR do?

Connected to #[relevant GitHub issues, eg 76]

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -29,7 +29,6 @@ github.com/containerd/containerd v1.3.2 h1:ForxmXkA6tPIvffbrDAcPUIB32QgXkt2XFj+F
 github.com/containerd/containerd v1.3.2/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtMXFTttgp+kVtyUA=
 github.com/cyberark/conjur-api-go v0.5.2 h1:8ntk07YNRz5bBwjNXkDEAPR70Yr+J2MN8NGlkhaMC3k=
 github.com/cyberark/conjur-api-go v0.5.2/go.mod h1:hwaReWirzgKor+JtH6vbwZaASDXulvd0SzGCloC5uOc=
-github.com/cyberark/conjur-authn-k8s-client v0.13.0/go.mod h1:JTeGIeRO59J7mMEc5yF6FPtk1QnaAzs4GyZa4WldqZc=
 github.com/cyberark/conjur-authn-k8s-client v0.19.0 h1:zHjdKyZ8bu4cRyV3iYMh1/XfIVtTugoiU/CflnboP9Q=
 github.com/cyberark/conjur-authn-k8s-client v0.19.0/go.mod h1:qacUJXCppU1Rg/C+br9B1jBitTq4yG04oc4a+cfI200=
 github.com/cyberark/secretless-broker v1.4.1-0.20191211191712-251c5ec034af/go.mod h1:+GueI3WCJL5gDYaYa38ZokAR8ceEyCVet7MkuZyjf80=
@@ -246,7 +245,6 @@ github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
-go.mongodb.org/mongo-driver v1.4.2 h1:WlnEglfTg/PfPq4WXs2Vkl/5ICC6hoG8+r+LraPmGk4=
 golang.org/x/crypto v0.0.0-20180621125126-a49355c7e3f8/go.mod h1:6SG95UA2DQfeDnfUPMdvaQW0Q7yPrPDi9nlGo2tz2b4=
 golang.org/x/crypto v0.0.0-20180904163835-0709b304e793/go.mod h1:6SG95UA2DQfeDnfUPMdvaQW0Q7yPrPDi9nlGo2tz2b4=
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
--- a/go.sum
+++ b/go.sum
@@ -29,6 +29,8 @@ github.com/containerd/containerd v1.3.2 h1:ForxmXkA6tPIvffbrDAcPUIB32QgXkt2XFj+F
 github.com/containerd/containerd v1.3.2/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtMXFTttgp+kVtyUA=
 github.com/cyberark/conjur-api-go v0.5.2 h1:8ntk07YNRz5bBwjNXkDEAPR70Yr+J2MN8NGlkhaMC3k=
 github.com/cyberark/conjur-api-go v0.5.2/go.mod h1:hwaReWirzgKor+JtH6vbwZaASDXulvd0SzGCloC5uOc=
+github.com/cyberark/conjur-authn-k8s-client v0.13.0/go.mod h1:JTeGIeRO59J7mMEc5yF6FPtk1QnaAzs4GyZa4WldqZc=
+github.com/cyberark/conjur-authn-k8s-client v0.19.0 h1:zHjdKyZ8bu4cRyV3iYMh1/XfIVtTugoiU/CflnboP9Q=
 github.com/cyberark/conjur-authn-k8s-client v0.19.0/go.mod h1:qacUJXCppU1Rg/C+br9B1jBitTq4yG04oc4a+cfI200=
 github.com/cyberark/secretless-broker v1.4.1-0.20191211191712-251c5ec034af/go.mod h1:+GueI3WCJL5gDYaYa38ZokAR8ceEyCVet7MkuZyjf80=
 github.com/cyberark/summon v0.7.0 h1:eBjyNzpJEeFb7BFcOgidM6S/UbEsN7LGSFeQE7szPsk=
@@ -244,6 +246,7 @@ github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+go.mongodb.org/mongo-driver v1.4.2 h1:WlnEglfTg/PfPq4WXs2Vkl/5ICC6hoG8+r+LraPmGk4=
 golang.org/x/crypto v0.0.0-20180621125126-a49355c7e3f8/go.mod h1:6SG95UA2DQfeDnfUPMdvaQW0Q7yPrPDi9nlGo2tz2b4=
 golang.org/x/crypto v0.0.0-20180904163835-0709b304e793/go.mod h1:6SG95UA2DQfeDnfUPMdvaQW0Q7yPrPDi9nlGo2tz2b4=
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
--- a/internal/plugin/resolver.go
+++ b/internal/plugin/resolver.go
@@ -98,31 +98,52 @@ func (resolver *Resolver) Resolve(credentials []*config_v2.Credential) (map[stri
 			resolver.LogFatalf("ERROR: Provider '%s' could not be used! %v", providerID, err)
 		}
 
-		// Create secretIds slice
-		var secretIds = make([]string, len(credentialsForProvider))
+		// Create secretIds slice and credentialBySecretId map
+		secretIds := make([]string, len(credentialsForProvider))
 		for idx, cred := range credentialsForProvider {
 			secretIds[idx] = cred.Get
 		}
 
 		// Resolves all credentials for current provider
-		secretValues, err := provider.GetValues(secretIds...)
+		providerResponses, err := provider.GetValues(secretIds...)
 		if err != nil {
-			errInfo := fmt.Sprintf("ERROR: Resolving credentials from provider '%s' failed: %v",
+			errInfo := fmt.Sprintf(
+				"ERROR: Resolving credentials from provider '%s' failed: %v",
 				provider.GetName(),
-				err)
+				err,
+			)
 			log.Println(errInfo)
 
 			errorStrings = append(errorStrings, errInfo)
 			continue
 		}
 
-		for idx, secretValue := range secretValues {
-			credential := credentialsForProvider[idx]
+		// Collect errors from provider responses
+		var hasErrors bool
+		for _, providerResponse := range providerResponses {
+			if providerResponse.Error != nil {
+				hasErrors = true
+				errorStrings = append(errorStrings, providerResponse.Error.Error())
+				continue
+			}
+		}
+		if hasErrors {
+			continue
+		}
 
-			result[credential.Name] = secretValue
+		// Set provider responses on result map before returning
+		for _, credential := range credentialsForProvider {
+			credentialName := credential.Name
+			secretValue := providerResponses[credential.Get].Value
+
+			result[credentialName] = secretValue
 
 			if resolver.EventNotifier != nil {
-				resolver.EventNotifier.ResolveCredential(provider, credential.Name, secretValue)
+				resolver.EventNotifier.ResolveCredential(
+					provider,
+					credentialName,
+					secretValue,
+				)
 			}
 		}
 	}
--- a/internal/plugin/resolver.go
+++ b/internal/plugin/resolver.go
@@ -81,19 +81,34 @@ func (resolver *Resolver) Resolve(credentials []*config_v2.Credential) (map[stri
 	errorStrings := make([]string, 0, len(credentials))
 
 	var err error
-	for _, credential := range credentials {
-		var provider plugin_v1.Provider
-		var value []byte
 
-		if provider, err = resolver.Provider(credential.From); err != nil {
-			resolver.LogFatalf("ERROR: Provider '%s' could not be used! %v", credential.From, err)
+	// Group credentials by provider
+	var credentialsByProvider = make(map[string][]*config_v2.Credential)
+	for _, credential := range credentials {
+		credentialsByProvider[credential.From] = append(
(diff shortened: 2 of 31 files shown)


k8s-ci/k8s_crds/test
```diff
@@ -16,6 +16,11 @@ exit_err() {
    cat ${log};
    rm -rf ${log};
 
+
+  # Snapshot of workloads
+  kubectl --namespace "$(current_namespace)" get pods || true
+  kubectl --namespace "$(current_namespace)" get events --field-selector type=Warning || true
```
sgnn7, line 22: Add debug echos to this - it's hard to tell what you might be looking at just from output of a kubectl command

k8s-ci/k8s_crds/test
```diff
@@ -161,4 +166,4 @@ function main() {
   run_test_case "update configuration object" ohnoeswechangedit supersecretpassword
 }
 
-main
+main || exit_err "Failed inside main"
```
sgnn7, line 169: Maybe we need something a bit more descriptive like the suggestion but feel free to change it.
```suggestion
main || exit_err "K8s-CRD test: Failed inside main()"
```

internal/summon/command/run.go
```diff
@@ -37,20 +38,48 @@ func buildEnvironment(secrets map[string]string, secretsMap secretsyml.SecretsMa
...
+			varSecretsSpecPaths = append(varSecretsSpecPaths, spec.Path)
 		} else {
 			// If the spec isn't a variable, use its value as-is
-			value = spec.Path
+			value := spec.Path
+			result[key] = value
+		}
+	}
+
+
```
sgnn7, line 56: Only needs 1 newline here

internal/summon/command/run.go
```diff
@@ -37,20 +38,48 @@ func buildEnvironment(secrets map[string]string, secretsMap secretsyml.SecretsMa
...
+	// Get the variable values
+	providerResponses, err := provider.GetValues(varSecretsSpecPaths...)
```
sgnn7, line 58: This should probably only be done if `if spec.IsVar()` is evaluated to true. You may want to consider an early exit instead when `if spec.IsVar()` is false earlier in the func.

internal/summon/command/run.go
```diff
@@ -37,20 +38,48 @@ func buildEnvironment(secrets map[string]string, secretsMap secretsyml.SecretsMa
...
+	if err != nil {
+		return
```
sgnn7, line 60: Not a blocker but the use of function def line variables makes it really difficult to tell that you're not returning a success here

internal/plugin/v1/provider.go
```diff
@@ -7,11 +7,49 @@ type ProviderOptions struct {
...
+// This is a convenience function since most providers with batch retrieval capabilities
+// will have need the exact same code. Note: most internal providers simply use this
+// function in their implementation of the Provider interface's GetValues method.
+func GetValues(
+	p singleValueProvider,
+	ids ...string,
+) (map[string]ProviderResponse, error) {
+	var responses = map[string]ProviderResponse{}
+
+	var setOfIds = map[string]struct{}{}
```
sgnn7, line 41: We don't generally use `foo := bar` vs `var foo = "bar"` in the codebase

sgnn7, line 41: This use of `setOfIds` seems overcomplicated - you can easily add results to `responses` as you get them and only fetch if the result is not in the map.

internal/plugin/v1/provider.go
```diff
@@ -7,11 +7,49 @@ type ProviderOptions struct {
...
+// GetValues takes in variable ids and returns their resolved values by making sequential
+// calls to a singleValueProvider.
+// This is a convenience function since most providers with batch retrieval capabilities
+// will have need the exact same code. Note: most internal providers simply use this
+// function in their implementation of the Provider interface's GetValues method.
+func GetValues(
+	p singleValueProvider,
+	ids ...string,
+) (map[string]ProviderResponse, error) {
+	var responses = map[string]ProviderResponse{}
```
sgnn7, line 39: Why not use a `func (p *Provider) GetValues` vs the global func?

internal/plugin/resolver.go
```diff
@@ -81,30 +81,70 @@ func (resolver *Resolver) Resolve(credentials []*config_v2.Credential) (map[stri
...
 		}
 
-		// This provider cannot resolve the named credential
-		if value, err = provider.GetValue(credential.Get); err != nil {
-			errInfo := fmt.Sprintf("ERROR: Resolving credential '%s' from provider '%s' failed: %v",
-				credential.Get,
-				credential.From,
-				err)
+		// Create secretIds slice and credentialBySecretId map
+		var secretIds = make([]string, len(credentialsForProvider))
```
sgnn7, line 102: nit: Mixing/matching `var foo = bar` vs `foo := bar`

internal/providers/vault/provider.go
```diff
@@ -54,6 +54,12 @@ func parseVaultID(id string) (string, string) {
 	}
 }
 
+// GetValues takes in variable ids and returns their resolved values. This method is
+// needed to the Provider interface
+func (p *Provider) GetValues(ids ...string) (map[string]plugin_v1.ProviderResponse, error) {
+	return plugin_v1.GetValues(p, ids...)
+}
```
sgnn7, line 61: Missing docs change in https://github.com/cyberark/secretless-broker/blob/master/internal/providers/vault/README.md

internal/plugin/v1/provider.go
```diff
@@ -7,11 +7,49 @@ type ProviderOptions struct {
...
+	for id := range setOfIds {
+		var pr ProviderResponse
+
+		pr.Value, pr.Error = p.GetValue(id)
+
+		responses[id] = pr
+	}
+
+	return responses, nil
+}
```
sgnn7, line 55: I couldn't find any tests for this func

internal/plugin/resolver.go
```diff
@@ -81,30 +81,70 @@ func (resolver *Resolver) Resolve(credentials []*config_v2.Credential) (map[stri
...
+		// Collect errors from provider responses
+		var hasErrors bool
+		for _, providerResponse := range providerResponses {
+			if providerResponse.Error != nil {
+				hasErrors = true
+				errorStrings = append(errorStrings, providerResponse.Error.Error())
+				continue
+			}
+		}
+		if hasErrors {
```
sgnn7, line 130: This brings an interesting question: do we care about finishing resolution of all providers if we hit an error in any provider? My guess is that a single error in fetching variables is enough to throw an error but I'd like to hear your thoughts on this.

internal/plugin/resolver.go
```diff
@@ -81,30 +81,70 @@ func (resolver *Resolver) Resolve(credentials []*config_v2.Credential) (map[stri
...
+		for _, credential := range credentialsForProvider {
+			credentialName := credential.Name
+			secretValue := providerResponses[credential.Get].Value
+
+			result[credentialName] = secretValue
 
-		if resolver.EventNotifier != nil {
-			resolver.EventNotifier.ResolveCredential(provider, credential.Name, value)
+			if resolver.EventNotifier != nil {
+				resolver.EventNotifier.ResolveCredential(
```
sgnn7, line 142: This seems like a very late time to notify that the credentials changed. I think we need to do this before sending the credential to a provider - WDYT?

internal/plugin/resolver.go
```diff
@@ -81,30 +81,70 @@ func (resolver *Resolver) Resolve(credentials []*config_v2.Credential) (map[stri
...
+					provider,
+					credentialName,
+					secretValue,
+				)
+			}
```
sgnn7, line 147: I think we need a few more tests for this since we introduced quite a few more logic bifurcation paths

sgnn7 (changes requested):
@doodlesbykumbi Great progress forward with some mostly-small comments. I would like to see multi-value fetch tests for each provider though in the e2e tests too.

internal/plugin/resolver.go
doodlesbykumbi: You need the credential value to notify so it needs to be after getting the credential from the provider.

internal/plugin/resolver.go
doodlesbykumbi: I think the developer UX of getting all the useful information about errors would be better here. There's no added cost because secret fetching is delegated to the provider and it is at the provider's discretion to fail fast or not. If we have the additional error info I think we ought to show it.

internal/providers/vault/provider.go
doodlesbykumbi: This is a new file. I needed to rebase to get this but the changes of this PR have no noticeable impact on the external UX that would warrant a change to this file. Unless if I missed something

internal/plugin/v1/provider.go
doodlesbykumbi: Can you elaborate on this one please ? I'm lost.

internal/plugin/v1/provider.go
doodlesbykumbi: Thank you, much better approach!

internal/summon/command/run.go
doodlesbykumbi: Addressed

internal/summon/command/run.go
doodlesbykumbi: I don't follow. `varSecretsSpecPaths` is only populated from specs that are truthy for `spec.IsVar()`.

internal/plugin/v1/provider.go
doodlesbykumbi: Definitely added tests now

internal/plugin/resolver.go
sgnn7: You are only notifying the management module and we don't want to pass the cred value so I think it can be done earlier

internal/plugin/v1/provider.go
```diff
@@ -7,11 +7,47 @@ type ProviderOptions struct {
...
+) (map[string]ProviderResponse, error) {
+	responses := map[string]ProviderResponse{}
+
+	for _, id := range ids {
+		if _, ok := responses[id]; ok {
+			continue
+		}
+
+		var pr ProviderResponse
+		pr.Value, pr.Error = p.GetValue(id)
```
sgnn7, line 47: Not a blocker but this would be nicer to see in the more-common `foo := ProviderResponse{}` pattern.

internal/summon/command/run.go
sgnn7: You're right - I totally didn't explain well lol. What I meant is that if `varSecretsSpec*` are empty, there's no need to send any requests to the provider.

k8s-ci/k8s_crds/test
```diff
@@ -16,7 +16,16 @@ exit_err() {
...
+
+  # Snapshot of workloads
+  echo "- Snapshot of workloads on exit"
+  echo "-- Current namespace: $(current_namespace)"
+  echo "-- Pods in namespace:"
+  kubectl --namespace "$(current_namespace)" get pods || true
+  echo "-- Warning events in namespace:"
+  kubectl --namespace "$(current_namespace)" get events --field-selector type=Warning || true
+
+  exit 1
```
sgnn7, line 28: nit: I could be wrong but the added line are not indented at the same level as earlier lines in that func. Either the func indent or the added line indent is wrong.

internal/plugin/v1/testutils/testutils.go
```diff
@@ -0,0 +1,50 @@
...
+	return func() {
+		values, err := provider.GetValues(id)
+
+		convey.So(err, convey.ShouldBeNil)
+		convey.So(values, convey.ShouldContainKey, id)
+		convey.So(values[id].Value, convey.ShouldBeNil)
+		convey.So(values[id].Error, convey.ShouldNotBeNil)
+		convey.So(values[id].Error.Error(), convey.ShouldEqual, expectedErrString)
+	}
+}
```
sgnn7, line 50: I still haven't seen new tests in the providers that fetch more than one secret. It would be good to be sure that we not only can fetch one value but that we can do the same for >1 when using `GetValues`.

sgnn7 (changes requested):
@doodlesbykumbi Just a few more comments but it's almost there!

internal/plugin/resolver.go
doodlesbykumbi: @sgnn7 You do pass the value `resolver.EventNotifier.ResolveCredential(provider, credential.Name, value)`

internal/plugin/v1/testutils/testutils.go
doodlesbykumbi: Initially I thought testing the `plugin_v1.GetValues` would suffice but your comment has highlighted that we can't take for granted that each provider is using `plugin_v1.GetValues`. Thanks. I've added the test cases for multiple provides :)

k8s-ci/k8s_crds/test
doodlesbykumbi: You were right. Fixed.

go.sum
```diff
@@ -29,6 +29,8 @@ github.com/containerd/containerd v1.3.2 h1:ForxmXkA6tPIvffbrDAcPUIB32QgXkt2XFj+F
 github.com/containerd/containerd v1.3.2/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtMXFTttgp+kVtyUA=
 github.com/cyberark/conjur-api-go v0.5.2 h1:8ntk07YNRz5bBwjNXkDEAPR70Yr+J2MN8NGlkhaMC3k=
 github.com/cyberark/conjur-api-go v0.5.2/go.mod h1:hwaReWirzgKor+JtH6vbwZaASDXulvd0SzGCloC5uOc=
+github.com/cyberark/conjur-authn-k8s-client v0.13.0/go.mod h1:JTeGIeRO59J7mMEc5yF6FPtk1QnaAzs4GyZa4WldqZc=
```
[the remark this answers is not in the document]
doodlesbykumbi: Resolved

go.sum
```diff
@@ -244,6 +246,7 @@ github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+go.mongodb.org/mongo-driver v1.4.2 h1:WlnEglfTg/PfPq4WXs2Vkl/5ICC6hoG8+r+LraPmGk4=
```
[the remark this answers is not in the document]
doodlesbykumbi: Definitely. Taken out

internal/plugin/resolver.go
sgnn7: Hmm... this is an odd one - I think the API definition is not quite right maybe. There is no reason that I can think of for why a management module would need to know about the value of the secret and we're not using it as a callback for the connector to get the value. For this PR, we can ignore it but it's something that probably should be figured out.

internal/summon/command/temp_factory.go
```diff
@@ -63,24 +63,33 @@ func defaultTempPath(sharedMemoryDir string) string {
...
+	}()
+
+	if err != nil {
+		return "", err
+	}
+
+	_, err = f.Write([]byte(value))
+	if err != nil {
+		return "", err
+	}
```
sgnn7, line 79: Couple of issues here:
- `f` must be a value before we set a `defer`
- `defer` can be inlined

```suggestion
func (tf *TempFactory) Push(value string) (string, error) {
    f, err := ioutil.TempFile(tf.path, ".summon")
    if err != nil {
        return "", err
    }
	
    defer f.Close()

    _, err = f.Write([]byte(value))
    if err != nil {
        return "", err
    }
```

internal/summon/command/run.go
```diff
@@ -25,33 +27,83 @@ type Subcommand struct {
...
 			// If the spec isn't a variable, use its value as-is
-			value = spec.Path
+			value := spec.Path
+			result[key] = value
 		}
-		result[key] = value
 	}
-	return
+
+	if atLeastOneVar := len(varSecretsSpecPaths) > 0; !atLeastOneVar {
```
sgnn7, line 73: Not a blocker but this could have been simpler with just the condition:
```go
# If there are no variables to resolve, return what we have
if len(varSecretsSpecPaths) == 0 {
```

internal/summon/command/temp_factory.go
```diff
@@ -63,24 +63,33 @@ func defaultTempPath(sharedMemoryDir string) string {
...
 	for _, file := range tf.files {
-		os.Remove(file)
+		_ = os.Remove(file)
 	}
 	// Also remove the tempdir if it's not the shared memory directory
 	if !strings.Contains(tf.path, defaultSharedMemoryDir) {
-		os.Remove(tf.path)
+		_ = os.Remove(tf.path)
 	}
-	tf = nil
```
sgnn7, line 85: Two things here:
- If we're not clearing the object (which I agree is weird), we need to at least reset `tf.files[]` to an empty array so that repeated invocation won't try to delete already-deleted files. 
- If we're not clearing the object here too, it is a bit problematic that we remove the directory in cleanup since it allows for calling `Push()` again but the directory won't be there.

internal/summon/command/temp_factory_test.go
```diff
@@ -39,9 +36,62 @@ func (e *envSnapshot) restoreEnv() {
...
+
+	Convey("Push reports errors", t, func() {
+		tempFactory := NewTempFactory("dir-not-found")
+		defer tempFactory.Cleanup()
+
+		_, err := tempFactory.Push("moo")
+		So(err, ShouldNotBeNil)
+		So(err.Error(), ShouldContainSubstring, "no such file or directory")
+	})
+
```
sgnn7, line 88: nit: Stray newline

sgnn7 (changes requested):
@doodlesbykumbi Great work! There's only a few more things in the new code added that need a bit of attention.

BradleyBoutcher (commented):
Nit: Don't forget to update the changelog and any applicable README's!

internal/summon/command/temp_factory.go
doodlesbykumbi: There is no such thing as clearing an object in Go as far as I know. But `tf.files = nil` is a worthwhile thing.

internal/plugin/resolver.go
```diff
@@ -81,37 +135,61 @@ func (resolver *Resolver) Resolve(credentials []*config_v2.Credential) (map[stri
...
 		}
 	}
 
 	err = nil
 	if len(errorStrings) > 0 {
-		err = fmt.Errorf(strings.Join(errorStrings, "\n"))
+		// Sort the error strings to provide deterministic output
+		sort.Strings(errorStrings)
+
+		err = fmt.Errorf(strings.Join(errorStrings, "; "))
```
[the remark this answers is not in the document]
doodlesbykumbi: Here's an example,`ERROR: Resolving credentials from provider 'env' failed: env cannot find environment variable 'something-also-not-in-env', env cannot find environment variable 'something-not-in-env'; ERROR: Resolving credentials from provider 'file' failed: open something-not-on-file: no such file or directory`. Hmm maybe newlines are a better option

internal/summon/command/temp_factory.go
sgnn7: > `tf.files = nil` is a worthwhile thing.

This may need to be a new empty array though instead of nil

doodlesbykumbi:
@BradleyBoutcher re:
> Nit: Don't forget to update the changelog and any applicable README's!. 

This is entirely an internal change so I think neither of these are needed

doodlesbykumbi:
Succeeded by #1356 (Cobertura got in the way)

PR #1345, opened 2020-09-25
Contrib: Support Vault backend KVv2. Connected to #1331.

sgnn7:
## Branch transfer from https://github.com/cyberark/secretless-broker/pull/1334

Signed-off-by: Meijer <michaelmeijer@outlook.com>

### What does this PR do?
- Support secrets of Vault's KV v2 backend which is incompatible with the current implementation in secretless-broker.
- New behavior is implemented with backwards compatability to current (old) behavior.

Connected to #1331 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/internal/providers/vault/provider.go
+++ b/internal/providers/vault/provider.go
@@ -16,7 +16,10 @@ type Provider struct {
 }
 
 // ProviderFactory constructs a Provider. The API client is configured from
-// environment variables.
+// environment variables. Underlying Vault API client by default uses:
+// - VAULT_ADDR: endpoint of Vault, e.g. http://vault:8200/
+// - VAULT_TOKEN: token to login to Vault
+// See Vault API docs at https://godoc.org/github.com/hashicorp/vault/api
 func ProviderFactory(options plugin_v1.ProviderOptions) (plugin_v1.Provider, error) {
 	config := vault.DefaultConfig()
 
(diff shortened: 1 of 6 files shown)


internal/providers/vault/provider.go
```diff
@@ -39,53 +42,87 @@ func (p *Provider) GetName() string {
...
+		err := fmt.Errorf("HashiCorp Vault provider expects secret in '%s' at '%s'", fields, path)
+		return nil, err
 	}
 
-	switch v := valueObj.(type) {
+	// Secret value must be either string or bytes
+	switch v := value.(type) {
 	case string:
-		value = []byte(v)
+		return []byte(v), nil
```
izgeri, line 97: @sgnn7 is this supposed to return `[]byte(value)`? `v` is `value.(type)`

same Q for below too

sgnn7: This didn't change from original code afaict it's just that the new code renamed `valueObj` -> `value`. In the diff, this makes it seem like we're returning a different thing but it's the same. You can see the difference easier by comparing these two snippets:
- [Old code](https://github.com/cyberark/secretless-broker/blob/d084b5c862cbac3ce2a542d400c08b6f6e8e1552/internal/providers/vault/provider.go#L76-L89)
- [New code](https://github.com/cyberark/secretless-broker/blob/9c9197fb4d976cbb562d502f97386bc305d386e7/internal/providers/vault/provider.go#L88-L103)

izgeri (approved):
LGTM. Thanks for contributing this @michael2m!!

PR #1354, opened 2020-10-16
1046 - Tests don't exit after first failure

### LINKED ISSUE #1046: Integration tests for HTTP generic and baseauth no longer exit on first failure
## Description
The integration tests for the secretless HTTP generic and HTTP basicauth connector plugins
currently exit upon the first test case failure that occurs.

What should happen is that testing continues after each failure, so that a complete picture
can be presented of what is (or is not) failing to facilitate debugging/triage.

The fix will be to modify the assert_contains() function in these tests to not exit upon failure.

AC:
- [x] Integration test for secretless HTTP basicauth connector plugin continues when a test case fails
- [ ] Changes merged into cyberark/secretless

BradleyBoutcher:
### What does this PR do?
Our generic and basicauth tests currently exit after the
first failure. Adjustments have been made to prevent
this, so tests display all failures before exiting.

Resolves #1046 

#### Change log
- [X] This PR does not include user-facing changes and doesn't require a CHANGELOG update

#### Test coverage
- [X] The changes in this PR do not require tests

#### Documentation
- [X] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -29,9 +29,7 @@ github.com/containerd/containerd v1.3.2 h1:ForxmXkA6tPIvffbrDAcPUIB32QgXkt2XFj+F
 github.com/containerd/containerd v1.3.2/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtMXFTttgp+kVtyUA=
 github.com/cyberark/conjur-api-go v0.5.2 h1:8ntk07YNRz5bBwjNXkDEAPR70Yr+J2MN8NGlkhaMC3k=
 github.com/cyberark/conjur-api-go v0.5.2/go.mod h1:hwaReWirzgKor+JtH6vbwZaASDXulvd0SzGCloC5uOc=
-github.com/cyberark/conjur-authn-k8s-client v0.13.0/go.mod h1:JTeGIeRO59J7mMEc5yF6FPtk1QnaAzs4GyZa4WldqZc=
-github.com/cyberark/conjur-authn-k8s-client v0.16.1 h1:tiYIUYuBr578BCYlx+mbPiE4dWi+NsCu6Dy9cSRv4/M=
-github.com/cyberark/conjur-authn-k8s-client v0.16.1/go.mod h1:qacUJXCppU1Rg/C+br9B1jBitTq4yG04oc4a+cfI200=
+github.com/cyberark/conjur-authn-k8s-client v0.19.0/go.mod h1:qacUJXCppU1Rg/C+br9B1jBitTq4yG04oc4a+cfI200=
 github.com/cyberark/secretless-broker v1.4.1-0.20191211191712-251c5ec034af/go.mod h1:+GueI3WCJL5gDYaYa38ZokAR8ceEyCVet7MkuZyjf80=
 github.com/cyberark/summon v0.7.0 h1:eBjyNzpJEeFb7BFcOgidM6S/UbEsN7LGSFeQE7szPsk=
 github.com/cyberark/summon v0.7.0/go.mod h1:S7grcxHeUxfL1vRTQUyq9jGK8yG6V/tSlLPQ6tHRO4k=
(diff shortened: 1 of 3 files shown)


go.sum
```diff
@@ -29,6 +29,7 @@ github.com/containerd/containerd v1.3.2 h1:ForxmXkA6tPIvffbrDAcPUIB32QgXkt2XFj+F
 github.com/containerd/containerd v1.3.2/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtMXFTttgp+kVtyUA=
 github.com/cyberark/conjur-api-go v0.5.2 h1:8ntk07YNRz5bBwjNXkDEAPR70Yr+J2MN8NGlkhaMC3k=
 github.com/cyberark/conjur-api-go v0.5.2/go.mod h1:hwaReWirzgKor+JtH6vbwZaASDXulvd0SzGCloC5uOc=
+github.com/cyberark/conjur-authn-k8s-client v0.13.0/go.mod h1:JTeGIeRO59J7mMEc5yF6FPtk1QnaAzs4GyZa4WldqZc=
```
sgnn7, line 32: This addition seems like an error

test/connector/http/basicauth/test
```diff
@@ -16,6 +16,8 @@ if $local_test; then
   docker_volume_args=(--volume "$(cd ../../../..; pwd):/secretless")
 fi
 
+isError=false
```
sgnn7: `isError` -> `testsHaveFailures` maybe

test/connector/http/basicauth/test
```diff
@@ -93,4 +95,8 @@ echo "Test: Secretless configured with incorrect password fails"
 bad_resp=$(ping_target_service_thru_secretless 8081)
 assert_contains "$bad_resp" "401 Unauthorized"
 
-echo "PASS"
+if [[ isError ]]; then
+  exit 1
+else
+  echo "PASS"
+fi
```
sgnn7: No need to do the `else` since we can early-exit.
```suggestion
if [[ isError ]]; then
  echo "FAIL: basicauth/test"
  exit 1
fi

echo "PASS: basicauth/test"
```

test/connector/http/generic/http_test.go
```diff
@@ -31,21 +32,16 @@ func TestCreds(t *testing.T) {
 	for _, testCase := range testCases {
 		t.Run(testCase.description, func(t *testing.T) {
 			srv, err := httpServer(testCase.serverUsername, testCase.serverPassword)
-			if !assert.NoError(t, err) {
-				return
-			}
+			assert.NoError(t, err)
```
sgnn7, line 35: The removal of the `if` is not correct. If the assertion fails, it will not immediately return but will continue running and likely cause a segfault. Same issue may be present in the other removals.

test/connector/http/basicauth/test
```diff
@@ -16,6 +16,8 @@ if $local_test; then
   docker_volume_args=(--volume "$(cd ../../../..; pwd):/secretless")
 fi
 
+export isError=false
```
sgnn7, line 19: There is no need for this to be exportable. Also, use string values (`"true"` and `"false"`) as it can make the comparison _very_ ambiguous depending on the conditional checks.

test/connector/http/basicauth/test
```diff
@@ -93,4 +95,8 @@ echo "Test: Secretless configured with incorrect password fails"
 bad_resp=$(ping_target_service_thru_secretless 8081)
 assert_contains "$bad_resp" "401 Unauthorized"
 
-echo "PASS"
+if [ "$isError" = true ]; then
+  exit 1
+else
+  echo "PASS"
+fi
```
sgnn7, line 102:
```suggestion
if [ "$isError" = true ]; then
  echo "FAIL: http/basicauth/test"
  exit 1
fi

echo "PASS: http/basicauth/test"
```

sgnn7 (changes requested):
@BradleyBoutcher A few comments

PR #1370, opened 2021-01-26
Fix aws connector automatic endpoint discovery

### LINKED ISSUE #1369: host header not set correctly in automatic endpoint discovery for aws connector
## Steps to Reproduce
1. Setup a Secretless service with an AWS connector at 8080, with automatic endpoint discovery (`--endpoint-url http://secretless.empty` in the alias below) .
```
alias aws="AWS_ACCESS_KEY_ID=x AWS_SECRET_ACCESS_KEY=x HTTP_PROXY=http://localhost:8080 aws --endpoint-url http://secretless.empty"
```
2. Run `aws s3 ls` to list all buckets. You'll likely get
```
An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: The specified bucket does not exist
```

**NOTE**: This bug doesn't show up for `aws sts get-caller-identity`, maybe this endpoint doesn't care about the Host header.

This bug does not show up when you manually set the endpoint. So `aws --endpoint-url http://s3.us-east-1.amazonaws.com s3 ls` works perfectly fine.

## Expected Results

It should just work. The AWS-CLI should return the list of buckets on the AWS account in question.

## Actual Results (including error logs, if applicable)

1. We get an error from AWS-CLI. 
    ```
    An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: The specified bucket does not exist
    ```
2. Upon inspection of debug logs from Secretless we see that the host header for the request to the AWS endpoint is not the right value
    ```
    2021/01/26 23:29:00 [DEBUG] HTTP Proxy on tcp://0.0.0.0:8080: Making request: 
    GET /?delimiter=%2F&encoding-type=url&list-type=2&prefix= HTTP/1.1
    Host: secretless.empty
    Accept-Encoding: identity
    ```

## Additional Information

N/A

doodlesbykumbi:
### What does this PR do?
_What's changed? Why were these changes made?_

There were 2 bugs.
1. We modified only req.URL{Scheme, Host} with the new endpoint, we missed req.Host.
2. We were modifying the endpoint after signing the request.

2 did not show up until 1 was resolved. This is likely because the request signing doesn't
take req.URL{Scheme, Host} into account.

_How should the reviewer approach this PR, especially if manual tests are required?_

N/A

_Are there relevant screenshots you can add to the PR description?_

N/A.

Resolves #1369

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Support for OpenShift 3.9 and 3.10 is removed in this release.
   [conjurdemos/kubernetes-conjur-demo#122](https://github.com/conjurdemos/kubernetes-conjur-demo/issues/122)
 
+### Fixed
+- Automatic endpoint discovery for the aws connector was updated. This
+  functionality modifies the request endpoint. There were 2 bugs where (1) the
+  request host header was not being updated to the discovered endpoint, and (2)
+  the request modification was being done after signing the request which would
+  result in a failing integrity check.
+  [cyberark/secretless-broker#1369](https://github.com/cyberark/secretless-broker/issues/1369)
+
 ## [1.7.1] - 2020-10-20
 
 ### Added
--- a/internal/plugin/connectors/http/aws/aws.go
+++ b/internal/plugin/connectors/http/aws/aws.go
@@ -216,6 +216,7 @@ func setAmzEndpoint(req *gohttp.Request, reqMeta *requestMetadata) error {
 
 	req.URL.Scheme = endpointURL.Scheme
 	req.URL.Host = endpointURL.Host
+	req.Host = endpointURL.Hostname()
 
 	return nil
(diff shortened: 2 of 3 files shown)


diverdane (approved):
This looks good, thanks for fixing this!
Approved, but just wanted to double-check whether this would be considered a user-facing bug and would therefore warrant a CHANGLOG entry?

doodlesbykumbi:
@diverdane Thanks, CHANGELOG entry added

diverdane (approved):
LGTM!

izgeri:
@doodlesbykumbi this may be out of scope for this PR, but I think I saw you mention a mock AWS server that we might be able to use for AWS integration tests. should we file an issue in this project to implement that? it would have been nice to have some tests for this connector.

doodlesbykumbi:
@izgeri I have an issue for that https://github.com/cyberark/secretless-broker/issues/1371. I'll tackle it at some point this week

PR #1374, opened 2021-02-04
PostgreSQL connector supports SSL Mode = prefer from client

SovakPaleny:
### What does this PR do?
- Improved connection handling in PostgreSQL for SSLRequest initial packet. It now correctly signalize not supported SSL and enables downgrade to unencrypted transport. https://www.postgresql.org/docs/current/protocol-flow.html 
-  Easy testable by `psql` command with pgsqlmode changes:
    - `PGSSLMODE=prefer` - didn't work in a current version
    - `PGSSLMODE=disable`- worked well and is still working
    - `PGSSLMODE=require` - bad signaling in the current version - now correctly prints that server doesn't support SSL instead of connection error

#1377

#### Change log
- [X] The CHANGELOG has been updated, or

#### Test coverage
- [X] This PR includes new unit and integration tests to go with the code changes, or

#### Documentation
- [X] This PR does not require updating any documentation, or

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- Fixes PostgreSQL signaling in case of SSLRequest from the client,
+  correctly sends SSL Not Supported message instead of general error.
+  Supports downgrade to unsecured connection in case of `SSLMode = prefer`.  
+  [cyberark/secretless-broker#1377](https://github.com/cyberark/secretless-broker/issues/1377)
+
 ## [1.7.2] - 2020-02-05
 
 ### Added
--- a/internal/plugin/connectors/tcp/pg/connector.go
+++ b/internal/plugin/connectors/tcp/pg/connector.go
@@ -1,6 +1,7 @@
 package pg
 
 import (
+	"io"
 	"net"
 
 	"github.com/cyberark/secretless-broker/internal/plugin/connectors/tcp/pg/protocol"
@@ -43,6 +44,13 @@ func (s *SingleUseConnector) abort(err error) {
 	s.clientConn.Write(pgError.GetPacket())
 }
 
+func (s *SingleUseConnector) sslNotSupported() {
+	if s.clientConn == nil {
+		return
+	}
+	s.clientConn.Write([]byte{protocol.SSLNotAllowed})
+}
+
 // Connect implements the tcp.Connector func signature.
 //
 // It is the main method of the SingleUseConnector. It:
@@ -60,7 +68,10 @@ func (s *SingleUseConnector) Connect(
 	s.clientConn = clientConn
 
 	if err = s.Startup(); err != nil {
-		s.abort(err)
+		// connection closed by the client (f.e. SSLMode=require). Do not send error to the closed connection.
+		if err != io.EOF {
+			s.abort(err)
+		}
 		return nil, err
 	}
 
--- a/internal/plugin/connectors/tcp/pg/startup.go
+++ b/internal/plugin/connectors/tcp/pg/startup.go
@@ -11,24 +11,23 @@ import (
 func (s *SingleUseConnector) Startup() error {
 	s.logger.Debugf("Handling connection %+v -> %+v", s.clientConn.RemoteAddr(), s.clientConn.LocalAddr())
 
-	messageBytes, err := protocol.ReadStartupMessage(s.clientConn)
+	version, options, err := s.parseStartupMessage()
 	if err != nil {
 		return err
 	}
-
-	version, options, err := protocol.ParseStartupMessage(messageBytes)
-	if err != nil {
-		return err
-	}
-
-	s.logger.Debugf(
-		"s.Client version: %v, (SSL mode: %v)",
-		version,
-		version == protocol.SSLRequestCode)
 
 	// Handle the case where the startup message was an SSL request.
 	if version == protocol.SSLRequestCode {
-		return fmt.Errorf("SSL not supported")
+		s.sslNotSupported()
+		version, options, err = s.parseStartupMessage()
+		if err != nil {
+			return err
+		}
+
+		// There should not be a second SSLRequest
+		if version == protocol.SSLRequestCode {
+			return fmt.Errorf("Unexpected SSL Request after SSL not supported response")
+		}
 	}
 
 	var ok bool
@@ -38,4 +37,23 @@ func (s *SingleUseConnector) Startup() error {
 	}
 
 	return nil
+}
+
+func (s *SingleUseConnector) parseStartupMessage() (int32, map[string]string, error) {
+	messageBytes, err := protocol.ReadStartupMessage(s.clientConn)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	version, options, err := protocol.ParseStartupMessage(messageBytes)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	s.logger.Debugf(
+		"s.Client version: %v, (SSL mode: %v)",
+		version,
+		version == protocol.SSLRequestCode)
+
+	return version, options, nil
 }
--- a/test/connector/tcp/pg/tests/essentials_test.go
+++ b/test/connector/tcp/pg/tests/essentials_test.go
@@ -93,7 +93,7 @@ func TestEssentials(t *testing.T) {
 					Password: "wrongpassword",
 					SSL:      true,
 				},
-				CmdOutput: StringPointer("SSL not supported"),
+				CmdOutput: StringPointer("server does not support SSL, but SSL was required"),
 			},
 		})
 	})
--- a/test/connector/tcp/pg/tests/sslmode_test.go
+++ b/test/connector/tcp/pg/tests/sslmode_test.go
@@ -1,0 +1,42 @@
+package tests
+
+import (
+	"os/exec"
+	"strings"
+	"testing"
+
+	. "github.com/cyberark/secretless-broker/test/util/testutil"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestCustom(t *testing.T) {
+
+	lc := FindLiveConfiguration(
+		AbstractConfiguration{
+			SocketType: TCP,
+			TLSSetting: TLS,
+			SSLMode:    Default,
+		},
+	)
+
+	connectPort := lc.ConnectionPort
+
+	connectionParams := []string{"dbname=postgres", "password=x"}
+	args := []string{
+		"-c", "select count(*) from test.test",
+		"--username", "x",
+		"--port", connectPort.ToPortString(),
+		"--host", connectPort.Host(),
+	}
+
+	Convey("sslmode=require", t, func() {
+		cmdOut, _ := exec.Command("psql", append(args, strings.Join(append(connectionParams, "sslmode=require"), " "))...).CombinedOutput()
+		So(string(cmdOut), ShouldContainSubstring, "psql: server does not support SSL, but SSL was required")
+	})
+
+	Convey("sslmode=prefer", t, func() {
+		cmdOut, _ := exec.Command("psql", append(args, strings.Join(append(connectionParams, "sslmode=prefer"), " "))...).CombinedOutput()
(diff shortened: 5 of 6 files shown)


doodlesbykumbi (changes requested):
Hi @SovakPaleny. 

Thank you for your contribution. I would like to start by directing you to our [contributor's guide](https://github.com/cyberark/community/blob/master/CONTRIBUTING.md) which covers some important considerations regarding contribution and offers general guidance. Reporting an issue is a great place to start, it allows for discussion , e.g.around things like motivation, and setting expectations before we get to implementation.

Re: the contents of the PR. My understanding is that the motivation behind this is to have the client accurately report that the server (i.e. Secretless in this case) doesn't support SSL. This is relevant to when the client attempts to connect via TLS to Secretless.  A solution relies on Secretless following the protocol and responding with the appropriate errors to the client. This is without-a-doubt a worthwhile issue for us to resolve.

It would need:

1. Updating the errors Secretless sends to the client
2. Adding test case/s to validate Secretless's responses based on the client's reporting

I see that this PR attempts to address both these points. I've left comments for both. I think (2) requires a change in approach. `AbstractConfiguration` configures `Secretless -> Target Service`. For `Client -> Secretless` you might need update `ClientConfiguration` or create an altogether independent test case where you can invoke `psql` with arbitrary arguments.

internal/plugin/connectors/tcp/pg/startup.go
```diff
@@ -28,7 +28,8 @@ func (s *SingleUseConnector) Startup() error {
 
 	// Handle the case where the startup message was an SSL request.
 	if version == protocol.SSLRequestCode {
-		return fmt.Errorf("SSL not supported")
+		s.sslNotSupported()
+		return s.Startup()
```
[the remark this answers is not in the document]
SovakPaleny: From the specification, when the client sends SSLRequest, the server must respond with `S` or `N`. If the response is `S`, SSL Handshake begins (this is not supported so we will never send `S`). In the case of the `N` response, the client must send StartupMessage, so it's the same start as `SSLMode=disable` so we need all the code in the Startup function again. 
So it's the same connection that will continue from the protocol start again.

test/connector/tcp/pg/tests/essentials_test.go
```diff
@@ -78,6 +78,24 @@ func TestEssentials(t *testing.T) {
 			},
 		})
 
+		RunTestCase(TestCase{
+			AbstractConfiguration: AbstractConfiguration{
+				SocketType:     TCP,
+				TLSSetting:     TLS,
+				SSLMode:        Prefer,
```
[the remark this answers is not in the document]
SovakPaleny: OK, thank you for an explanation. I'll rework it.

internal/plugin/connectors/tcp/pg/startup.go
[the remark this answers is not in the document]
doodlesbykumbi: Thank you for the explanation. I can see the recursion resulting in an infinite loop since we rely on the client to provide the base case. One way we can circumvent this is by ensuring startup can at most recurse N (I assume once is enough) number of times before the connection closes.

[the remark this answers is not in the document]
SovakPaleny: I agree, I'll fix the infinite loop. Is it a test that is failing? I run the bin/test and it looks fine :-/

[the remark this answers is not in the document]
doodlesbykumbi: No, no failing test. Purely anticipation. It's definitely not a case that we test right now.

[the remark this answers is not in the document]
SovakPaleny: 👍  it's rewritten without a recursion now

test/connector/tcp/pg/tests/essentials_test.go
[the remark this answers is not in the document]
SovakPaleny: It's reworked, I hope in a better way :-)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Support for OpenShift 4.3 and 4.5.
   [conjurdemos/kubernetes-conjur-demo#122](https://github.com/conjurdemos/kubernetes-conjur-demo/issues/122)
+- Support for `SQLMode = prefer` in PostgreSQL
```
doodlesbykumbi, line 12: Let's update this to be under `### Fixed` since this is a bug. I think we should be more descriptive and provide some detail about the fix. Like how it fixes the error reporting for when PostgreSQL clients attempt to connect to Secretless with TLS, and now Secretless correctly informs the client that TLS is not supported.

internal/plugin/connectors/tcp/pg/startup.go
```diff
@@ -11,24 +11,23 @@ import (
...
-		return fmt.Errorf("SSL not supported")
+		s.sslNotSupported()
+		version, options, err = s.parseStartupMessage()
+		if err != nil {
+			return err
+		}
+
+		// There should not be a second SSLRequest
+		if version == protocol.SSLRequestCode {
+			return fmt.Errorf("SSL not supported")
```
doodlesbykumbi, line 29:
```suggestion
			return fmt.Errorf("SSL Request after SSL not supported response")
```

test/util/testutil/types.go
```diff
@@ -100,6 +100,8 @@ const (
 	Default SSLMode = ""
 	// Disable SSLMode
 	Disable = "disable"
+	// Prefer SSLMode
+	Prefer = "prefer"
```
doodlesbykumbi, line 104: I think the new test cases get us to a good place, but we might have to find another place for this value. `types.go` contains values that are really dedicated to `AbsractionConfiguration` i.e. `Secretless -> Target Service`. These are values common to the Secretless DB connectors configuration. I think somewhere closer to `ClientConfiguration` would be better, though "prefer" is really just a concept in PostgeSQL (I think).

I'll see if I can suggest a better place for it.

doodlesbykumbi (changes requested):
Thank you for addressing the feedback. The PR is moving in the right direction. I've left some additional feedback.

CHANGELOG.md
doodlesbykumbi: Please link to the issue in the CHANGELOG entry, like the one above.

SovakPaleny: fixed

test/util/testutil/types.go
SovakPaleny: I removed the constant completely, it's just as a literal in the test. If you have a better idea, let me know and I'll fix it.

internal/plugin/connectors/tcp/pg/connector.go
```diff
@@ -60,7 +68,10 @@ func (s *SingleUseConnector) Connect(
 	s.clientConn = clientConn
 
 	if err = s.Startup(); err != nil {
-		s.abort(err)
+		// connection closed by the client (f.e. SSLMode=require). Do not send error to the closed connection.
```
doodlesbykumbi, line 71: Nice

internal/plugin/connectors/tcp/pg/startup.go
```diff
@@ -11,24 +11,23 @@ import (
...
-		return fmt.Errorf("SSL not supported")
+		s.sslNotSupported()
+		version, options, err = s.parseStartupMessage()
+		if err != nil {
+			return err
+		}
+
+		// There should not be a second SSLRequest
+		if version == protocol.SSLRequestCode {
+			return fmt.Errorf("SSL Request after SSL not supported response")
```
doodlesbykumbi, line 29: Sorry after some reflection I think this is a better message
```suggestion
			return fmt.Errorf("Unexpected SSL Request after SSL not supported response")
```

test/util/testutil/types.go
doodlesbykumbi: Thank you. I'd like to keep `types.go` decoupled from the clients. I think for now we can use just literal strings for `ClientConfiguration.SSLMode`.

doodlesbykumbi: A better way would probably be to define sslmode string constants local to each client.

doodlesbykumbi (commented):
Responded to some comments. I think the PR looks good. What remains is regarding the tests. 

To avoid coupling with `types.go` the `SSLmode` values on `ClientConfiguration` should be a string literal or we should create service local string consts.

Since the `ClientConfiguration` has changed we'll have to update all usages of it e.g. the mysql essential tests use it too. I think there are compilation errors right now.

I think it might be simpler to not bother with `ClientConfiguration` at all, and instead add a bespoke test set using psql with sslmode=prefer. This whole framework with `ClientConfiguration` etc. is quite clunky and on the list of tech debt to be addressed soon.

doodlesbykumbi:
@SovakPaleny For the bespoke tests I did a quick POC. I did it in GH codespaces but I'm struggling to push to here's the diff. With this approach you don't need to touch `ClientConfiguration` etc. at all.

```
diff --git a/test/connector/tcp/pg/tests/example_test.go b/test/connector/tcp/pg/tests/example_test.go
new file mode 100644
index 0000000..78b8b14
--- /dev/null
+++ b/test/connector/tcp/pg/tests/example_test.go
@@ -0,0 +1,44 @@
+package tests
+
+import (
+	"os/exec"
+	"strings"
+	"testing"
+
+	. "github.com/cyberark/secretless-broker/test/util/testutil"
+)
+
+func TestCustom(t *testing.T) {
+
+	lc := FindLiveConfiguration(
+		AbstractConfiguration{
+			SocketType: TCP,
+			TLSSetting: TLS,
+			SSLMode:    Default,
+		},
+	)
+
+	connectPort := lc.ConnectionPort
+
+	connectionParams := []string{"dbname=postgres", "password=x"}
+	args := []string{
+		"-c", "select count(*) from test.test",
+		"--username", "x",
+		"--port", connectPort.ToPortString(),
+		"--host", connectPort.Host(),
+		
+	}
+
+	Convey("sslmode=require", t, func ()  {
+		cmdOut, _ := exec.Command("psql", append(args, strings.Join(append(connectionParams, "sslmode=require"), " "))...).CombinedOutput()
+		So(string(cmdOut), ShouldContainSubstring, "psql: server does not support SSL, but SSL was required")
+	})
+
+	Convey("sslmode=prefer", t, func ()  {
+		cmdOut, _ := exec.Command("psql", append(args, strings.Join(append(connectionParams, "sslmode=prefer"), " "))...).CombinedOutput()
+		So(string(cmdOut), ShouldContainSubstring, "count")
+	})
+
+	
+}
diff --git a/test/util/testutil/run_test_case.go b/test/util/testutil/run_test_case.go
index 09c6ac8..fd1a404 100644
--- a/test/util/testutil/run_test_case.go
+++ b/test/util/testutil/run_test_case.go
@@ -37,6 +37,14 @@ func NewRunTestCase(runQuery RunQueryType) RunTestCaseType {
 	}
 }
 
+// FindLiveConfiguration takes AbstractConfiguration and finds a live configuration for
+// an active Secretless endpoint
+func FindLiveConfiguration(ac AbstractConfiguration) LiveConfiguration {
+	_, testSuiteConfigurations := GenerateConfigurations()
+
+	return testSuiteConfigurations.Find(ac)
+}
+
 // RunQueryType represents a function that takes in database credentials
 // and options, uses them to execute a test query, and returns the output
 // of that query.
```

SovakPaleny:
@doodlesbykumbi I used your POC, thank you for it. I reverted all the other test changes.

doodlesbykumbi:
@SovakPaleny This looks good to me. It needs rebasing and squashing into a singular (I think that should be fine) commit that captures the change. I will then close the PR, cherry-pick your commit/s, and create a branch in this repo so that Jenkins can run the build (which only happens for internal branches).

SovakPaleny:
@doodlesbykumbi rebased and squshed

doodlesbykumbi (approved):
LGTM! Thank you for your work @SovakPaleny. This is a great contribution!

doodlesbykumbi:
@SovakPaleny 😓 It was only in the wake of merging that Capybara decided to complain about a drop in test coverage so we've had to add some unit tests for the startup logic in the connector. Thought it might be of interest, see https://github.com/cyberark/secretless-broker/pull/1382.

PR #1382, opened 2021-02-08
Add unit tests for postgres connector startup message logic

doodlesbykumbi:
### What does this PR do?

Add unit tests for postgres connector handling startup messages

The test cases might look a bit scary but they're hyper specific and only deal with a single request-response message types. These are:
1. Startup message for Postgres (from the client) which contains a "version" at the start (tells us to use TLS or not), followed by options (e.g. username, database etc.).
2. Response to the client. For the tests we actually only care about the first byte which tells us if it's an error or not.

There are 3 tests
1. Corresponds to `sslmode=disable`. The client doesn't request TLS.
2. Corresponds to `sslmode=prefer`. The client requests TLS, Secretless says TLS not supported, The client sends the startup message against with no TLS request
3. Corresponds to `sslmode=INSIST!`. This one is made up but validates the logic for a client that keeps sending startup messages requesting TLS when Secretless has responded with TLS not supported. We'd never expect a normal client to do this but you get abusive clients. This test would catch things like infinite recursion had we decided to implement the second read of the startup message using recursion.

`sslmode=require` would correspond to (2) with the client closing the connection after Secretless responds.

Resolves build failures from low coverage. Capybara 😓 

#### Change log

#### Test coverage

#### Documentation


diverdane (approved):
LGTM! Your explanation of what the test was doing in the PR description was key to understanding this. Nice bits of UT!

PR #1441, opened 2021-12-13
Update golang to 1.17

szh:
Use the latest major version of Go. Also fix ssh tests by replacing RSA test cert with ECDSA

### Connected Issue/Story

CyberArk internal issue link: [ONYX-14555](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-14555)

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]()

#### Behavior

#### Security


doodlesbykumbi (commented):
@szh I think we still need to add `HostKeyAlgorithms=ssh-rsa` to the ssh test service command. Without it we will have occasional failures like this:
```
/ # ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -p 2222 secretless ls -la
Unable to negotiate with 172.28.0.4 port 2222: no matching host key type found. Their offer: ssh-rsa
```

I'm not sure why these are occasional and not all the time. But when we add the `HostKeyAlgorithms` option it works every time. I know that the occasional failures are resolved by this because the Secretless SSH server uses an ssh-rsa key (and so the client needs to enable that).

doodlesbykumbi (approved):
LGTM! Thanks for doing this

PR #1446, opened 2022-01-13
Bump authn-k8s, conjur-api-go, summon and other dependencies

szh:
### Desired Outcome

- Use the latest versions of conjur-authn-k8s-client with support for JWT and tracing.
- Use the latest version of other dependencies.

### Implemented Changes

- Updated versions of dependencies
- Modified `internal/providers/conjur/provider.go` to use the new authn flow that supports JWT
- Added `go:build ignore` flags to sample CRD go files to avoid compilation errors. These files are provided for documentation purposes and are not meant to be compiled with the project.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-16170](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-16170)

### Definition of Done

- [x] Using latest version of cyberark dependencies
- [x] All tests pass

#### Changelog

- [ ] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]()

#### Behavior

#### Security


tzheleznyak (dismissed):
LGTM

szh:
Waiting to update until new authn-k8s-client is released

go.mod
```diff
@@ -4,9 +4,9 @@ require (
 	github.com/aws/aws-sdk-go v1.15.79
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/codegangsta/cli v1.20.0
-	github.com/cyberark/conjur-api-go v0.5.2
-	github.com/cyberark/conjur-authn-k8s-client v0.19.1
-	github.com/cyberark/summon v0.7.0
+	github.com/cyberark/conjur-api-go v0.8.1
+	github.com/cyberark/conjur-authn-k8s-client v0.22.1-0.20220112215633-88e85b7e63ea
```
tzheleznyak, line 8: Do we want to wait for release so here it will be just 0.23.0 ?

tzheleznyak: And in secrets-provider too?

internal/providers/conjur/provider.go
```diff
@@ -97,10 +95,17 @@ func ProviderFactory(options plugin_v1.ProviderOptions) (plugin_v1.Provider, err
 		log.Printf("Info: Conjur provider using Kubernetes authenticator-based authentication")
 
 		// Load the authenticator with the config from the environment, and log in to Conjur
-		if authenticator, err = loadAuthenticator(provider.AuthnURL, provider.Version, provider.Config); err != nil {
+		conjurAuthenticatorConf, err = authnConfig.NewConfigFromEnv()
```
diverdane, line 98: Since you're making changes in this section of the code...  This may be a personal preference sort of thing, but instead of using a but instead of bunch of `if`...`if-else`...`if-else` statements, we could use a switch:
```
switch {
case provider.Username != "" && provider.APIKey != "":
    // Conjur provider using API key. Do stuff
case tokenFile != "":
    // Conjur provider using access token. Do stuff
case provider.AuthnURL != "" && strings.Contains(provider.AuthnURL, "authn-k8s"):
    // Conjur provider using authenticator. Do stuff
default:
    // Return error
}
```
For me, the switch is easier to see what's going on, but your call, don't want to hold up on this.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Use latest version of conjur-authn-k8s-client which supports JWT loging and tracing.
+  [cyberark/secretless-broker#1446](https://github.com/cyberark/secretless-broker/pull/1446)
+
```
diverdane, line 12: I'm wondering if we should include a CHANGELOG.md entry to say that we bumped K8s API version to 1.23.
This *might* be considered a user-facing change, since it affects in which Kubernetes clusters the SP can be run.
(Although this may only affect users who are using CRD/CRs to configure the SP... since I think that's why we're depending on Kubernetes API client and machinery?).

go.mod
```diff
@@ -22,10 +22,10 @@ require (
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/crypto v0.0.0-20211215153901-e495a2d5b3d3
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.22.3
+	k8s.io/api v0.23.1
```
diverdane, line 25: Nice work. These K8s updates are always so tricky.

diverdane (commented):
Looks great!!! Thanks to you and @tzheleznyak for getting this moving along!
A couple of suggestions, but your call on those.

CHANGELOG.md
szh: I believe it only affect CRDs

internal/providers/conjur/provider.go
szh: Looking at the Effective Go [style guide](https://go.dev/doc/effective_go#switch):
> It's therefore possible—**and idiomatic**—to write an if-else-if-else chain as a switch
So I'll make this change

diverdane: Nice, it's much easier to read!

diverdane (approved):
LGTM!!!

PR #1447, opened 2022-01-14
Bump version to 1.7.9

szh:


NOTICES.txt
```diff
@@ -173,7 +173,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 
->>> k8s.io/api-0.0.0-20180712090710-2d6f90ab1293
+>>> k8s.io/api-0.23.1
 
 Copyright [yyyy] [name of copyright owner]
```
john-odonnell, line 178: I see a lot of lines like this in `NOTICES.txt`, with placeholders for year and copyright owner. Should these be filled out?

I'll still approve, as the file has had these placeholders for at least 2 years.

john-odonnell (approved):
Approved. Curious about placeholders in `NOTICES.txt`, but they've been in the file for ages, so I won't let it get in the way.

PR #1463, opened 2022-05-10
Fix `go mod tidy` goreleaser dependency execution

neil-k-zero:
- Gorelease and github ci pipeline now reference
the `bin/go_mod_tidy` shell script to update the
go.mod file with the correct dependencies

### Desired Outcome

- Generate the correct `go mod tidy -compat` version based
on the go.mod file version

### Implemented Changes

- Added a `bin/go_mod_tidy` script that is called by goreleaser
and Github ci pipeline

#### Changelog

  CHANGELOG update


szh (approved):
Brilliant!

